PR #100, opened 2014-02-20
Added option to disable suffix on diffForHumans

BlueHayes:
I was disappointed to find that there is no method for a 'human time' without the suffix in Carbon. I added a second parameter to diffForHumans were the suffix can be disabled for this reason.


briannesbitt:
See #230

PR #105, opened 2014-03-15
Default Localization Method

hkan:
According to the [Wikipedia](http://en.wikipedia.org/wiki/Date_format_by_country), a huge number of countries use the *little-endian* format for dates. That's why I added `toLocalizedDateString` method which does that without bloating the views (or whereever developer converts dates to localized strings).

There of course must be a locale check to decide which \*-endian format to use, but until me or someone else does it, this should be enough.

**The problem with the readme.md** is because I don't have German locale installed on my machine and my default locale is Turkish. So a lot of thing has changed. It would be probably better if you run the PHP again and replace the readme file.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -177,13 +177,13 @@ To accompany `now()`, a few other static instantiation helpers exist to create w
 
 ```php
 $now = Carbon::now();
-echo $now;                               // 2014-01-06 22:52:03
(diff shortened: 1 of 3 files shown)


briannesbitt:
No tests.  I also think this is better as your own helper method.

Issue #87, opened 2014-01-28
New Feature: Human Readable DateInterval


benglass:
Any interest in a feature to take a DateInterval or number of seconds and pretty print it in human readable form of "1 day 3 hours 2 minutes" and vice versa. Need some code like this in a project recently and checked to see if it was available in Carbon. If its of interest I can look at submitting a pull request.

PR #112, opened 2014-03-29
[WIP] Add DateInterval handling to Carbon

benglass:
This is a work in progress to address #87

It provides a CarbonInterval class which has a handle create method for creating an interval without using the strange interval_spec ('P1Y2M') using CarbonInterval::create(1, 4). It also has a intervalForHumans method which will convert an interval into a readable string ('1 year 4 months').


briannesbitt:
https://github.com/briannesbitt/Carbon/commit/552017d416c056f36955162440b5e9da495c6a7a

Thanks !

PR #124, opened 2014-05-15
Changed return type to static

dcsg:
This change will allow people that extends `Carbon` to have correct typehint in the IDE.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -191,7 +191,7 @@ class Carbon extends DateTime
     *
     * @param  DateTime $dt
     *
-    * @return Carbon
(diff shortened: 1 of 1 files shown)


nicholasnet:
This merge fixes the extension issue but breaks other one. Consider this example.

```
class Foo
{
    /**
     * @var \Carbon\Carbon
     */
    protected $foo;

    public function __construct()
    {
        $this->foo = \Carbon\Carbon::now();
    }
}

class Bar extends Foo
{
    public function test()
    {
        $this->foo->format('Y-m-d'); // This won't resolve now because of static phpdoc comment.
    }
}
```
In my opinion we should think about this merge again.

PR #141, opened 2014-07-04
Added fr formatter

bgaillard:
Add a French Carbon Diff Formatter.


briannesbitt:
See implementation on localized branch.

PR #194, opened 2014-09-26
Upgrade to php 54

lucasmichot:
> Maybe we dump PHP 5.3 support anyway since its end of life as of Aug 14.

Related to https://github.com/briannesbitt/Carbon/pull/192
Here it is

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.3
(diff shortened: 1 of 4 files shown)


briannesbitt:
I am going to close this as I think 2.0 will jump to 5.5.  The 1.X branch will still be 5.3 and supported for awhile.

PR #205, opened 2014-10-16
timezone support for method formatLocalized

php-workx:
formatLocalized didn't respect the timezone set on a date object. this can be fixed by using ...

    strtotime($this)

instead of 
    
    $this->timestamp


F21:
Would love to see this merged as well!

briannesbitt:
Ah yes ... good fix.  Merging.

PR #206, opened 2014-10-24
Added secondsSinceMidnight and secondsUntilEndOfDay helper methods

RyanNielson:
Wrote these two helpers methods because they are often quite useful. They're based of similar methods in Ruby On Rails found [here](http://api.rubyonrails.org/classes/DateTime.html#method-i-seconds_since_midnight) and [here](http://api.rubyonrails.org/classes/DateTime.html#method-i-seconds_until_end_of_day).

- secondsSinceMidnight gets the number of seconds since midnight of the current Carbon object.
- secondsUntilEndOfDay get the number of seconds from the current Carbon object to 23:59:59 of the current day.


smazurov:
Thanks for adding those & Merging!

PR #234, opened 2014-12-18
Fix wrong output from tomorrow's date in the readme

sp-gonzalo-serrano:
Hi!

I think i found a typo, if you execute the following script in a composer-based environment you can check the correct output.

```
<?php

require_once 'vendor/autoload.php';

use Carbon\Carbon;

Carbon::setTestNow(Carbon::createFromTimestamp(strtotime('2014-09-30 21:03:42')));

$now = Carbon::now();
echo $now . "\n";

echo $today . "\n";
$tomorrow = Carbon::tomorrow('Europe/London');
echo $tomorrow . "\n";
$yesterday = Carbon::yesterday();
echo $yesterday . "\n";

```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -181,7 +181,7 @@ echo $now;                               // 2014-09-30 21:03:42
 $today = Carbon::today();
 echo $today;                             // 2014-09-30 00:00:00
 $tomorrow = Carbon::tomorrow('Europe/London');
-echo $tomorrow;                          // 2014-10-02 00:00:00
+echo $tomorrow;                          // 2014-10-01 00:00:00
 $yesterday = Carbon::yesterday();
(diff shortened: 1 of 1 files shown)


briannesbitt:
First I get a $today is not defined.  But more importantly I get Tomorrow ('Europe/London') being 2014-10-02 correctly.  Today in my timezone is 2014-09-30.

Tomorrow is implemented as ```now($tz)->startOfDay()->addDay();```.  When you break this down for the given timezone now in `Europe/London` time is already 2014-10-01 (remember I am UTC -5), then you add a day and you get 2014-10-02 as provided.

All depends on what time of day you run it at and what timezone you are in.

PR #235, opened 2014-12-28
Add yearIso property

barryvdh:
This time of year usually gives problems with dates. I have a project that works with week numbers.

$date = new \Carbon\Carbon('2014-12-30');
echo $date->weekOfYear .' '.$date->year;

Gives `1 2014`. So the correct usage (in this particular case) is `$date->format('o')`, which gives 2015.

Would it be good to give this a property on the date object? Something like `$date->yearIso` / `$date->IsoYear`? Even though it's not much shorter then `format->('o')` but it might make it more obvious to people with autocomplete etc.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -22,6 +22,7 @@ use InvalidArgumentException;
  * A simple API extension for DateTime
  *
  * @property      integer $year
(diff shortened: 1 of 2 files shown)


dereuromark:
:+1:

Issue #236, opened 2015-01-01
Month overflow


tshelburne:
This isn't necessarily an issue, since it appears to be tested [here](https://github.com/briannesbitt/Carbon/blob/master/tests/AddTest.php#L56), but I'm curious as to why the preferred mechanism for adding a month would allow for overflow? Conceptually, I expect adding a month to increment the month by one, and if overflow happens, it will end up on the last day of the next month.

Is there an alternative to `addMonth` that works this way? Would you be opposed to either a boolean flag or a new method that would work this way? I'm going to be looking through the code, but... it's a pretty long file.

kylekatarnls:
It's a bit tricky since adMonth directly call the DateTime::modify function. But you can fix the date after modify:
<pre>$oldDate = Carbon::createFromDate(2012, 1, 31);
$newDate = $oldDate->addMonth();
if($newDate->day < $oldDate->day) {
    $newDate->subMonth();
    $newDate->day = $newDate->daysInMonth;
}</pre>

tshelburne:
@kylekatarnls That is exactly the PR I would submit, except that it would be the default behavior of the `addMonth` function. From my perspective, the more natural behavior is to return a date in the following month, regardless of how many days are in that month. When I call `addDate` on a date in January, I will always expect a date in February to be returned.

briannesbitt:
Rolling over is the expected default behaviour.

Sounds like you want a addMonthNoOverflow() which wouldn't be hard to add, just depends on desired results.

If it were Feb 28 and you added 1 month would you expect Mar 28 or Mar 31 ? (Ie. Do you want it to stick to the end of the month when its longer as well?) Since Jan 31 + 1 month you want Feb 28.

tshelburne:
@briannesbitt Good question - I just submitted PR #239 that shows the exact behavior I would expect, plus my idea for what I think would be a more intuitive interface. That said, it's all subjective.

Happy new year!

kylekatarnls:
Intuitive to you indeed. In all project, we do not break existing functions in minor versions. So if every body think the default PHP DateTime behaviour is a bug, it should only be added in the next major version. But I think (and I'm pretty sure I'm not the only one), default overflow is a good behavior, and you should do what @briannesbitt suggest: create addMonthNoOverflow() and as it do not seem to be a very needed function, you should better to extend Carbon with your own class and add the function in it. Anyway, it's a very interesting poin and it should be great some users give their opinion.

[...]

Issue #217, opened 2014-10-31
Problem Setting Month


warkior:
Hi there,

I've put the following code on a page of my site, and it is returning the wrong month:
`Carbon::now(new DateTimeZone('America/Vancouver'))->month(6)->toRfc850String()`

When dumped returns:
`string 'Tuesday, 01-Jul-14 09:48:35 PDT' (length=31)`

Note that month 6 is returning July. I assume setting the month manually is not zero based.

This *may* be related to this issue, but then I'm not sure.
https://github.com/briannesbitt/Carbon/issues/216

Thanks in advance,
James.

tshelburne:
Have a look at PR #239 - it creates a set of methods that handle this situation. I also submitted an idea to update the interface to make it more "predictable", but that's much more subjective.

kylekatarnls:
Same problem source as #236 created by @tshelburne and so you can check the day after set the month in the same way:
<pre>$oldDate = Carbon::now(new DateTimeZone('America/Vancouver'));
$newDate = $oldDate->month(6)->toRfc850String();
if($newDate->day &lt; $oldDate->day) {
    $newDate->subMonth();
    $newDate->day = $newDate->daysInMonth;
}</pre>

briannesbitt:
I'll look at the PR

PR #239, opened 2015-01-01
Add / Subtract months with no overflow

tshelburne:
This PR answers both issues #236 and #217 - in order for the tests to all pass on Travis, it depends on another PR I submitted this morning fixing a `diffForHumans` issue that popped up after the new year.

Although this is influenced by the default PHP behavior, I also think it is worth considering adding an argument flag to the "base" add and subtract methods - the behavior seems counter-intuitive as it is, and it doesn't feel nearly as semantically satisfying to preserve it.

Just a thought - perhaps something like the following, where the overflow method would be protected:

```
public function addMonths($value, $useDefault = false)
{
    return $useDefault ? $this->modify((int) $value . ' month') : $this->addMonthsNoOverflow($value);
}

Carbon::now()->addMonths(2); // intuitive behavior
Carbon::now()->addMonths(2, true); // PHP default behavior
```

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1349,6 +1349,64 @@ class Carbon extends DateTime
     }
 
     /**
+     * Add months without overflowing to the instance. Positive $value 
+     * travels forward while negative $value travels into the past.
+     *
+     * @param integer $value
+     *
+     * @return static
+     */
+    public function addMonthsNoOverflow($value)
+    {
+        // jump straight into the month to avoid any php overflow weirdness
+        $newYear   = $this->year + floor(($this->month + $value) / 12);
+        $newMonth = ($this->month + $value) % 12;
+        $newDate = self::create($newYear, $newMonth, $this->day, $this->hour, $this->minute, $this->second, $this->getTimeZone());
+
+        if(($newDate->day != $this->day)) {
+            // it seems like this would be dependent on whether the value is
+            // positive or negative, but because of how we set the month above,
+            // we should always zero out and subtract a month
+            $newDate->day(1)->subMonth();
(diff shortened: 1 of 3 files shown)


briannesbitt:
Default on 2nd param needs to be true so its not a breaking change.

tshelburne:
Yea, I was adding that bit of code in the description (it's not in the actual PR) in as a possible direction down the road, like if you are updating versions, where a breaking change would be acceptable. That interface is subjective, but I like it.

kylekatarnls:
It's not intuitive to me. Do not break the existing function, keep the PHP default as Carbon default and add a flag to change it to your own need.

eblanshey:
+1. Would like to see this merged (with $useDefault defaulted to true, of course).

PrplHaz4:
`useDefault` parameter does not seem very user-friendly to me - maybe `noOverflow` or something that more clearly states what behavior will be used?

tshelburne:
@PrplHaz4 I agree with that - so it's said, that feature isn't *actually* part of this PR. All this PR does is provide the `addMonthNoOverflow` method. I figured that was a pretty painless addition, and could be merged while this discussion continues.

PrplHaz4:
@tshelburne oh, yea, completely forgot you had broken them up (which is definitely the right way to approach it)...FTR, I think this is a great addition.

judgej:
These `addMonthsNoOverflow()` functions and friends just return a new value, and don't update the value of the underlying object like `addMonths()` and friends do. Was that deliberate, or an oversight?

tshelburne:
That's *probably* an oversight - my tendency is to always go immutable, so I imagine I just didn't pay close attention to the other functions, and went with my gut. A good catch - best to ask @briannesbitt his preference.

PR #260, opened 2015-02-23
Ternary expression

kylekatarnls:
If you use ternary expression, it can be two in one.


briannesbitt:
I don't mind this one!  Thanks for your efforts!

Might as well just put the params in the method call now !

PR #261, opened 2015-02-23
Only one return statement

kylekatarnls:
Best-pratice: have only one return statement at the end of the method.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1950,19 +1950,20 @@ class Carbon extends DateTime
 
         $time = static::translator()->transChoice($unit, $count, array(':count' => $count));
 
-        if ($absolute) {
-            return $time;
+        if (! $absolute) {
+
+            $params = array(':time' => $time);
+    
+            if ($isNow) {
+                $transId = $isFuture ? 'from_now' : 'ago';
+            } else {
+                $transId = $isFuture ? 'after' : 'before';
+            }
+            
+            $time = static::translator()->trans($transId, $params);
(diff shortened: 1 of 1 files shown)


dereuromark:
Returning early is a useful and good coding practice. Besides the minor speed improvement.
I would not sacrifice that here.

kylekatarnls:
It's not a speed issue, it's about readability, you cannot miss the return statement if there is only one at the end, you know immediatly if the function is void or not and what is returned. If you have many return statement, you can miss those who are "hidden" in some other statement and it more confuse to see what is excuted or not. Here, all the non-absolute is in a if(! $absolute) statement, that is more semantic. You know immediatly that this part is not executed if $absolute is falsy.

Here is just a suggestion and I understand the speed argument.

dereuromark:
Speed is irrevelant (or so minor that it would be in this case). As pointed out, the coding practice "return early" is what most developers would prefer here. It reduces code complexity and unnecessary nesting.

kylekatarnls:
Yes it can be easier to read a function using return early but I would not say it's true in this case. IMHO. If you prefer less nesting, it could also be done like this:
<pre>
        return $absolute ?
            $time :
            static::translator()->trans(
                $isNow ? ($isFuture ? 'from_now' : 'ago') : ($isFuture ? 'after' : 'before'),
                array(':time' => $time)
            );</pre>

All is OK for me. I just would to submit some ideas.

briannesbitt:
I much prefer the return early and less nesting.  I'll pull #260.

PR #263, opened 2015-02-23
Alternative to #260 and #261 if prefered

kylekatarnls:
Short and fast. Feel free to close if not appropriate.


briannesbitt:
This looks like something I would write in perl .... when I was 20 :-)

PR #264, opened 2015-02-23
Added french localization

Cederman:
French localization added. Hopefully no big errors :smile:


briannesbitt:
Thanks but French was already added.  I think you forked from master?

PR #280, opened 2015-03-01
Simplify addMonthsNoOverflow

lucasmichot:


GrahamCampbell:
```diff
There were 2 failures:
1) DiffTest::testDiffForHumansNowAndMonth
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-4 weeks ago
+1 month ago
/home/travis/build/briannesbitt/Carbon/tests/DiffTest.php:470
2) DiffTest::testDiffForHumansOtherAndFutureMonth
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-4 weeks after
+1 month after
```

lucasmichot:
Yeah, I know but this is not related to this change. Tests are even failing on master.
I need to fix that - depending on the period of the year it can breaks so I think this might be related to Winter/Summer time

briannesbitt:
I think its because the time change is coming up soon.   Thought it was fixed with the move to `$this->diff()`.  Could also be the short feb month.

briannesbitt:
It was erroring out on subWeeks(4) ~= "4 weeks ago"....  Today being Mar 2... 4 weeks ago is Feb 2 which is actually "1 month ago" as the test works now.  Its a better implementation now as this shows !

PR #287, opened 2015-03-03
fi language

kylekatarnls:
Add fi language from laravel-date project


jozan:
As a native I can say looks right.

ngtranthanhtoan:
Does the pakagist have more function? Where can I find it, may be it not enough for me to manipulate the date and time for my project

kylekatarnls:
Sorry, what do you want to do exactly?

PR #289, opened 2015-03-03
Apply PSR-2 syntax for localized branch

lucasmichot:


src/Carbon/Carbon.php
```diff
@@ -79,7 +79,7 @@ class Carbon extends DateTime
         self::WEDNESDAY => 'Wednesday',
         self::THURSDAY => 'Thursday',
         self::FRIDAY => 'Friday',
-        self::SATURDAY => 'Saturday'
+        self::SATURDAY => 'Saturday',
```
GrahamCampbell: this is not a psr-2 fix...

src/Carbon/Carbon.php
```diff
@@ -152,7 +152,7 @@ class Carbon extends DateTime
     protected static function safeCreateDateTimeZone($object)
     {
         if ($object === null) {
-            return null;
```
GrahamCampbell: neither is that

src/Carbon/Carbon.php
```diff
@@ -97,7 +97,7 @@ class Carbon extends DateTime
         '-',
         'first',
         'last',
-        'ago'
+        'ago',
```
GrahamCampbell: or that

src/Carbon/Carbon.php
```diff
@@ -1281,7 +1286,7 @@ public function isSameDay(Carbon $dt)
      */
     public function addYears($value)
     {
-        return $this->modify((int) $value . ' year');
+        return $this->modify((int) $value.' year');
```
GrahamCampbell: or that

src/Carbon/Carbon.php
```diff
@@ -1326,7 +1331,7 @@ public function subYears($value)
      */
     public function addMonths($value)
     {
-        return $this->modify((int) $value . ' month');
+        return $this->modify((int) $value.' month');
```
GrahamCampbell: or that

src/Carbon/Carbon.php
```diff
@@ -1422,7 +1427,7 @@ public function subMonthsNoOverflow($value)
      */
     public function addDays($value)
     {
-        return $this->modify((int) $value . ' day');
+        return $this->modify((int) $value.' day');
```
GrahamCampbell: or that, etc...

PR #292, opened 2015-03-05
Add the ability to specify the format

barbuslex:
Hi, 

This modification adds the ability to specify the format of the output string.

Can you implement it to your distribution ?

Thanks

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -836,9 +836,9 @@ class Carbon extends DateTime
      *
      * @return string
      */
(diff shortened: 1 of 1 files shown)


Anahkiasen:
I'm not sure I get the purpose, the point of those methods is to refer to a specific format, if you need a custom one why not just do `->format('Your format')`

barbuslex:
Yes still work with format() function

briannesbitt:
Agreed.  Wouldn't have merged this.

PR #294, opened 2015-03-05
Update Carbon.php

angelomaragna:
I think it's easy for non english people to call addYear instead of addYears.
In my case, I wrongly wrote addYear(2). 

if you call function addYear(2) actually adds one year. 

I think it should raise an exception or better, work as an alias for addYears.

It is right addYear() to add one year.
It is wrong addYear(45) to add one year.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1232,9 +1232,9 @@ class Carbon extends DateTime
      *
      * @return static
      */
-    public function addYear()
+    public function addYear($value = 1)
     {
-        return $this->addYears(1);
+        return $this->addYears($value);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


briannesbitt:
I agree with you.

billmn:
Agree

Now it's important to not introduce breaking changes and fix tests.

briannesbitt:
The failing tests were unrelated.

briannesbitt:
Errr...  `return $this->addWeeks($value = 1);` this looks wrong !

briannesbitt:
... wow... and default params should be "1" not "-1"... missed it but this implementation its very wrong.

angelomaragna:
I agree, sub -1 should add and it is wrong. I am sorry for the 
$this->addWeeks($value = 1);
it is not wrong, it is incredibly wrong. I will compel myself to the good practice of writing tests :(

Sorry again.

briannesbitt:
Thanks.  I thought the failed tests were unrelated at first glance but alas they weren't.  Seemed like an easy feature on the surface, but tests testing the new feature is always a good idea.

PR #321, opened 2015-03-23
Added ability to pass offset in hours instead of timezone name

slik:
Ability to pass offset instead of timezone name was very needed for one of my project, and I decided to create a pull request.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -150,6 +150,11 @@ class Carbon extends DateTime
             return $object;
         }
 
+        if (is_numeric($object)) {
+            $timezone_offset = $object * 3600;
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -131,9 +131,9 @@ class Carbon extends DateTime
     protected static $testNow;
 
     /**
-     * Creates a DateTimeZone from a string or a DateTimeZone
+     * Creates a DateTimeZone from a string, a DateTimeZone or an integer offset
      *
-     * @param DateTimeZone|string|null $object
+     * @param DateTimeZone|string|integer|null $object
```
GrahamCampbell: int

PR #322, opened 2015-03-24
Update Carbon.php: Update minValue() and maxValue() for 64bit systems

lanort:
On 64 bit systems `maxValue()` and `minValue` produce akward dates (see end for examples) that can not be used for database or html date input fields (see bottom for examples).

I propose to use more common max and min values in the case, that integer values are 64 bit, ie '1001-01-01 00:00:00' to '9999-12-31 23:59:59'.

It is possible to use these values for an html date input field as well as in MySQL for DATE, DATETIME ([MySQL 5.0 Reference Manual: 11.3.1 The DATE, DATETIME, and TIMESTAMP Types](https://dev.mysql.com/doc/refman/5.0/en/datetime.html)). MySQL TIMESTAMP only supports a range of '1970-01-01 00:00:01' UTC to '2038-01-19 03:14:07' UTC.

Examples
=======

64 bit
------
64 bit `maxValue()`  

    //PHP_INT_MAX = 9223372036854775807
    Carbon {
      +"date": "-292277022657-01-27 09:29:51.000000"
      +"timezone_type": 3
      +"timezone": "Europe/Vienna"
    }
    
64 bit `minValue()`    

    //~PHP_INT_MAX = 9223372036854775807 (same as PHP_INT_MAX)
    Carbon {
      +"date": "-292277022657-01-27 09:29:52.000000"
      +"timezone_type": 3
      +"timezone": "Europe/Vienna"
    }
    
32 bit
------
32 bit `maxValue()`

    //PHP_INT_MAX = 4294967295
    Carbon {
      +"date": "2106-02-07 07:28:15.000000"
      +"timezone_type": 3
      +"timezone": "Europe/Vienna"
    }
    
32 bit `minValue()`

    //~PHP_INT_MAX = -4294967296
    Carbon {
      +"date": "1833-11-24 18:31:44.000000"
      +"timezone_type": 3
      +"timezone": "Europe/Vienna"
    }

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -299,7 +299,7 @@ class Carbon extends DateTime
             return static::createFromTimestamp(~PHP_INT_MAX);
         } else {
             // 64 bit
-            return static::create(1000, 1, 1, 0, 0, 0);
+            return static::create(1, 1, 1, 0, 0, 0);
         } 
     }
 
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -299,7 +299,7 @@ class Carbon extends DateTime
             return static::createFromTimestamp(~PHP_INT_MAX);
         } else {
             // 64 bit
-            return static::create(1001, 1, 1, 0, 0, 0);
+            return static::create(1000, 1, 1, 0, 0, 0);
         } 
     }
 
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -278,7 +278,13 @@ class Carbon extends DateTime
      */
     public static function maxValue()
     {
-        return static::createFromTimestamp(PHP_INT_MAX);
+        if (PHP_INT_SIZE == 4) {
+            // 32 bit (and additionally Windows 64 bit)
+            return static::createFromTimestamp(PHP_INT_MAX);
+        } else {
+            // 64 bit
+            return static::create(9999, 12, 31, 23, 59, 59);
+        }
     }
 
     /**
@@ -288,7 +294,13 @@ class Carbon extends DateTime
      */
     public static function minValue()
     {
-        return static::createFromTimestamp(~PHP_INT_MAX);
+        if (PHP_INT_SIZE == 4) {
+            // 32 bit (and additionally Windows 64 bit)
+            return static::createFromTimestamp(~PHP_INT_MAX);
+        } else {
+            // 64 bit
+            return static::create(1001, 1, 1, 0, 0, 0);
+        } 
     }
 
     /**


briannesbitt:
FYI: My windows version shows a PHP_INT_MAX as 2147483647... ie signed 32 bit max.

Which produces : `1901-12-13 15:45:52` and `2038-01-18 22:14:07`.

lanort:
On windows x86_64, PHP_INT_MAX is 2147483647. This is because in the underlying c-code, a long is 32 bit. (see eg http://stackoverflow.com/a/3233881/936284)

briannesbitt:
When I run this on a linux box 64 bit PHP 5.5.12 I get:

```
9223372036854775807
object(Carbon\Carbon)#1 (3) {
  ["date"]=>
  string(27) "292277026596-12-04 10:30:07"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(15) "America/Toronto"
}
object(Carbon\Carbon)#1 (3) {
  ["date"]=>
  string(27) "292277026596-12-04 10:30:08"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(15) "America/Toronto"
}
```

briannesbitt:
Reading the mysql doc looks like the min year should be `1000` ?

lanort:
Linux supports the 64 bit integer, so on your 64 bit linux you should get this huge integers. 

Sorry for the mistakes:

 * `~PHP_INT_MAX` of course is `-9223372036854775808`.
 * Min year should be `1000`. I prepared another commit for you.

lucasmichot:
@lanort, Well many developers use different DBMS. The aim of minValue and maxValue was to get the min and max possible date of the system.
Or do you plan to mimic MySQL date ?

lanort:
@lucasmichot: I used MySQL because to my knowledge many php web projects use MySQL. But you are right, that does not mean everybody wants that.

Please find other popular RDBMS examples below: 

**Oracle** Database can store dates in the Julian era, ranging from January 1, 4712 BCE through December 31, 9999 CE (Common Era, or 'AD').
(http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT413)

**Microsoft SQL** datetime can store:
Date range: January 1, 1753, through December 31, 9999
Time range: 00:00:00 through 23:59:59.997
(https://msdn.microsoft.com/de-at/library/ms187819.aspx)

**SAP MaxDB** datetimeformat
Date and Time Format (datetimeformat)
'YYYY': Four-digit year format (0001-9999)
'MM': Two-digit month format (01-12)
'DD': Two-digit day format (01-31)
Time value
'HHHH'/'HH': Four-digit/two-digit hour format
'MM': Two-digit minute format (00-59)
'SS': Two-digit second format (00-59)
(http://maxdb.sap.com/doc/7_7/1c/44a74607704a5bb9fa1d6380efd525/content.htm)

**IBM DB2**
DATE: 0001-01-01 to 9999-12-31
TIME: 00.00.00  to 24.00.00
(http://www.devx.com/dbzone/Article/28713)

briannesbitt:
Matching a particular RDBMS isn't that interesting as Carbon is used with other systems and not just DBs.

For what its worth here are python, C# and Java:

Python
datetime.min = 1-1-1 00:00:00
datetime.max = 9999-12-31 23:59:59

C#
DateTime.MinValue = 0001-01-01 12:00:00 AM
DateTime.MaxValue = 9999-12-31 11:59:59 PM

Java
LocalDateTime.MAX = 999999999-12-31T23:59:59.999999999
LocalDateTime.MIN = -999999999-01-01T00:00:00


I am in favour of either leaving it as is or making the max / min on 64 bit systems:
0001-01-01 00:00:00
9999-12-31 23:59:59

lanort:
Hmm, I did not know, that the min/max values are so different. Both your solutions sound good.

briannesbitt:
If you update the min value year to 1 I'll merge this.

CarsonF:
+1

CarsonF:
@briannesbitt good to go?

dereuromark:
We actually found an easier fix for all this: https://github.com/cakephp/chronos/pull/28/files

CarsonF:
@dereuromark That works, but the year is still negative which is confusing. I would rather have the min year be 1.

dereuromark:
OK, then this is different from fixing the min/max values on Win OS etc.

PR #354, opened 2015-04-26
Persian language added.

easteregg:
I've added persian language in the Locale directory. 
Giving the fact that I'm fairly new to Git community, it is not as clean as it should be.


JohnnyEvo:
You have to add ".idea" at your global .gitignore, not .gitignore project.

PR #361, opened 2015-05-14
Generate dates based on Google's Material design

rkrater:


briannesbitt:
Seems to be a narrow use case to put in the core?

PR #375, opened 2015-05-29
Use OS language instead of English

Formatink:
Use the default OS language instead of English


briannesbitt:
I get "Call to undefined function locale_get_default()" and I am on PHP 5.6.6.

PR #380, opened 2015-05-31
Dynamic first/last day of week and weekend days

dani3l:
In Israel, week starts at Sunday and ends at Saturday, weekend days are Friday and Saturday
and many other countries has their own weekends and weird calendars.. so I think it might help others

Added:
protected static $weekStartsAt integer (use Carbon::DAYCONST)
protected static $weekEndsAt integer (use Carbon::DAYCONST)
protected static $weekendDays array (array of Carbon::DAYCONST)
public static function getWeekStartsAt()
public static function setWeekStartsAt($day)
public static function getWeekEndsAt()
public static function setWeekEndsAt($day)
public static function getWeekendDays()
public static function setWeekendDays($days)

Updated:
isWeekday() - to use !$this->isWeekend()
isWeekend() - to check if current day is in static::$weekendDays
startOfWeek() - to use static::$weekStartsAt
endOfWeek() - to use static::$weekEndsAt

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -141,7 +141,7 @@ class Carbon extends DateTime
      *
      * @var array
      */
-    protected static $weekendDays = [self::SATURDAY, self::SUNDAY];
+    protected static $weekendDays = array(self::SATURDAY, self::SUNDAY);
 
     /**
      * A test Carbon instance to be returned when now instances are created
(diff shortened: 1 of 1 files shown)


dani3l:
If you accept this one, the next level would be to automate it with locales
Set the first/last days of week and weekend days inside a locale file and set them when calling setLocale() but then you'd have to use the full xx_XX locale convention to make it perfect

PR #383, opened 2015-06-02
Adding Microtime

bassrock:
The following adds microtome to the constructor.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -288,7 +288,14 @@ class Carbon extends DateTime
             $time = $testInstance->toDateTimeString();
         }
 
-        parent::__construct($time, static::safeCreateDateTimeZone($tz));
+        $timezone = static::safeCreateDateTimeZone($tz);
+        if (is_null($time) || $time === 'now') {
+            $dateTime = new DateTime('now', $timezone);
+            $micro = preg_replace('/0\.(\d{6}).*/', '$1', microtime());
+            $time = $dateTime->format(static::DEFAULT_TO_STRING_FORMAT).'.'.$micro;
+        }
+
+        parent::__construct($time, $timezone);
     }
 
     /**
--- a/tests/Carbon/GettersTest.php
+++ b/tests/Carbon/GettersTest.php
@@ -80,7 +80,9 @@ class GettersTest extends AbstractTestCase
         }
 
         $d = Carbon::now();
-        $this->assertSame(0, $d->micro);
+        $aMicrotime = explode(' ', microtime());
+        $aMicrotime = explode('.', $aMicrotime[0]);
+        $this->assertGreaterThan($d->micro, $aMicrotime[1]);
     }
 
     public function testDayOfWeeGetter()


aziraphale:
Please? :) I'd love for this to be implemented - it's currently really messy and needlessly complicated to create/display a date including microseconds, so having Carbon doing the heavy lifting for me would be awesome.

My only potential concern is regarding performance - the extra work involved in grabbing the microsecond value might be quite high, especially when the majority of uses of Carbon (and date/time comparisons in general) really don't care about microseconds.

Maybe it could be implemented by way of a `Carbon::nowWithMicroseconds()` method? It's only ever when you request "now" that the extra microseconds code involved in this PR is required (the alternatives being parsing existing dates/times that may or may not include microseconds - but certainly don't care about the *current* microsecond value), so I think it's perfectly reasonable to have a separate method that gives us "now" including microseconds.

Feel free to point out everywhere I'm wrong, of course! ;)

dani3l:
microtime would be freaking awesome
but this implementation only solves current microtime (and as far as I know it's impossible with php's DateTime)

it's also a shame that you can't fetch microseconds with mysqlnd yet :(

lucasmichot:
@bassrock Can you rebase and squash this PR ?

bassrock:
@lucasmichot Rebased and squashed!

bassrock:
Interesting so some of the tests where there is an assert on time equals fail because the micro times don't match, but on my local system and most of the test environments they pass. Any thoughts?

src/Carbon/Carbon.php
```diff
@@ -288,7 +288,16 @@ public function __construct($time = null, $tz = null)
             $time = $testInstance->toDateTimeString();
         }
 
-        parent::__construct($time, static::safeCreateDateTimeZone($tz));
+        if (is_null($time) || $time === 'now') {
```
lucasmichot: Why not something shorter like:
```php
if (is_null($time) || $time === 'now') {
    $dateTime = new DateTime('now', static::safeCreateDateTimeZone($tz));
    $micro = preg_replace('/0\.(\d{6}).*/', '$1', microtime());
    $time = $dateTime->format(static::DEFAULT_TO_STRING_FORMAT).'.'.$micro;
}

parent::__construct($time, static::safeCreateDateTimeZone($tz));
```

bassrock:
@lucasmichot Made the suggested change.

src/Carbon/Carbon.php
```diff
@@ -288,6 +288,12 @@ public function __construct($time = null, $tz = null)
             $time = $testInstance->toDateTimeString();
         }
 
+        if (is_null($time) || $time === 'now') {
+            $dateTime = new \DateTime('now', static::safeCreateDateTimeZone($tz));
```
lucasmichot: `\DateTime` is already imported so `new DateTime` is just fine

src/Carbon/Carbon.php
```diff
@@ -288,6 +288,12 @@ public function __construct($time = null, $tz = null)
...
+            $micro = preg_replace('/0\.(\d{6}).*/', '$1', microtime());
+            $time = $dateTime->format(static::DEFAULT_TO_STRING_FORMAT).'.'.$micro;
+        }
+
         parent::__construct($time, static::safeCreateDateTimeZone($tz));
```
lucasmichot: `static::safeCreateDateTimeZone($tz)` is calculated twice, this can be extracted into a variable

tests/Carbon/GettersTest.php
```diff
@@ -80,7 +80,9 @@ public function testMicroGetterWithDefaultNow()
         }
 
         $d = Carbon::now();
-        $this->assertSame(0, $d->micro);
+        $aMicrotime = explode(' ', microtime());
+        $aMicrotime = explode('.', $aMicrotime[0]);
+        $this->assertGreaterThan($d->micro, $aMicrotime[1]);
```
lucasmichot: For some reason some tests is failing only with PHP 7.0.x...

lucasmichot:
@bassrock I've fixed some failing tests, can you rebase again ?

src/Carbon/Carbon.php
lucasmichot: Ping @bassrock

bassrock:
@lucasmichot Rebased and fixed the variable

Glavic:
Is this still needed since >=7.1.x  versions already have microseconds part...

kylekatarnls:
Hi, it depends. I see @lucasmichot changed 1.x milestone to 2.x. But it's because 2.x was expected very sooner. So we may reconsider the milestone of this issue.

If 2.x (PHP 7.1+ as probable Carbon compatibility), this PR should not be merged.
If 1.x (the hack should be triggered only if PHP_VERSION < 7.1, we should not hack PHP "now" to put microseconds if already available natively).

kylekatarnls:
new DateTime neither is not compatible, now we have to handle setTestNow. I will try to propose something.

kylekatarnls:
This is my counter-proposal: #1140 Tests passed locally on different PHP versions, I just have to check why some fail on Travis but it should not be too bad.

This one could be merged it in a 1.x release then dropped when we'll drop PHP < 7.1

PR #385, opened 2015-06-02
Update zh-TW.php

dha0stw:
fix term usage


wangchj:
I don't think this modification is correct and should not be merged.

briannesbitt:
I have no idea !! Let me know if this is needed.

dha0stw:
@briannesbitt I think about this for a while, i'ts no need.
sorry to make you confused.
I should use  'ago' and 'from_now' to get my usage.

PR #390, opened 2015-06-20
Provide a toDatetime function

lucasmichot:
For a reason we might want to retrieve the base DateTime from a Carbon instance

### DIFF
--- a/tests/DateTimeTest.php
+++ b/tests/DateTimeTest.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Carbon package.
+ *
+ * (c) Brian Nesbitt <brian@nesbot.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Carbon\Carbon;
+
+class DateTimeTest extends TestFixture
+{
+    public function testToDateTime()
+    {
+        $c = Carbon::now()->startOfYear();
+        $d = $c->toDatetime();
+
+        $this->assertInstanceOf('\DateTime', $d);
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Indeed, +1

wangchj:
Carbon object is an instance of DateTime

lucasmichot:
@wangchj 
> Carbon object is an instance of DateTime

Datetime object is not an instance of Carbon

GrahamCampbell:
:-1: I don't understand why you'd want to do this?

lucasmichot:
Because we might want to get the parent object from a Carbon

GrahamCampbell:
Why?

tests/DateTimeTest.php
```diff
@@ -0,0 +1,28 @@
...
+use Carbon\Carbon;
+
+class DateTimeTest extends TestFixture
+{
+    public function testToDateTime()
+    {
+        $c = Carbon::now()->startOfYear();
+        $d = $c->toDatetime();
+
+        $this->assertInstanceOf('\DateTime', $d);
```
GrahamCampbell: Leading slashes are incorrect.

lucasmichot:
Because we might want a Datetime without $days property

lucasmichot:
(As an example)

GrahamCampbell:
I still have no idea why you'd want that?

kylekatarnls:
@lucasmichot $days is a static property so it should not bother on an instance.
As a other example (maybe not better) it could be used to get the "pure" value, if you have a Carbon or ChildOfCarbon, SubChildOfCarbon, etc. object, ->toDatetime() will allow you to extract the date with no wrapper.

PR #395, opened 2015-06-24
Convenience methods isMonday(), ... isSunday() 

wangchj:


src/Carbon/Carbon.php
```diff
@@ -1278,6 +1278,69 @@ public function isSameDay(Carbon $dt)
         return $this->toDateString() === $dt->toDateString();
     }
 
+    /**
+     * Checks if this day is a Sunday.
+     * @return boolean
```
GrahamCampbell: Missing a blank line before the return annotation.

tests/IsTest.php
```diff
@@ -130,4 +130,138 @@ public function testIsSameDayFalse()
         $current = Carbon::createFromDate(2012, 1, 2);
         $this->assertFalse($current->isSameDay(Carbon::createFromDate(2012, 1, 3)));
     }
+
+    public function testIsSunday()
+    {
+        //True in the past past
```
GrahamCampbell: missing a space after the `//` in every case

PR #400, opened 2015-07-03
Use the OS language instead of English

Formatink:
Use the default OS language instead of English if the intl extension is loaded

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -885,11 +885,13 @@ class Carbon extends DateTime
     protected static function translator()
     {
         if (static::$translator == null) {
-            static::$translator = new Translator('en');
+            // If intl extension is loaded, the OS language is used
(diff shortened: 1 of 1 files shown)


src/Carbon/Carbon.php
```diff
@@ -885,11 +885,13 @@ public static function hasRelativeKeywords($time)
...
-            static::$translator = new Translator('en');
+            // If intl extension is loaded, the OS language is used
+            $defaultLocaleLanguage = extension_loaded('intl') ? substr(locale_get_default(), 0, 2) : 'en';
+            
+            static::$translator = new Translator($defaultLocaleLanguage);
             static::$translator->addLoader('array', new ArrayLoader());
-            static::setLocale('en');
+            static::setLocale($defaultLocaleLanguage);
         }
-
```
GrahamCampbell: no

Formatink: @GrahamCampbell I close and I pull request again ? :/

PR #401, opened 2015-07-03
Use the OS language instead of English

Formatink:
Use the default OS language instead of English if the INTL extension is loaded.

If the user's OS has the INTL extension, Carbon will use the intl.default_locale variable in the PHP configuration as default language.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -885,9 +885,12 @@ class Carbon extends DateTime
     protected static function translator()
     {
         if (static::$translator == null) {
-            static::$translator = new Translator('en');
+            // If intl extension is loaded, the OS language is used
+            $defaultLocaleLanguage = extension_loaded('intl') ? substr(locale_get_default(), 0, 2) : 'en';
+            
+            static::$translator = new Translator($defaultLocaleLanguage);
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I don't think using the OS language makes sense.

Formatink:
If the user's OS has the INTL extension, Carbon will use the intl.default_locale variable in the PHP configuration for the language. Users who have this extension (there are many) will no longer have to use a method like ->setLocale('fr') before display the date ... It's very sense and better than forcing users to use English **by default** :] (They can change if they wish)

dereuromark:
:+1: I agree, sounds quite useful to default to the actual "current" default.

kylekatarnls:
Good idea. But it should not be merge in a minor version since it can break english apps running on OS in another language.

briannesbitt:
Ya... BC change.

GrahamCampbell:
I'm :-1: anyway.

kylekatarnls:
I propose instead a `Carbon::setLocale('auto');` to get the `LC_TIME` and set it as Carbon locale. So you can align `formatLocalized` and `diffForHumans` by running it. It will try to detect the closest available locale.

See #1336

PR #414, opened 2015-08-03
Change "occurence" to "occurrence" in Carbon.php

arkeidolon:
All of them, in the PHPdocs


caycek:
I fixed this in my version as well. I didn't feel like making a pull so I salute you!

PR #424, opened 2015-08-17
Add test for add and sub weekdays during weekend

cl4rke:
Test for attempts to add or subtract weekdays on a weekend. For when someone tries to make an attempt to change add and sub weekends functionality (planning on doing this).


briannesbitt:
For the rest of this fix see d02c4136c3300e5044226e5be8fcfc5b4b6fc272

Issue #406, opened 2015-07-10
Carbon fork that implements DateTimeInterface and extends both DateTime and DateTimeImmutable


jan-j:
Hi. I created fork of this library that use DateTimeInterface and extends both DateTime and DateTimeImmutable.

https://github.com/meebio/Carbon
#### Here is a small description:

[Carbon](https://github.com/briannesbitt/Carbon) library fork that extends both DateTime and DateTimeImmutable extensions with shared interface.

Original Carbon class is broken into CarbonInterface (interface extending DateTimeInterface) and CarbonTrait that implements that interface. There are 2 usable classes Carbon extending DateTime (that has exactly the same functionality as original Carbon class) and CarbonImmutable extending DateTimeImmutable. Both classes implement CarbonInterface and use CarbonTrait for common methods.
#### Summary:
- Fork was done at 1.20.0 release.
- No additional features were added.
- All original unit tests are passed both by mutable and immutable implementation. I had to modify some tests for immutable class to properly work.
- All tests that were applied to original Carbon class, have been duplicated and now this tests are applied to both mutable and immutable implementation. There should be better way to do this because there is a lot code duplication. I tried to find a way to use one code to test both implementations, but I couldn't find anything useful. Anyone know better way to organize it? 
- There are only few sample unit tests that checks mutability and immutability, that should be improved.
- DateTimeInterface and DateTimeImmutable are available for PHP >= 5.5.0 only and this is biggest issue here.

What are your thoughts about it? What can be improved?

Related issues: #103 #2 #233

dereuromark:
You should allow issues on your fork IMO.
My question would be: Should it also have a bool `isMutable()` check for both classes to detect at runtime whether it is mutable or not? Would in same cases help for dynamically created objects to know what course of action to take.

jan-j:
@dereuromark 
I enabled issues on my fork. Thanks for pointing that out :)

Method `isMutable()` totally makes sense, `isImmutable()` can be added too.

I didn't mention that, but there are two conversion methods added to one for each implementation: CarbonMutable has `toImmutable()` and CarbonImmutable has `toMutable()`.

dereuromark:
I think having one with boolean result might suffice :)

ektarum:
hi all,

hi bump this issues. it's possible to merge advance in this subject?

Thanks in advance.

jan-j:
@ektarum Hi
[...]

PR #426, opened 2015-08-22
DateTimeInterface implementation with two variations: extension of DateTime and DateTimeImmutable.

jan-j:
Original Carbon class is broken into CarbonInterface (interface extending DateTimeInterface) and CarbonTrait that implements that interface. There are 2 usable classes Carbon extending DateTime (that has exactly the same functionality as original Carbon class) and CarbonImmutable extending DateTimeImmutable. Both classes implement CarbonInterface and use CarbonTrait for common methods.

Backwards compatibility is kept, but only for versions of PHP greater than or equal to 5.5.0. Earlier versions are unsupported, because of DateTimeInterface and DateTimeImmutable, which were introduced in PHP 5.5 unfortunately. All original tests for mutable version of Carbon are passed.

This is only proposal and I know it needs to be discussed thoroughly.

For other info see issue #406.


briannesbitt:
The general idea here is good but needs to obviously be in a v2.

morloderex:
Any new regarding this?

IonutBajescu:
@briannesbitt there doesn't seem to be any breaking change at first sight though. Wouldn't that warrant a minor version bump at most?

jan-j:
@IonutBajescu It does use DateTimeInterface and DateTimeImmutable which is supported from PHP version 5.5.0 and it is breaking change if someone is using lower version. Although without strictly following semver, only minor version bump is not that terrible idea. Extra constraint can be added for PHP version to composer.json file to make sure, it's not going to be installed on lower versions silently.

kylekatarnls:
Thanks, it will be done with traits to avoid duplicate code.

Fixed via #1315

Issue #369, opened 2015-05-25
Creation of a 0 interval time


sitawit:
### Suggest
To be able to create a time interval of **0**.

So the following code would work.
```php
CarbonInterval::months(0);   // 0 months
CarbonInterval::hours(0);    // 0 hours
```

### Current version
So far, CarbonInterval can't create a 0 interval of any period.

For example:
```php
CarbonInterval::months(0); 
CarbonInterval::hours(0); 
```

Will result in 

```php
Exception
DateInterval::__construct() [<a href='0dateinterval.construct0'>dateinterval.construct0</a>]: Unknown or bad format (P)
```

### Background
At first, it seems make no sense to create a 0 interval because in fact, it doesn't contain any interval. However, when it comes to a situation where you create a set of time intervals. And the first slot start from 00:00. So in this case, you have to create this 0 interval using other method rather than CarbonInterval. In order to make your code consistency, using CarbonInterval for all the time in the set would be great. And you don't have to create a logic where it catches an exception of 0 interval.

dereuromark:
Indeed: Sounds useful to support that. Especially with dynamic data (>= 0) coming in.

Sverri:
When the constructor checks the parameter it uses the expression, `$months > 0`. It looks like this is just a simple mistake that can be fixed by simply changing it to `$months > -1`. PHP's  `DateInterval` allows a "zero interval", i.e. something like `P0M`, so it should not be a problem (unless I am missing something about how `CarbonInterval` works).

jarnovanleeuwen:
https://github.com/briannesbitt/Carbon/pull/437 should fix this.

briannesbitt:
Merged #437

PR #437, opened 2015-09-16
Allow for creation of a zero interval.

jarnovanleeuwen:
Fixes issue https://github.com/briannesbitt/Carbon/issues/369. This also allows us to construct a new interval using method chaining, where the first method contains a zero timestamp. Example:

    CarbonInterval::days(0)->weeks(1);

Jarno


dereuromark:
Looks like a test case for it is missing.

jarnovanleeuwen:
Well, the current test case is testing the opposite behavior, that it should not allow the zero interval. That is why it is currently failing.

Before changing this test case I would like to know the community's thoughts on this. PHP's DateInterval does allow the zero interval (even though the construction is a little bit awkward).

dbaeck:
This functionality sounds reasonable to me, any news on why CarbonInterval does not allow this currently?

briannesbitt:
Merged and updated test.

PR #438, opened 2015-09-16
Fix: createFromTimestamp() vs createFromTimeStamp()

localheinz:
This PR

* [x] fixes a small thing in `README.md` where it is proposed to use `Carbon::createFromTimeStamp()`, but actually, it is `Carbon::createFromTimestamp()`


PSmolic:
Nice! And documentation could use update too: http://carbon.nesbot.com/docs/

PR #453, opened 2015-10-16
Composer will use the caret operator when using requiring a package.

NoUseFreak:
Composer will use the caret sign in newer versions.

```bash
$ composer require nesbot/carbon
Using version ^1.20 for nesbot/carbon
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing nesbot/carbon (1.20.0)
    Downloading: 100%         
```

```bash
$ cat composer.json 
{
    "require": {
        "nesbot/carbon": "^1.20"
    }
}
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -52,7 +52,7 @@ $ composer require nesbot/carbon
 ```json
 {
     "require": {
-        "nesbot/carbon": "~1.14"
+        "nesbot/carbon": "^1.20"
     }
 }
 ```


briannesbitt:
According to my understanding of https://getcomposer.org/doc/articles/versions.md I think the way it is fine ?!?

GrahamCampbell:
Yeh, this doesn't need changing. I agree with @briannesbitt.

briannesbitt:
Confirmed @GrahamCampbell ... closing.

NoUseFreak:
No problem. I agree this will act the same but the documentation https://getcomposer.org/doc/articles/versions.md#caret states `This is the recommended operator for maximum interoperability when writing library code.`

PR #456, opened 2015-10-31
Update fa.php

mahdi-younesi:
I just added a few more translations.


briannesbitt:
Currently the months and days aren't used.  Did you change ago, from_now, after or before?  Its hard for me to tell since your also changed the order.

PR #457, opened 2015-11-01
CarbonRange

20TRIES:
Hey, i needed a way of describing a date range in an application i was writing and thought it may be a useful extension here.

Its not finished, so there still needs to be some more testing and the code could be a little clearer/clearer; however i wanted to see if you thought it may be a useful addition before going all out with the idea.

It would allow stuff like the following:

`$week_start = Carbon::now()->startOfWeek();`

`$week_end = Carbon::now()->endOfWeek();`

`$date_range = CarbonRange::between($week_start, $week_end);`

`$date_range->forHuman();`

With a result of "This Week"

### DIFF
--- a/src/Carbon/CarbonRange.php
+++ b/src/Carbon/CarbonRange.php
@@ -1,0 +1,330 @@
+<?php namespace Carbon;
+
+/**
+ * An object for interacting with a date range.
+ *
+ * @package Carbon
+ * @author Marcus T <marcust261@icloud.com>
+ */
+class CarbonRange
+{
+    /**
+     * @var string
+     */
+    protected $name;
+
+    /**
+     * @var Carbon
+     */
+    protected $after;
+
+    /**
+     * @var Carbon
+     */
+    protected $before;
+
+    /**
+     * @var string A time period
+     */
+    const DAY = 'Day';
+
+    /**
+     * @var string A time period
+     */
+    const WEEK = 'Week';
+
+    /**
+     * @var string A time period
+     */
+    const MONTH = 'Month';
+
+    /**
+     * @var string A time period
+     */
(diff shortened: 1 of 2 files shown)


dereuromark:
What is the difference of Range vs the existing Interval?

20TRIES:
I was thinking that while the interval is say, 3 months. The range would have a "context", so it would have a start and end date.

I haven't used the interval class too much in the past; maybe if the implementation is very similar this would be better as an extension of the interval class instead?

src/Carbon/CarbonRange.php
```diff
@@ -0,0 +1,330 @@
+<?php namespace Carbon;
+
+/**
+ * An object for interacting with a date range.
+ *
+ * @package Carbon
```
GrahamCampbell: this should be ommited

src/Carbon/CarbonRange.php
```diff
@@ -0,0 +1,330 @@
...
+ * @author Marcus T <marcust261@icloud.com>
```
GrahamCampbell: this should be ommited

src/Carbon/CarbonRange.php
```diff
@@ -0,0 +1,330 @@
+<?php namespace Carbon;
```
GrahamCampbell: incorrect cs

GrahamCampbell: should be:

```php
<?php
/*
 * This file is part of the Carbon package.
 *
 * (c) Brian Nesbitt <brian@nesbot.com>
 *
 * For the full copyright and license information, please view the LICENSE
 */
namespace Carbon;

```

PR #461, opened 2015-11-03
Set PSR-4 autoloading

lucasmichot:
For `/src` and `/tests`
* also separate `Carbon` and `CarbonInterval` tests into distinct folders (`tests/Carbon` & `tests/CarbonInterval`
* Let the `TestFixture` consequently be an abstract test case
* Give `MyCarbon` class its own file within a `Fixtures` directory (from `StringsTest.php` to `tests/Carbon/Fixtures/MyCarbon.php`)
* Move source files and `Lang/` folder from `src/Carbon/` to `src/`


vinkla:
:+1:

lucasmichot:
(rebased)

lucasmichot:
(rebased)

lucasmichot:
rebased and green @briannesbitt

lucasmichot:
(rebased)

briannesbitt:
Sorry.... I kept skipping over this one !

GrahamCampbell:
This PR also affects all others too.

GrahamCampbell:
This also will break stuff for people not using composer.

lucasmichot:
Might be easier to check locally

GrahamCampbell:
I'd suggest not moving the contents of "src" around until 2.x.

briannesbitt:
Agreed.

lucasmichot:
> This also will break stuff for people not using composer.
Not the argument I would keep IMO

GrahamCampbell:
We should still move to PSR4, just without moving the stuff in src.

PR #466, opened 2015-11-04
Updated the StyleCI badge styling

GrahamCampbell:


briannesbitt:
Why are the badge tags so inconsistent !  :+1:

GrahamCampbell:
IMO, everyone should use StyleCI's default styling coz it's so pretty, but they don't, so StyleCI provides a flag you can use to make it look like the others. :)

PR #476, opened 2015-11-04
Fix another docblock

lucasmichot:
Fix another docblock


src/Carbon/Carbon.php
```diff
@@ -2487,7 +2487,7 @@ public function nthOfMonth($nth, $dayOfWeek)
      *
      * @param int $dayOfWeek
```
GrahamCampbell: just noticed this should be int|null too

PR #480, opened 2015-11-04
Also test on php 5.3.3

GrahamCampbell:


briannesbitt:
Why specifically 5.3.3 ?

GrahamCampbell:
Well, you say you;re compatible with 5.3.0, and 5.3.3 has the best travis has to offer to check that is actually true.

briannesbitt:
Ya I guess that should be the best of 5.3 (ie. 5.3.29 which is what travis uses) but I see what you are saying.  Anyway v2 should be 5.5+

PR #486, opened 2015-11-04
Remove usless brackets

lucasmichot:
Remove usless brackets

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -352,17 +352,17 @@ class Carbon extends DateTime
      */
     public static function create($year = null, $month = null, $day = null, $hour = null, $minute = null, $second = null, $tz = null)
     {
-        $year = ($year === null) ? date('Y') : $year;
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -2527,7 +2527,7 @@ public function nthOfQuarter($nth, $dayOfWeek)
         $year = $dt->year;
         $dt->firstOfQuarter()->modify('+'.$nth.' '.static::$days[$dayOfWeek]);
 
-        return ($lastMonth < $dt->month || $year !== $dt->year) ? false : $this->modify($dt);
+        return $lastMonth < $dt->month || $year !== $dt->year ? false : $this->modify($dt);
```
briannesbitt: This is the only one that I find loses readability when removing the brackets.

lucasmichot: OK reverting this one then

lucasmichot: Done

PR #488, opened 2015-11-04
Add the cast fixer

lucasmichot:
Add another fixer

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -22,6 +22,7 @@ $fixers = array(
     'phpdoc_types',
     'phpdoc_var_without_name',
     'return',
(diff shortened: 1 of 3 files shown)


briannesbitt:
Not a personal fan but ok !

dereuromark:
Me neither :D
Casts belong to the code itself they cast.

lucasmichot:
Well in fact I don't have any opinion on that, but all casts were separated by a space in the code but this one ;)

dereuromark:
php-cs-fixer definitely needs a NoSpaceCastFixer as contrib one :)
But good point, so far this is consistent then now.

PR #497, opened 2015-11-09
Update cs.php

Tlapi:
This is the proper translation.


lucasmichot:
@Tlapi can you fix your code with https://styleci.io/analyses/qyVa5z ?

Tlapi:
@lucasmichot Well, i can. But I've just made fork and edited only cs.php, i did not manipulated src/Carbon/Carbon.php nither src/Carbon/CarbonInterval.php. And those are the files not passing StyleCI.

lucasmichot:
Strange indeed @Tlapi , I was suspecting an old rebase and thus a rebase would rebase would have been required.
Also previously merged PR #496 was respecting the standard
Any idea @GrahamCampbell ?

PR #500, opened 2015-11-10
Simplify calls to parent functions

lucasmichot:
Simplify calls to parent functions

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -761,9 +761,7 @@ class Carbon extends DateTime
      */
     public function setTimezone($value)
     {
-        parent::setTimezone(static::safeCreateDateTimeZone($value));
-
-        return $this;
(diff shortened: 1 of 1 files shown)


dereuromark:
The actual code coverage is probably more around 97%
The lang files should probably not count in too much :)

lucasmichot:
Indeed @dereuromark / currently looking at the codecov settings to fix that ;-)

lucasmichot:
But weel in facts, the localizations should be covered !!!! :-D
Which is not the case for most of them ATM

lucasmichot:
@dereuromark you were real close ;-) : 98.23 %
https://github.com/briannesbitt/Carbon/pull/501#issuecomment-155408089

dereuromark:
@lucasmichot Well, I got my number from https://coveralls.io/github/cakephp/chronos which is pretty much 100% API-consistent.

PR #501, opened 2015-11-10
Reorganize localizations tests

lucasmichot:
Reorganize localizations tests.

`\Tests\Carbon\LocalizationTest` takes care of checking file presence, setting locale and setting translator

Each language has its own test file in `Tests\Localization` namespace.


briannesbitt:
Nice ;). Now adding a badge won't be so embarrassing, haha

lucasmichot:
Ah ah ah ! Yes

briannesbitt:
Maybe we should actually run them! d07c1de684af598244739f7f56aadc51bff620a3

PR #506, opened 2015-11-10
Avoid calling __set many times

lucasmichot:
Avoid calling __set many times


GrahamCampbell:
These coverage comments are so annoying. Can they not be disabled?

lucasmichot:
I suppose they can as long as the codecov/project PR status is available

briannesbitt:
I limited the comment to just the title.

I think you can always see the results on their site though..  https://codecov.io/github/briannesbitt/Carbon

PR #507, opened 2015-11-10
Fix StyleCI for now

GrahamCampbell:


briannesbitt:
My inbox thanks you !

PR #517, opened 2015-11-13
Add header fixing

lucasmichot:
For StyleCI, this can be added via the web interface.

### DIFF
--- a/src/Carbon/Lang/de.php
+++ b/src/Carbon/Lang/de.php
@@ -3,11 +3,10 @@
 /*
  * This file is part of the Carbon package.
  *
- * (c) Michael Hohl <me@michaelhohl.net>
+ * (c) Brian Nesbitt <brian@nesbot.com>
  *
- * This file is released under the terms of CC0.
- * CC0 is even more permissive than the MIT license, allowing you to use the code in
- * any manner you want, without any copyright headers, notices, or other attribution.
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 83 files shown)


GrahamCampbell:
NB, @briannesbitt, you can get StyleCI to do this too from the setting page on the web interface.

src/Carbon/Lang/de.php
```diff
@@ -3,11 +3,10 @@
 /*
  * This file is part of the Carbon package.
  *
- * (c) Michael Hohl <me@michaelhohl.net>
```
OskarStark: is this correct?

lucasmichot: Yes. If there are support or issues, people shall contact @briannesbitt and not the committer of this files or change.

GrahamCampbell: > is this correct?

Yes, all files are copyright Brian, even if he didn't author them.

OskarStark: :+1: thanks for the feedback

briannesbitt:
This duplicates the header in the test files... 
https://github.com/lucasmichot/Carbon/blob/feature/master.phpcs-header/tests/Carbon/AddTest.php

lucasmichot:
Fixed @briannesbitt 
(Namespaces decarations had to be moved under the existing file docblock)

briannesbitt:
k cool

Issue #514, opened 2015-11-13
Feature request: Sleeping in a test


rrehbeindoi:
While unit testing a class that includes sleeps, could Carbon include a shim for the sleep method?

```php
Carbon::setTestNow(Carbon::createFromDate(2000, 1, 1));
Carbon::sleep(5); // Modifies $testNow to +5 seconds, does not actually call php's sleep

Carbon::setTimeNow();
Carbon::sleep(5); // call php's sleep
```

PR #518, opened 2015-11-13
Added fake sleep to Carbon

rrehbeindoi:
Adds feature of fake sleeping from my request in #514

### DIFF
--- a/src/Carbon/CarbonSleeper.php
+++ b/src/Carbon/CarbonSleeper.php
@@ -34,15 +34,15 @@ class CarbonSleeper
      */
     public static function sleep($seconds)
     {
+        $seconds = (int) $seconds;
+
+        if ($seconds < 0) {
+            throw new InvalidArgumentException(
+                'sleep(): Number of seconds must be greater than or equal to 0'
+            );
+        }
+
         if (Carbon::hasTestNow()) {
-            $seconds = (int) $seconds;
-
-            if ($seconds < 0) {
-                throw new InvalidArgumentException(
-                    'sleep(): Number of seconds must be greater than or equal to 0'
-                );
-            }
-
             Carbon::setTestNow(Carbon::getTestNow()->modify("+{$seconds} seconds"));
 
             return 0;
--- a/src/Carbon/CarbonSleeper.php
+++ b/src/Carbon/CarbonSleeper.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Carbon package.
+ *
+ * (c) Brian Nesbitt <brian@nesbot.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Carbon;
+
+/**
+ * A shim between system sleep and Carbon
+ */
+class CarbonSleeper
+{
+    /**
+     * Delay execution
+     *
+     * When Carbon's testing time is set, it will fake sleeping
+     *
+     * @param int $seconds Halt time in seconds
+     *
+     * @return int
+     */
+    public static function sleep($seconds)
+    {
+        if (Carbon::hasTestNow()) {
+            $seconds = intval($seconds);
+
+            if (0 > $seconds) {
+                trigger_error('sleep(): Number of seconds must be greater than or equal to 0', \E_USER_WARNING);
+
+                return false;
+            }
+
+            Carbon::setTestNow(Carbon::getTestNow()->modify("+{$seconds} seconds"));
+
+            return 0;
+        }
+
+        return sleep($seconds);
+    }
+}
--- a/src/Carbon/CarbonSleeper.php
+++ b/src/Carbon/CarbonSleeper.php
@@ -11,6 +11,8 @@
 
 namespace Carbon;
 
+use InvalidArgumentException;
+
 /**
  * A shim between system sleep and Carbon
  */
@@ -23,17 +25,19 @@ class CarbonSleeper
      *
      * @param int $seconds Halt time in seconds
      *
+     * @throws InvalidArgumentException
+     *
      * @return int
      */
     public static function sleep($seconds)
     {
         if (Carbon::hasTestNow()) {
-            $seconds = intval($seconds);
+            $seconds = (int)$seconds;
 
-            if (0 > $seconds) {
-                trigger_error('sleep(): Number of seconds must be greater than or equal to 0', \E_USER_WARNING);
-
-                return false;
+            if ($seconds < 0) {
+                throw new InvalidArgumentException(
+                    'sleep(): Number of seconds must be greater than or equal to 0'
+                );
             }
 
             Carbon::setTestNow(Carbon::getTestNow()->modify("+{$seconds} seconds"));
--- a/src/Carbon/CarbonSleeper.php
+++ b/src/Carbon/CarbonSleeper.php
@@ -23,6 +23,9 @@ class CarbonSleeper
      *
      * When Carbon's testing time is set, it will fake sleeping
      *
+     * Like PHP's sleep function, this will accept zero without error. Unlike
+     * PHP's sleep function, negative sleep will throw an exception.
+     *
      * @param int $seconds Halt time in seconds
      *
      * @throws InvalidArgumentException
@@ -32,7 +35,7 @@ class CarbonSleeper
     public static function sleep($seconds)
     {
         if (Carbon::hasTestNow()) {
-            $seconds = (int)$seconds;
+            $seconds = (int) $seconds;
 
             if ($seconds < 0) {
                 throw new InvalidArgumentException(
--- a/tests/CarbonSleeper/SleepTest.php
+++ b/tests/CarbonSleeper/SleepTest.php
@@ -34,8 +34,7 @@ class SleepTest extends AbstractTestCase
 
                 $self->assertLessThanOrEqual(
                     1, // Allow up to 1 second in case of timing issue between calls
-                    $after - $before,
-                    'Carbon sleep called system sleep in test condition'
+                    $after - $before
                 );
             }
         );
@@ -52,11 +51,7 @@ class SleepTest extends AbstractTestCase
 
                 $after = Carbon::now()->getTimestamp();
 
-                $self->assertEquals(
-                    5,
-                    $after - $before,
-                    'Carbon sleep called system sleep in test condition'
-                );
+                $self->assertEquals(5, $after - $before);
             }
         );
     }
@@ -83,6 +78,6 @@ class SleepTest extends AbstractTestCase
 
         $after = time();
 
-        $this->assertNotEquals($after, $before, 'Sleep did not delay execution');
+        $this->assertNotEquals($after, $before);
     }
 }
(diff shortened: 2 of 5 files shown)


dereuromark:
Is that method needed in the core class itself? It has nothing to do with the carbon functionality IMO.

rrehbeindoi:
It could be split off.  Would Carbon\CarbonSleeper be a better place?

src/Carbon/CarbonSleeper.php
```diff
@@ -0,0 +1,46 @@
...
+     * @param int $seconds Halt time in seconds
+     *
+     * @return int
+     */
+    public static function sleep($seconds)
+    {
+        if (Carbon::hasTestNow()) {
+            $seconds = intval($seconds);
+
+            if (0 > $seconds) {
```
lucasmichot: We don't use yoda conditions, please use ` $seconds < ...`

lucasmichot: Does it make sense to make `CarbonSleeper` with `seconds = 0` ?
I don't think so, I would do : `if ($seconds < 1) {` instead

src/Carbon/CarbonSleeper.php
```diff
@@ -0,0 +1,46 @@
...
+                trigger_error('sleep(): Number of seconds must be greater than or equal to 0', \E_USER_WARNING);
```
lucasmichot: Better raise `\InvalidArgumentException` here and do not return a boolean

src/Carbon/CarbonSleeper.php
```diff
@@ -0,0 +1,46 @@
...
+
+                return false;
```
lucasmichot: `@return int`

src/Carbon/CarbonSleeper.php
rrehbeindoi: The error and return match up with php.net/sleep, which does accept zero seconds without error.

src/Carbon/CarbonSleeper.php
```diff
@@ -0,0 +1,46 @@
...
+     * When Carbon's testing time is set, it will fake sleeping
+     *
+     * @param int $seconds Halt time in seconds
+     *
+     * @return int
+     */
+    public static function sleep($seconds)
+    {
+        if (Carbon::hasTestNow()) {
+            $seconds = intval($seconds);
```
dereuromark: I'd usually go for the (int) cast.

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+namespace Tests\Carbon;
```
lucasmichot: This should be put in the `tests/CarbonSleeper` folder and the filename should be `Tests\CarbonSleeper\SleepTest`

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,85 @@
...
+ * file that was distributed with this source code.
+ */
+
+use Carbon\Carbon;
+use Carbon\CarbonSleeper;
+use Tests\AbstractTestCase;
+
+class SleepTest extends AbstractTestCase
+{
+    protected function tearDown()
```
lucasmichot: Avoid it, please wrap your tests with `\Tests\AbstractTestCase::wrapWithTestNow`.
See `\Tests\Localization\FrTest::testDiffForHumansLocalizedInFrench` for a good example

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,85 @@
...
+        CarbonSleeper::sleep(5);
+
+        $after = time();
+
+        $carbonAfter = Carbon::now();
+
+        $this->assertLessThanOrEqual(
+            1, // Allow up to 1 second in case of timing issue between calls
+            $after - $before,
+            'Carbon sleep called system sleep in test condition'
```
lucasmichot: We never pass messages to assert... functions

tests/CarbonSleeper/SleepTest.php
rrehbeindoi: The test that throws an exception leaves side effects due to wrapWithTestNow cleanup being bypassed by a throw.

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->assertEquals(
+            5,
+            $after - $before,
+            'Carbon sleep called system sleep in test condition'
+        );
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
+     * @expectedMessage Number of seconds must be greater than or equal to 0
```
lucasmichot: Can be safely removed, we only test the exception type

tests/CarbonSleeper/SleepTest.php
rrehbeindoi: Is there a reason not to?

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $this->assertEquals(
+            5,
+            $after - $before,
+            'Carbon sleep called system sleep in test condition'
+        );
+    }
+
+    /**
+     * @expectedException InvalidArgumentException
```
lucasmichot: Please use teh FQN : `@expectedException \InvalidArgumentException`

tests/CarbonSleeper/SleepTest.php
lucasmichot: Interesting point @rrehbeindoi , probably some involvments to add to wrapWithTestNow then. You cna leave it as it for now then

tests/CarbonSleeper/SleepTest.php
lucasmichot: We prefer all tests to be the same, and the testFunction name is self-explanatory

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,85 @@
...
+        Carbon::setTestNow(Carbon::now());
+
+        CarbonSleeper::sleep(-1);
+    }
+
+    public function testRealSleep()
+    {
+        $before = time();
+
+        CarbonSleeper::sleep(1);
```
lucasmichot: Well in fact your code DOES delay execution, try with `CarbonSleeper::sleep(999999);`

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,88 @@
...
+        );
+    }
+
+    public function testRealSleep()
+    {
+        Carbon::setTestNow();
+
+        $before = time();
+
+        CarbonSleeper::sleep(1);
```
lucasmichot: The code execution is delayed, try `Carbon:sleep(999999);`

tests/CarbonSleeper/SleepTest.php
rrehbeindoi: Yes. The assertion message is for in case of failure, this is why we are expecting whatever we are asserting.

See: https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertEquals

> Reports an error identified by $message if the two variables $expected and $actual are not equal.

The message is not the success condition.  It is the failure condition.  The message is used when the test fails.

ie:
```php
    public function testRealSleep()
    {
        // Carbon::setTestNow(); - Wrapper left the test time set

        $before = time();

        CarbonSleeper::sleep(1);

        $after = time();

        $this->assertNotEquals($after, $before, 'Sleep did not delay execution');
    }
```

With message passed to the assertion
```sh
There was 1 failure:

1) Tests\CarbonSleeper\SleepTest::testRealSleep
Sleep did not delay execution
Failed asserting that 1447435126 is not equal to 1447435126.
```

Without message passed to the assertion
```sh
There was 1 failure:

1) Tests\CarbonSleeper\SleepTest::testRealSleep
Failed asserting that 1447435126 is not equal to 1447435126.
```

So, you are correct, the real sleep test does really call the real delay.  But that is not what the message is for.  It is for the case as documented in PHPUnit's documentation for when the message will be used.

lucasmichot: I completely understand the usage of the $message argument; but this is not the point.
if we use `CarbonSleeper:sleep(99999)` the code execution is locked.
I am not sure we would like that
Thoughts @briannesbitt ?

tests/CarbonSleeper/SleepTest.php
rrehbeindoi: In the conditions as described in my request #514, really calling sleep in non-test-time conditions is the point.

s4wny:
I think this is a great addition to Carbon. Carbon is all about time, and sleeping/delaying code execution is highly related to time.

This method would be really useful for code like this:
```php
//...
$IRCBot->say('Count down...')
for($i=10;$i>0;$i--) {
    $IRCBot->say($i . ' seconds...');
    sleep(1);
}
//...
```

Currently this code is untestable / really slow to test, but with @rrehbeindoi's method it wouldn't be. With `Carbon::sleep(1)` the program would sleep for one second in production, but not at all in a testing environment where `Carbon::setTestNow(Carbon::now())` is used.

src/Carbon/CarbonSleeper.php
```diff
@@ -0,0 +1,53 @@
...
+     * Delay execution
+     *
+     * When Carbon's testing time is set, it will fake sleeping
+     *
+     * Like PHP's sleep function, this will accept zero without error. Unlike
+     * PHP's sleep function, negative sleep will throw an exception.
+     *
+     * @param int $seconds Halt time in seconds
+     *
+     * @throws InvalidArgumentException
```
lucasmichot: Should be `@throws \InvalidArgumentException`

src/Carbon/CarbonSleeper.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /**
+     * Delay execution
+     *
+     * When Carbon's testing time is set, it will fake sleeping
+     *
+     * Like PHP's sleep function, this will accept zero without error. Unlike
+     * PHP's sleep function, negative sleep will throw an exception.
+     *
+     * @param int $seconds Halt time in seconds
```
lucasmichot: Usually no argument description is set, the function description itself should be sufficient.

tests/CarbonSleeper/SleepTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+use Carbon\Carbon;
+use Carbon\CarbonSleeper;
+use Tests\AbstractTestCase;
+
+class SleepTest extends AbstractTestCase
+{
+    public function testFakeSleepDelay()
+    {
+        $self = $this;
```
lucasmichot: Maybe use `$scope = $this;` instead

lucasmichot:
@rrehbeindoi , deleted fork ?

rrehbeindoi:
> deleted fork?

Yes. Based on comments such as:

> The code execution is delayed, try Carbon:sleep(999999);

And that this had no activity for a few months, it seemed to be a dead and unwanted feature request. Sorry, for some reason I thought I'd also closed the PR when I did house cleaning.

Issue #515, opened 2015-11-13
Feature request: PHPUnit listener to reset


rrehbeindoi:
``` php
class CarbonClear extends \PHPUnit_Framework_BaseTestListener
{
    public function endTest(\PHPUnit_Framework_Test $test, $time)
    {
        Carbon::setTestNow();
    }
}
```

phpunit.xml.dist

``` xml
<phpunit>
    <listeners>
        <listener class="CarbonClear" />
    </listeners>
</phpunit>
```

lucasmichot:
See https://github.com/briannesbitt/Carbon/pull/520

PR #520, opened 2015-11-13
Added resetting Carbon test time during tests

rrehbeindoi:
From my request: #515 

In theory, could be used by consumers of the carbon library, not exclusively for Carbon tests.

Question:
As a usable outside of carbon, but still for unit tests, should it be in tests/... or src/...?

### DIFF
--- a/tests/ClearCarbonListener.php
+++ b/tests/ClearCarbonListener.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Tests;
+
+use PHPUnit_Framework_BaseTestListener;
+use PHPUnit_Framework_Test;
+use Carbon\Carbon;
+
+/**
+ * Reduce side effects of unit tests with Carbon setTimeNow usage
+ *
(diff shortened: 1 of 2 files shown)


tests/ClearCarbonListener.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Tests;
+
+use PHPUnit_Framework_BaseTestListener;
+use PHPUnit_Framework_Test;
+use Carbon\Carbon;
+
+/**
+ * Reduce side effects of unit tests with Carbon setTimeNow usage
```
GrahamCampbell: missing fullstop

lucasmichot:
Why not simply adding `Carbon::setTestNow();` to `\Tests\AbstractTestCase::tearDown()` ?

briannesbitt:
@lucasmichot your idea seems like an easier change for this.

PR #526, opened 2015-11-19
Get spec_interval string from CarbonInterval

### LINKED ISSUE #525: Ability to get interval_spec from CarbonInterval
This is a feature request that I am happy to implement and send a Pull request with, but before that I'd like to discuss it.

I am currently in a need of getting `interval_spec` from `CarbonInterval` object. I imagine it could be achieved by implementing `spec` method on the `CarbonInterval` class. Do you see any potential issues with this?

martinssipenko:
Fixes #525

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,5 +36,10 @@
     "psr-4": {
       "Tests\\": "tests/"
     }
+  },
+  "extra": {
+      "branch-alias": {
+          "dev-master": "1.x-dev"
+      }
   }
 }
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -520,4 +520,39 @@ class CarbonInterval extends DateInterval
 
         return $this;
     }
+
+    /**
+     * Get the interval_spec string
+     *
+     * @return string
+     */
+    public function spec()
+    {
+        $date = array_filter(array(
+            static::PERIOD_YEARS  => $this->y,
+            static::PERIOD_MONTHS => $this->m,
+            static::PERIOD_DAYS   => $this->d,
+        ));
+
+        $time = array_filter(array(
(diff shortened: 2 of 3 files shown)


martinssipenko:
Looks like CI is failures are false positives.

dereuromark:
What about __toString returning this string?

    $this->assertEquals('PT1M', (string)$ci);

For me spec() doesn't say much about what it does or returns.

composer.json
```diff
@@ -36,5 +36,10 @@
     "psr-4": {
       "Tests\\": "tests/"
     }
+  },
+  "extra": {
+      "branch-alias": {
+          "dev-master": "1.x-dev"
```
GrahamCampbell: don't add this

GrahamCampbell: move to a different PR

src/Carbon/CarbonInterval.php
```diff
@@ -520,4 +520,39 @@ public function add(DateInterval $interval)
...
+        }
+
+        if (count($time) > 0) {
+            $specString .= static::PERIOD_TIME_PREFIX;
+            foreach ($time as $key => $value) {
+                $specString .= $value . $key;
+            }
+        }
+
+        return $specString === static::PERIOD_PREFIX ? 'PT0S' : $specString ;
```
dereuromark: looks like there is a whitespace too much at the very end (before ;).

martinssipenko:
Re: `__toString`, it would make it not backwards compatible.

martinssipenko:
Closing this PR and moving code change to https://github.com/briannesbitt/Carbon/pull/530, that removes the updates in `composer.json`. Have to create a new PR, because my forks `master` branch is used by other apps i'm working on, and I have to keep the `composer.json` change in there.

PR #530, opened 2015-11-28
Get spec_interval string from CarbonInterval

### LINKED ISSUE #525: Ability to get interval_spec from CarbonInterval

martinssipenko:
Fixes #525

New version of https://github.com/briannesbitt/Carbon/pull/526 that removes updates to `composer.json`

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -552,6 +552,6 @@ class CarbonInterval extends DateInterval
 	    }
 	}
 
-	return $specString === static::PERIOD_PREFIX ? 'PT0S' : $specString ;
+	return $specString === static::PERIOD_PREFIX ? 'PT0S' : $specString;
     }
 }
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -527,31 +527,31 @@ class CarbonInterval extends DateInterval
      */
     public function spec()
     {
-	$date = array_filter(array(
-	    static::PERIOD_YEARS  => $this->y,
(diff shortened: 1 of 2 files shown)


dereuromark:
I still think it would be better to use __toString() here as I did in https://github.com/cakephp/chronos/pull/59 respectively.

martinssipenko:
@GrahamCampbell Can you help me with failing StyleCI, I'm not sure I understand the issue.

Thanks!

src/Carbon/CarbonInterval.php
```diff
@@ -519,4 +519,39 @@ public function add(DateInterval $interval)
...
+        $specString = static::PERIOD_PREFIX;
+
+        foreach ($date as $key => $value) {
+            $specString .= $value . $key;
+        }
+
+        if (count($time) > 0) {
+            $specString .= static::PERIOD_TIME_PREFIX;
+            foreach ($time as $key => $value) {
+            $specString .= $value . $key;
```
dereuromark: this line is shown as invalid: https://styleci.io/analyses/q5v6Pz

martinssipenko:
Sorry, my bad, was looking at wrong line.

nicholas-eden:
This results in a lot of duplication of code in the constructor.  Maybe have a buildSpec method that accepts $years, $months, etc.  Then have both the constructor and spec method call it.

PR #532, opened 2015-11-29
Added armenian language translation

mhamlet:


lucasmichot:
Thanks @mhamlet , can you add tests for that ?
You can use `\Tests\Localization\FrTest` as a starting point

lucasmichot:
Thanks @mhamlet !

PR #537, opened 2015-12-02
Fix Dutch translation for years

adriaanzon:
Same reason as c9e9c256211f0a247daaec9b9eb19a6f7d398790

### DIFF
--- a/src/Carbon/Lang/nl.php
+++ b/src/Carbon/Lang/nl.php
@@ -12,7 +12,7 @@
  * for possible formats.
  */
 return array(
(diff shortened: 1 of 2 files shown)


lucasmichot:
Thanks @adriaanzon  , can you also add tests for that ?
You can use `\Tests\Localization\FrTest` as a starting point

lucasmichot:
Thanks for https://github.com/adriaanzon/Carbon/commit/0261b135d70411233b1237f73c825bac71df8800 @adriaanzon !

Issue #427, opened 2015-08-22
issues with HHVM


MaxRink:
https://travis-ci.org/briannesbitt/Carbon
https://github.com/briannesbitt/Carbon/blob/master/tests/DiffTest.php

nicholas-eden:
There is a bug with the magic setter and getter on the DateInterval which seems to be fixed in a later version:
https://github.com/facebook/hhvm/issues/2952

Having the Carbon classes call parent::__set and __get seems to solve the problem.

The DatePeriod had a bug as well:
https://github.com/facebook/hhvm/issues/3947

Still trying to work this out.

inctor:
+1 for this, just encountered it, and almost went mental trying to figure out what was going on.

Any work-arounds for this, that does not include altering vendor-files?

nicholas-eden:
I have a pull request that puts in a work around for all but one hhvm bug. If your hhvm version is 3.8 or higher then they ask are fixed.

briannesbitt:
See comments in #541

PR #541, opened 2015-12-03
Changes for HHVM compatibility

nicholas-eden:
https://github.com/briannesbitt/Carbon/issues/427

It looks like later versions will fix some of the bugs, not sure which versions this will pass for besides 3.10.1

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -13,8 +13,10 @@ namespace Carbon;
 
 use Closure;
 use DateTime;
+use DateTimeImmutable;
 use DateTimeZone;
 use DatePeriod;
+use Exception;
 use InvalidArgumentException;
 use Symfony\Component\Translation\Translator;
 use Symfony\Component\Translation\TranslatorInterface;
@@ -196,6 +198,18 @@ class Carbon extends DateTime
         }
 
         return $tz;
+    }
+
+    /**
+     * Determines if code is being executed by HHVM
+     *
+     * @return bool
+     *
+     * @codeCoverageIgnore
+     */
+    public static function isHhvm()
+    {
+        return defined('HHVM_VERSION');
     }
 
     ///////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


nicholas-eden:
We are down to 1 failing assert for hhvm and it isn't one that can be fixed with a workaround.

There are a few things I'm not too proud of, a few hacks in isHhvm blocks.

src/Carbon/Carbon.php
```diff
@@ -198,6 +200,18 @@ protected static function safeCreateDateTimeZone($object)
         return $tz;
     }
 
+    /**
+     * Determines if code is being executed by HHVM
+     *
+     * @return bool
+     *
+     * @codeCoverageIgnore
```
lucasmichot: We avoid that `@codeCoverageIgnore` tag.

src/Carbon/Carbon.php
```diff
@@ -772,11 +786,23 @@ public function tz($value)
...
      */
     public function setTimezone($value)
     {
-        return parent::setTimezone(static::safeCreateDateTimeZone($value));
+        try {
+            return parent::setTimezone(static::safeCreateDateTimeZone($value));
+        } catch (Exception $e) {
+            // @codeCoverageIgnoreStart
+            $message = $e->getMessage();
+            if (self::isHhvm() && 0 === strpos($message, 'DateTimeZone::__construct()')) {
```
lucasmichot: We don't use yoda conditions, `strpos($message, 'DateTimeZone::__construct()') === 0` is to be chosen

lucasmichot:
Just an idea, would it not make sense to make an extended class which only target HHVM...  ~~ warum nicht
@briannesbitt @dereuromark, thoughts ?

nicholas-eden:
Ideally there would be no code changes required when switching engines, so if it were a new class specific to hhvm some form of composition might work.

src/Carbon/Carbon.php
GrahamCampbell: also, should be static:: rather than self::

nicholas-eden:
That last test will be fixed a slightly later version of hhvm: https://github.com/facebook/hhvm/pull/5240

I'm guessing 3.7, not sure how often Travis CI updates hhvm or if @briannesbitt needs to do it.

nicholas-eden:
Nevermind, the answer is here: http://stackoverflow.com/questions/32151774/does-travis-ci-support-hack

Travis is currently locked to 3.6.  We could put in a skip for that assertion based on the hhvm version.  Everything else would pass, allowing hhvm to be removed from the ignore list.

lucasmichot:
Can you rebase @nicholas-eden ?

lucasmichot:
Can you rebase @nicholas-eden ?

kylekatarnls:
Hi,

IMO, we simply should drop HHVM support. HHVM brought so much pain to so many people. It was a good idea and the performance worsted this pain before, but now with the PHP 7 epoch it's tile to let this old good friend rest in peace mostly if it needs many specific code.

I'm sorry for your hard work, it would indeed be relevant in 2015 when  you did it, but I think I would close this PR if anyone else has arguments to oppose.

briannesbitt:
Based on https://hhvm.com/blog/2017/09/18/the-future-of-hhvm.html it looks like they will no longer officially be supporting all of the PHP language features.  I think its best this doesn't get merged and we don't specifically support HHVM moving forward.  This could be done in a subclass if need be in the future, but probably in a different repo.

Issue #539, opened 2015-12-03
endOfMonth() not working correctly with 0,0,0 setTime


m4r3x:
I have following code:

```
    $startMonth = (new Carbon('first day of this month'))->setTime(0,0,0);
    $endMonth = clone $startMonth;
    $endMonth2 = clone $startMonth;
    var_dump($endMonth);
    var_dump($endMonth->endOfMonth());
    var_dump($endMonth2->addMinutes(1)->endOfMonth());
    die();
```

And I get following result: https://gist.github.com/m4r3x/146355c982f48d97be54

I don't have a time to look it further or write some details, I'll stick with addMinutes() fix for now. Looking forward knowing why it's happening.

Cheers, have a nice day,
Mark.

nicholas-eden:
I hate to point to php, but you may have found a bug.

```
$dt = new \DateTime('first day of this month');
        echo $dt->format(\DateTime::ISO8601) . PHP_EOL;
        $dt->setTime(0,0,0);
        echo $dt->format(\DateTime::ISO8601) . PHP_EOL;
        $dt->setDate(2015,12,31);
        echo $dt->format(\DateTime::ISO8601) . PHP_EOL;
        $dt->setTime(23,59,59);
        echo $dt->format(\DateTime::ISO8601) . PHP_EOL;
```

```
2015-12-01T11:27:18-0500
2015-12-01T00:00:00-0500
2015-12-01T00:00:00-0500
2015-12-01T23:59:59-0500
```

It is not allowing the date to be changed if the object was declared with a natural language string.  It doesn't seem related to the setTime.  I think you're going to have to code around it.

I found a reported issue which seems to be what we have here: https://bugs.php.net/bug.php?id=63863

nicholas-eden:
The comment on that bug had an obscure workaround.  Pull request submitted to put it in.

lucasmichot:
Closes by https://github.com/briannesbitt/Carbon/issues/542

PR #542, opened 2015-12-04
Workaround for PHP bug

nicholas-eden:
https://github.com/briannesbitt/Carbon/issues/539

Calling modify before setDate releases whatever is locking the day.


lucasmichot:
Nice one @nicholas-eden :+1: 
`$this->modify('+0')` also works by the way

PR #543, opened 2015-12-04
Use assertInstanceOfCarbon

lucasmichot:
We have `\Tests\AbstractTestCase::assertInstanceOfCarbon` so let's use it.


GrahamCampbell:
:+1:

PR #544, opened 2015-12-04
Use only one assertion

lucasmichot:
Useless to run multiple assertions where we can make a single assertion on an array of actual and expected properties


danielstjules:
It looks like this PR makes the assertions harder to debug on failure, since you're dropping the `$message` from `assertSame`

GrahamCampbell:
I disagree, This makes it easier to debug because all the bits are actually assented rather than exiting after the first failure. It's defintely easier to debug. The array keys make it clear what's going on.

:+1:

Issue #435, opened 2015-09-11
Carbon very slow object instantiation


tschallacka:
I'm not sure if something can be done about this, but if possible, it would be great :-)

I have a simply loop of 1095 times.
This loop takes 30 ms to complete solely thanks to carbon instantiation.

```
$weekModifier = 0;
$rules = [];
$startt = microtime(true);
$date = new DateTime();
/** starting on monday **/
if($date->format('N') !== 1) {
    $date->sub(new DateInterval('P'. $date->format('N') . 'D'));
}
$interval = new DateInterval('P'.abs($weekModifier).'W');
if($weekModifier > 0) {
    $date->add($interval);  
} else {
    $date->sub($interval);  
}
$carbon = Carbon::instance($date);
/** Generate all rules for 3 years */
for($i = 1; $i <= 365*3; $i++) {
$curdate = $carbon->addDay();//->format('l Y-m-d') . "<br>\n";
$rules[] = $carbon->copy();
}
$endt = microtime(true);
$creationtime = ($endt - $startt)*1000;
echo "Page created in $creationtime milliseconds<BR/>"; 
```

31.199932098389 ms to complete.

if you take out the carbon copy it completes in 0ms

So it's definately carbon.

So my question is, can object instantiation be made cheaper and more efficient? 
Because 30ms is a lot on 1095 objects of supportive nature.

tschallacka:
I've been doing some experimentation, and if I use

    $copy = clone $carbon;

 I get some speed improvements. Mabe you can use that in the $carbon->copy() method where you copy the internal DateTime object into a new carbon instance, instead of going through the new error checking, parsing, datetimesetting hoops carbon jumps through.

Sverri:
It takes ~200ms for me.

Windows 8.1, PHP 5.6.8, i7 4970k @ ~4.4GHz, 16GB 1600 MHz.

![cachegrind out](https://cloud.githubusercontent.com/assets/769089/9835155/fdb53a50-59db-11e5-92fa-791822e22e50.png)

nicholas-eden:
I did several tests on this as well, going through the constructor seems to be rather slow.

Using copy: https://blackfire.io/profiles/65f87955-7040-4a7c-a541-355074e2cd97/graph
Using clone: https://blackfire.io/profiles/620ea414-30e6-4f23-9e15-a30ca1d346c0/graph

Some other things to potentially optimize:
[...]

PR #546, opened 2015-12-06
Optimize Copy

nicholas-eden:
Based on this thread: https://github.com/briannesbitt/Carbon/issues/435

I don't see a reason to go through the full instantiation for a copy, clone should work fine.

copy-optimize

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -472,7 +472,7 @@ class Carbon extends DateTime
      */
     public function copy()
     {
-        return static::instance($this);
+        return clone $this;
     }
 
     ///////////////////////////////////////////////////////////////////


lucasmichot:
:+1:

briannesbitt:
I'll merge this if you can revert the unrelated hasTestNow change.

lucasmichot:
> I'll merge this if you can revert the unrelated hasTestNow change.

Ping @nicholas-eden

nicholas-eden:
@briannesbitt done

hirayama-onagigawa-computing:
@briannesbitt 
This is essentially right I think, but changes behavior of subclasses.

- Constructor is now _not_ called.
- Properties are now copied (unexpectedly).

```php
<?php

require_once 'src/Carbon/Carbon.php';

use Carbon\Carbon;

class CarbonChild extends Carbon
{
    public $prop = 'foo';

    public function __construct($time = null, $tz = null)
    {
        echo 'constructor called.', "\n";

        parent::__construct($time, $tz);
    }
}

$obj1 = new CarbonChild;
$obj1->prop = 'bar';

$obj2 = $obj1->copy();
echo $obj2->prop, "\n";
```
1.21 or lower:
```
> php test.php
constructor called.
constructor called.
foo
```
1.22 or higher:
```
> php test.php
constructor called.
bar
```

PR #547, opened 2015-12-07
Change variable name

lucasmichot:
Change variable name


GrahamCampbell:
Seem a bit pointless, no?

lucasmichot:
I don't think so
The variable name is misleading

PR #90, opened 2014-01-31
Issue 88 : modify with timezone

kylekatarnls:
[Issue 88 : Adding to a time provides an inconsistent result](https://github.com/briannesbitt/Carbon/issues/88)

PR #548, opened 2015-12-07
Fix issue 88

### LINKED ISSUE #88: Adding to a time provides an inconsistent result
Firstly I love the library (especially as part of Laravel) and this is a PHP issue more than a Carbon issue.

When adding/subtracting times for a value against a Timezone other than UTC, the add/subtract features do not take daylight saving into consideration

```php
<?php
// Clocks go forward on this day: 00:59:59 > 02:00:00
$a = new Carbon('2014-03-30 00:00:00', 'Europe/London'); 
$b = $a->copy();
$b->addHours(24);
echo $b; // Produces 2014-03-31 00:00:00, expected 2014-03-31 01:00:00
echo $a->diffInHours($b); // Produces 23, expected 24.
```

It seems that PHP's modify() function ignores daylight saving values entirely. The documentation often mentions that comparisons are always done in UTC, so I would expect the additions to add UTC units of time as well.

This also leads the possibility of time loops!
http://uk3.php.net/manual/en/datetime.modify.php#113549

lucasmichot:
Replace #90 and fix #88

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2642,7 +2642,7 @@ class Carbon extends DateTime
     /**
      * Consider the timezone when modifying the instance.
      *
-     * @param  string $modify
+     * @param string $modify
      *
      * @return static
      */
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:+1:

th3fallen:
Any ETA on this getting merged? this will resolve an issue were experiencing.

lucasmichot:
ping @briannesbitt

PR #553, opened 2015-12-12
Unalign array arrows

lucasmichot:
Arrows are not aligned in the code - there is no reason they should be aligned in lang files.
Plus it's adding useless noise when adding or removing an element in the array

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -35,6 +35,7 @@ $fixers = array(
     'phpdoc_var_without_name',
     'return',
     'spaces_cast',
+    'unalign_double_arrow',
     'unalign_equals',
     'unused_use',
     'whitespacy_lines',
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:-1:

dereuromark:
:+1: (!)

GrahamCampbell:
> Plus it's adding useless noise when adding or removing an element in the array

And what is this change? The maximum amount of noise possible, destroying `git blame`.

dereuromark:
Git blame by its very nature cannot work with one of the worst coding practices there are so far: Inline alignment. So at least from now on it can :)

lucasmichot:
A good article on that: http://blog.madewithlove.be/post/code-style-options-for-cleaner-diffs/

dereuromark:
Indeed, the advantages in readability do not outweigh the disadvantages here.

PR #561, opened 2015-12-29
Add getter to test for "long years"

ssfinney:
I've had trouble with my date math in a project because 2015 has 53 weeks in it, not 52. I would like to add a method to check and see if this year is a "long year", [according to the ISO standard](https://en.wikipedia.org/wiki/ISO_week_date#Weeks_per_year).

According to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Week_dates), "28 December is always in the last week of its year." So I just made a new Carbon instance on the 28 of December for the year, and check if the week number is 53.

I fully tested the method according to your standards. It's an easy method to add, and one that would be very helpful in our project. Please merge :)

(P.S. I got this idea from [Glavić's post on StackOverflow](http://stackoverflow.com/a/21480444/2916818)).

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1576,6 +1576,16 @@ class Carbon extends DateTime
     }
 
     /**
+     * Determines if the instance is a long year
+     *
+     * @return bool
+     */
+    public function isLongYear()
+    {
+        return static::create($this->year, 12, 28, 0, 0, 0, $this->tz)->weekOfYear === 53;
+    }
+
+    /**
      * Checks if the passed in date is the same day as the instance current day.
      *
      * @param \Carbon\Carbon $dt
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1578,6 +1578,8 @@ class Carbon extends DateTime
     /**
      * Determines if the instance is a long year
      *
+     * @see https://en.wikipedia.org/wiki/ISO_8601#Week_dates
+     *
      * @return bool
      */
     public function isLongYear()
--- a/tests/Carbon/GettersTest.php
+++ b/tests/Carbon/GettersTest.php
@@ -236,6 +236,16 @@ class GettersTest extends AbstractTestCase
         $this->assertFalse(Carbon::createFromDate(2011, 1, 1)->isLeapYear());
     }
 
+    public function testIsLongYearTrue()
+    {
+        $this->assertTrue(Carbon::createFromDate(2015, 1, 1)->isLongYear());
+    }
+
+    public function testIsLongYearFalse()
+    {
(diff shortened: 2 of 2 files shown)


ssfinney:
This would also bring the number of tests up to 800, which is a nice round number :)

src/Carbon/Carbon.php
```diff
@@ -1576,6 +1576,16 @@ public function isLeapYear()
     }
 
     /**
+     * Determines if the instance is a long year
+     *
+     * @return bool
+     */
+    public function isLongYear()
+    {
+        return static::create($this->year, 12, 28, 0, 0, 0, $this->tz)->weekOfYear === 53;
```
lucasmichot: Why the `28` and not the last day of the year?

ssfinney: Occasionally, December 31st falls on week 1 of the following year.

For example, when I make a Carbon for 12-31-2018, it's considered week 1. So we wouldn't be able to tell if there were 53 weeks in 2018 by using the 31st.

```
[9] > $date = Carbon\Carbon::create(2018, 12, 31, 0, 0, 0, 'America/New_York');
// object(Carbon\Carbon)(
//   'date' => '2018-12-31 00:00:00.000000',
//   'timezone_type' => 3,
//   'timezone' => 'America/New_York'
// )
[10] > $date->format('W');
// '01'
```

The 28th always falls on week 52 or if it's a long year, week 53. Reference: https://en.wikipedia.org/wiki/ISO_8601#Week_dates

lucasmichot: It would be worth mentioning this source then:
`* @see https://en.wikipedia.org/wiki/ISO_8601#Week_dates`

ssfinney: Ok, I'll make that change.

Here's a source that defines a "long year", would you like it in there also? https://en.wikipedia.org/wiki/ISO_week_date#Weeks_per_year

lucasmichot: Nah, I don't think that's necessary - I've read it too
https://en.wikipedia.org/wiki/ISO_8601#Week_dates is already self-explanatory IMO

ssfinney: Ok, I've made that change. Thank you for looking at this!

lucasmichot:
:+1:

ssfinney:
The build is failing on a test that *seems* unrelated to this change (and only on the PHP 5.6 box). Could someone please re-run the build?

lucasmichot:
@ssfinney you can squash your commits and push force, the tests will run again

ssfinney:
@lucasmichot Thanks, everything is passing now.

briannesbitt:
There are 812 tests now so it will be 814 but I'll merge this anyway ;)

PR #565, opened 2016-01-07
Remove system ignore items and suffix vendor.

lucasmichot:
Some ignored files are system-wide ignored files.
Note : `vendor/` can be suffixed. The behavior is the same, but IDE understand it better.


briannesbitt:
Maybe not everyone will have a "core.excludesfile" setup?

PR #570, opened 2016-01-08
Cover unknown method call in carbon interval

EGYG33K:


EGYG33K:
Please ignore it is by mistake

PR #575, opened 2016-01-14
Fix typo

BraxtonFair:
Not exactly needed, but just thought that this would be a small contribution


lucasmichot:
@BraxtonFair , it seems Travis failed for as strange reason. Can you restart the tests (close then re-open this PR)

PR #581, opened 2016-01-15
Use interface instead of concrete implementation

theofidry:
Using `\DateTimeInterface` is far more flexible and allow one to instantiate a `Carbon` instance via a `\DateTimeImmutable` instead of a `\DateTime` instance for example.


dereuromark:
Refs https://github.com/briannesbitt/Carbon/pull/233

theofidry:
Ha my bad, only checked the issues -_-. Thanks @dereuromark

PR #584, opened 2016-01-15
Hungarian translation improved A LOT

balping:
... after I realised that actually I can use second_before, second_ago, etc keys.
Reading the original source code can be really useful!

Also, unit test added for Hungarian language

### DIFF
--- a/src/Carbon/Lang/hu.php
+++ b/src/Carbon/Lang/hu.php
@@ -17,15 +17,40 @@
  * Extracted from https://github.com/jenssegers/laravel-date/blob/master/src/lang/hu/date.php
  */
 return array(
-    'year'      => '1 évvel|:count évvel',
-    'month'     => '1 hónappal|:count hónappal',
-    'week'      => '1 héttel|:count héttel',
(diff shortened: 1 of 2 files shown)


balping:
StyleCI formatting corrected

GrahamCampbell:
> StyleCI formatting corrected

Not exactly.

dereuromark:
Just some minor whitespace: https://styleci.io/analyses/qoaeZ8
@balping Always keep your "trim right" option of your IDE enabled to prevent this.

balping:
Nah, finally StyleCI has passed

@dereuromark Thanks for the suggestion

lucasmichot:
Great work @balping !
Yeah lots of changes, this is what I was suspecting... I am a bit skeptical about the qualities of other translations files BTW, I've been checking fr/he/de files which are OK, but your PR is a good indicator of issues of untested and mostly imported language files.

balping:
Thanks @lucasmichot  !

Yeah, I was working on my own projects when I realized that in many cases the printed text made actually no sense. Maybe that's the case with other translations of smaller languages as well

PR #586, opened 2016-01-21
Handles negative and +4 digit-long years.

lucasmichot:
This PR makes the following possible:
- create a date with a negative year
- create a date with 5 digit-long dates (-10000 / +10000)

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -387,7 +387,9 @@ class Carbon extends DateTime
             $second = $second === null ? 0 : $second;
         }
 
-        return static::createFromFormat('Y-n-j G:i:s', sprintf('%s-%s-%s %s:%02s:%02s', $year, $month, $day, $hour, $minute, $second), $tz);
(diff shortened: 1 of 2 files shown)


briannesbitt:
I think this in combination with the modify method breaks.  There was no DST until the early 1900's so this `var_dump(Carbon::createFromTime(12, 0, 0, new DateTimeZone('Europe/London')))` produces a datetime on `2016-06-25 13:00:00.000000`.

I think its because the new implementation initially creates the date in year 0 (no DST), then adds 2016 years, but does it in UTC time, because of the new modify().  So it adds 2016 years in UTC timezone, then converts it back into London time which in 2016 observes DST.  This then pushes it ahead by an hour which is not really what the user would expect.

I have a hack fix in the create to handle this.  Basically the last 1/2 of create turns into :

```php
        $fixYear = null;

        if ($year < 0) {
          $fixYear = $year;
          $year = 0;
        } else if ($year > 9999) {
          $fixYear = $year - 9999;
          $year = 9999;
        }

        $instance = static::createFromFormat('Y-n-j G:i:s', sprintf('%s-%s-%s %s:%02s:%02s', $year, $month, $day, $hour, $minute, $second), $tz);
        
        if ($fixYear !== null) {
          $instance->addYears($fixYear);
        }

        return $instance;
```

PR #587, opened 2016-01-25
Create solarCarbon

persiscms:
adds solar calendar (shamsi) for persian time to carbon

### DIFF
--- a/solarCarbon
+++ b/solarCarbon
@@ -1,0 +1,355 @@
+<?php namespace Carbon;
+
+use DateTimeZone;
+
+/**
+ * Class jDate
+ * @package jDate
+ * @property string solarDay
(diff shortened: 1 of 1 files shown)


lucasmichot:
@persiscms thanks for your PR.
I feel like this feature would also apply to hebrew or arabic calendars.
Nevertheless, I am not convinced this is is the good approach.

Also I feel like we would need some tests.

---

Different topic, if you are a native farsi speaker, we would love to get tests for farsi translation ;-)

persiscms:
@lucasmichot thanks for reply.
absence of such great library like carbon for persian calendar is felt.
it would be grateful if you use my codes.
and also id love to test for your farsi translations

Issue #583, opened 2016-01-15
Check for invalid date


73sl4:
Can we have a isValid(), currently there is no way to check if a date is valid or not.
Also, if I use Carbon as:
```php
$dt = Carbon::createFromDate(2016, 01, 32);
echo $dt->toDateString(); // 2016-02-01
```
Returned instance is for incorrect date, it should either return a null object or throw InvalidDateException

lucasmichot:
@73sl4 this is the normal behaviour of `\DateTime` which is extended.
Nevertheless, you can maybe try something like this...

```php
public static function createSafe($year = null, $month = null, $day = null, $hour = null, $minute = null, $second = null, $tz = null)
{
    $year = $year === null ? date('Y') : $year;
    $month = $month === null ? date('n') : $month;
    $day = $day === null ? date('j') : $day;

    if ($hour === null) {
        $hour = date('G');
        $minute = $minute === null ? date('i') : $minute;
        $second = $second === null ? date('s') : $second;
    } else {
        $minute = $minute === null ? 0 : $minute;
        $second = $second === null ? 0 : $second;
    }

    $attributes = array_filter(compact($year, $month, $day, $hour, $minute, $second));

    $unsafeInstance = static::create($year, $month, $day, $hour, $minute, $second, $tz);

    foreach ($attributes as $key => $value) {
        if ($unsafeInstance->$key != $value) {
            throw new \InvalidArgumentException(sprintf("Attribute %s : %s is not a valid value", $key, $value));
        }
    }

    return $unsafeInstance;
}
```

But this is quite hacky

73sl4:
@lucasmichot I think this will work. Are you going to add it to Carbon repo?

lucasmichot:
We're open to PRs submission

RRStoyanov:
There is a PHP function [checkdate](http://php.net/manual/en/function.checkdate.php) which validates  give Gregorian date. So I think it can be added to Carbon as isValidGregorian() method. I might think of adding is to my current project and maybe create a PR?

lucasmichot:
Indeed, this function is useful. Nevertheless, I think that it would be more useful to provide the key (year, month, day, hour, minute or second) and the "unsafe" value 

If not PR comes the next days, I’ll take care of it.

On 26 January 2016 at 20:19:12, Rostislav Stoyanov (notifications@github.com) wrote:

[...]

PR #588, opened 2016-01-26
Add createSafe

lucasmichot:
This is a PR concerning #583

Note:

* `createSafe`, I could not find a better name.
* `checkDate` is not reliable at the moment as we can only enter a year in of 0-9999 range, `checkDate` allows 0-32767
* We wish to indicate which argument throws an exception and with which value
* Should the `createSafe` only allows hour, minute and second arguments to be ignored (set to null) ? 

Suggestions are welcomed.

----

* [x] Add tests
* [x] Add `createSafe`

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -454,7 +454,7 @@ class Carbon extends DateTime
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


shooka:
The method should be named `safeCreate` to match the existing method `safeCreateDateTimeZone`.

PR #590, opened 2016-01-28
Fix mistype

riccai:
Just small mistype fix

### DIFF
--- a/src/Carbon/Lang/lt.php
+++ b/src/Carbon/Lang/lt.php
@@ -16,7 +16,7 @@
 
 return array(
     'year'      => '1 metus|:count metus|:count metų',
-    'month'     => '1 menesį|:count menesius|:count menesių',
+    'month'     => '1 menesį|:count mėnesius|:count mėnesių',
     'week'      => '1 savaitę|:count savaites|:count savaičių',
     'day'       => '1 dieną|:count dienas|:count dienų',
     'hour'      => '1 valandą|:count valandas|:count valandų',


lucasmichot:
Thanks @riccai !
Are you a native Lithuanian speaker ? It would be awesome if you could also provide a test for this language

You can copy and paste `\Tests\Localization\FrTest` as a starting point

riccai:
@lucasmichot yes I am, I will take a look on that. Today I found something weird, maybe a bug.
For example I have time on database 2016-01-27 21:18:49, current server time is 2016-01-28 18:32:00 and when I use $classified->updated_at->diffForHumans() it returns "prieš 1 valandą", but it should be "prieš 21 valandą". When I switch language to English everything works fine.

lucasmichot:
Thanks you !

lucasmichot:
@riccai what is the difference between mėnesius and mėnesių for the plural of minutes?
Is mėnesius only used under a specific count ?

keksas:
@lucasmichot it's difficult to explain the difference.

Both words are only used for plural (and they mean Months not minutes as you stated).

`mėnesių` is plural genitive and `mėnesius` is plural accusative.

Sentence tense defines how they are used:

"Po 9 mėnesių" -> After nine months
"Prieš 2 mėnesius" -> Two months ago

lucasmichot:
OK I see, maybe check for https://github.com/balping/Carbon/blob/master/src/Carbon/Lang/hu.php in #584 from @balping
He's covering lots of different use cases in Magyar

lucasmichot:
Yeah, I've just checked, Magyar do have cases too (up to 18... !!!!)

lucasmichot:
See @balping very complete test too : https://github.com/balping/Carbon/blob/master/tests/Localization/HuTest.php

PR #592, opened 2016-01-28
Use consistant variable name

lucasmichot:
Use consistant variable name with camelcase


briannesbitt:
Why not just get rid of the `$timezone_offset` var and pass in `$object * 3600` to the function.

lucasmichot:
> Why not just get rid of the $timezone_offset var and pass in $object * 3600 to the function.

True. Done @briannesbitt

PR #593, opened 2016-01-28
Allows to create an instance from a \DateTimeImmutable

lucasmichot:
Allows to create an instance from a \DateTimeImmutable

PR #594, opened 2016-01-28
[2.0] Allows instantiation from DateTimeImmutable and DateTimeInterface

lucasmichot:
Allows instantiation from DateTimeImmutable and DateTimeInterface

---

**This PR is BC. It should be safe to merge it for version `2` only**
/cc @briannesbitt 

---

replace #233 and #593

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -13,6 +13,7 @@ namespace Carbon;
 
 use Closure;
 use DateTime;
+use DateTimeInterface;
 use DateTimeImmutable;
 use DateTimeZone;
 use DatePeriod;
@@ -241,7 +242,7 @@ class Carbon extends DateTime
     /**
      * Create a Carbon instance from a DateTime one.
      *
-     * @param \DateTime|\DateTimeImmutable $dt
+     * @param \DateTime|\DateTimeImmutable|\DateTimeInterface $dt
      *
      * @return static
      */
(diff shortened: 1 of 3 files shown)


src/Carbon/Carbon.php
```diff
@@ -240,17 +242,19 @@ public function __construct($time = null, $tz = null)
     /**
      * Create a Carbon instance from a DateTime one.
      *
-     * @param \DateTime $dt
+     * @param \DateTime|\DateTimeImmutable|\DateTimeInterface $dt
      *
      * @return static
      */
-    public static function instance(DateTime $dt)
+    public static function instance($dt)
```
dereuromark: By changing the public API aren't you introducing a BC break for all those that extend this class and this method?

lucasmichot: Indeeed https://github.com/lucasmichot/Carbon/commit/910028cdf4879dea0a718f8c6157ee370f8e1e9f is failing...
I think it will be safe to wait 2.0

lucasmichot: And I think (and hope) we will upgrade to PHP>=5.5

artursvonda: And shouldn't we just use DateTimeInterface? And stub it for older versions if needed?

artursvonda: And I agree with @lucasmichot that we should support only 5.5+ since anything older is not supported anymore by PHP at all.

dereuromark:
> This PR is BC

Yo mean "BC breaking" as BC means "backwards compatible".

For anyone who needs/wants to use the new 5.5 features, [Chronos](https://github.com/cakephp/chronos) might be interesting.

src/Carbon/Carbon.php
lucasmichot: @artursvonda this is in the plans

nicholas-eden: Shouldn't this be typehinting DateTimeInterface?

lucasmichot: I wouldn't : the current version is a bit locked because the argument is typehinted.

PR #595, opened 2016-01-28
Minor CS adjustments

lucasmichot:
Minor CS adjustments

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -16,6 +16,7 @@ Symfony\CS\Fixer\Contrib\HeaderCommentFixer::setHeader($header);
 
 $fixers = array(
     '-psr0',
+    'concat_without_spaces',
     'extra_empty_lines',
     'header_comment',
     'long_array_syntax',
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,8 +1,10 @@
 preset: psr2
 
 enabled:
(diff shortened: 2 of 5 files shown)


GrahamCampbell:
Why not enable these fixers?

lucasmichot:
Are they not already ?

lucasmichot:
Nvmm...

lucasmichot:
Added `concat_without_spaces` 
No clue of the fixer to add to ensure there is a linebreak between functions, any idea @GrahamCampbell ?

GrahamCampbell:
There is a method separation fixer. Not sure if it exists in php-cs-fixer 1.x.

dereuromark:
I vote against concat_without_spaces. Should be concat_with_spaces IMO.

GrahamCampbell:
concat without is more standard and is used by laravel and symfony, to name a few

dereuromark:
Not CakePHP for example. And concat is like any other operation, `+`, ... and all those have a spacing.
It is only consistent to use one. Only because some frameworks do not follow the consistency does not make it right, though :)

lucasmichot:
@dereuromark `concat_without_spaces` would make more sense as all of the other concatenation in the code are already spaceless.
@GrahamCampbell :+1: thx

PR #596, opened 2016-01-28
Update Lithuanian test

riccai:
Updated unit test for Lithuanian language.


lucasmichot:
Would probably be worth to add your commits from https://github.com/briannesbitt/Carbon/pull/590 in this  PR?

Issue #599, opened 2016-02-01
Serializing a Carbon\Carbon date doesn't work in HHVM


JoelESvensson:
It's easily reproducible in 3.11.1 (lastest HHVM as of now) with this one-liner:

`serialize(Carbon\Carbon::now());`

GrahamCampbell:
Surely you should report this to the HHVM team? It's not the job of PHP libraries to change to work on HHVM, it's the job of HHVM to be properly PHP compliant.

JoelESvensson:
Don't worry, i've also reported it on their issue tracker as well.

farizluqman:
Any updates?

kylekatarnls:
I agree with @GrahamCampbell and I would not spent efforts for HHVM (already spent too much in too many libraries), now with PHP 7, HHVM is no longer enough compatible and no longer provide performance advantage.

If you know how to fix and want to propose a PR for this, it is welcome, we will merge it if it does not break anything.

But our very small core team (compared to HHVM one) will focus on features and bugs with official PHP releases.

Thanks,

PR #600, opened 2016-02-02
Add serialize function

lucasmichot:
Add `serialize` and `fromSerialized` functions

---

Related to #599


briannesbitt:
Is this that helpful without an fromSerialized() static ?  Is this to support saving in sessions ?

GrahamCampbell:
Why is this needed. If HHVM os broken, HHVM should fix it?

Issue #606, opened 2016-02-05
[Idea] Concept of custom formatters / alias.


garygreen:
It would be good if you could specify custom-named formats straight within Carbon. This would be immensely useful for normalizing various systems that have many different 'named' date formats, without having to write a wrapper on top of Carbon every time.

```
Carbon::addAlias('cms_date_time', 'd/m/Y H:i');
Carbon::createFromAlias('cms_date_time', '12/05/2015 05:36');
Carbon::now()->toAliasFormat('cms_date_time');
```

lucasmichot:
See #608

briannesbitt:
For formatting specifically we have enough ways to store and read strings that can just as easily be passed to the format() function.  For real functionality I don't see this as a better solution to a subclass.

PR #608, opened 2016-02-09
Macro feature for Carbon.

lucasmichot:
Enable macros for Carbon.
Inspired by Laravel macro system.

---

Possible solution for #606 from @garygreen

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -11,6 +11,7 @@
 
 namespace Carbon;
 
+use BadMethodCallException;
 use Closure;
 use DateTime;
 use DateTimeZone;
@@ -160,6 +161,13 @@ class Carbon extends DateTime
      * @var \Symfony\Component\Translation\TranslatorInterface
      */
     protected static $translator;
+
+    /**
+     * The registered macros.
+     *
+     * @var array
+     */
+    protected static $macros = array();
 
     /**
      * Creates a DateTimeZone from a string, DateTimeZone or integer offset.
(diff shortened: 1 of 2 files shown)


briannesbitt:
Kinda cool.  I am not a huge fan of black magic though...  same as a subclass.

GrahamCampbell:
I don't think this is needed. Should be closed IMO.

GrahamCampbell:
Laravel's macro system has a very specific use case.

BrandonSurowiec:
Having macros would be really useful for adding helper functions to Carbon. Like one that can convert a Carbon instance to use the user's local time for instance. It's kind of hard to extend Carbon with a subclass in Laravel when the Model auto resolves dates to a Carbon instance. I'm a big fan of this PR.

BrandonSurowiec:
@lucasmichot Any chance of this being revisited?

GrahamCampbell:
Still a big :-1: from me.

lucasmichot:
See https://github.com/laravel/framework/pull/19771 for Laravel users

BrandonSurowiec:
@lucasmichot Thank you!

browner12:
@GrahamCampbell could you explain a little bit what your concerns are?  At first glance this kinda seems like a cool thing for packages where it is difficult to subclass.

kylekatarnls:
Hi. It's an anti-pattern only according the pattern you use in your app. And I don't see the difference with other macro functions in Laravel. What about letting the user decide to use it or not?

PR #611, opened 2016-02-15
Semester

lucasmichot:
Add:
- `semester` getter
- `startOfSemester` function
- `endOfSemester` function


GrahamCampbell:
Not sure this is actually useful to bloat the core with. It's pretty easy for someone to just check the month is less than 7?

briannesbitt:
It also seems to be different everywhere in the world.  When I think semester I think Sept - Dec, Jan - Apr.

Issue #333, opened 2015-03-31
Comparison fails on microseconds


martinlindhe:
Hello.
When using the microsecond precision, comparion ignores this part;

Example & workaround solution here:

``` php
use Carbon\Carbon;

$arr = [
    Carbon::parse('2015/03/31 12:59:10.012203'),
    Carbon::parse('2015/03/31 12:59:10.034936'),
    Carbon::parse('2015/03/31 12:59:10.012393'),
];

usort($arr, function ($a, $b) {
    if ($a->eq($b)) {
        return $a->micro < $b->micro ? -1 : 1;
    }
    return $a->lt($b) ? -1 : 1;
});


var_dump($arr);
```

briannesbitt:
I get them in order...

```
array(3) {
  [0]=>
  object(Carbon\Carbon)#2 (3) {
    ["date"]=>
    string(26) "2015-03-31 12:59:10.012203"
    ["timezone_type"]=>
    int(3)
    ["timezone"]=>
    string(15) "America/Toronto"
  }
  [1]=>
  object(Carbon\Carbon)#4 (3) {
    ["date"]=>
    string(26) "2015-03-31 12:59:10.012393"
    ["timezone_type"]=>
    int(3)
    ["timezone"]=>
    string(15) "America/Toronto"
  }
  [2]=>
  object(Carbon\Carbon)#3 (3) {
    ["date"]=>
    string(26) "2015-03-31 12:59:10.034936"
    ["timezone_type"]=>
    int(3)
    ["timezone"]=>
    string(15) "America/Toronto"
  }
}
```

If you switch your comparison I get them back in desc order.

martinlindhe:
Using what code? The snippet i posted solves the bug i am trying to report by checking the microsecond when eq() returns true even though microsecond differs, so it's supposed to return them like you posted.

All three of those timestamps has eq === TRUE in my env. I figured it was to do with __toString() using the default formatting, which does not have a .u (microsecond) and thus all three

briannesbitt:
Ah gotcha.  Actually the implementation just uses the underlying comparison operators (==, <, >) and DateTime doesn't do anything with microseconds.

j16sdiz:
Any update on this? I am hitting the same bug.

http://php.net/manual/en/datetime.diff.php#114656

martinlindhe:
personally, i ditched the php code i had that was depending on this and rewrote it in golang; https://github.com/martinlindhe/subtitles

[...]

PR #614, opened 2016-02-17
[WIP] Microtime camparison

lucasmichot:
Microtime comparison 

TODO
- [ ] Dedicated tests (might need to add a microtime setter)
- [ ] Closest
- [ ] Farthest

---

Solution for #333

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -30,10 +30,11 @@ use Symfony\Component\Translation\Loader\ArrayLoader;
  * @property      int $hour
  * @property      int $minute
  * @property      int $second
+ * @property      int $milliSeconds
+ * @property      int $micro
  * @property      int $timestamp seconds since the Unix Epoch
  * @property      \DateTimeZone $timezone the current timezone
  * @property      \DateTimeZone $tz alias of timezone
- * @property-read int $micro
  * @property-read int $dayOfWeek 0 (for Sunday) through 6 (for Saturday)
  * @property-read int $dayOfYear 0 through 365
  * @property-read int $weekOfMonth 1 through 5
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -234,6 +258,11 @@ public function __construct($time = null, $tz = null)
             $time = $testInstance->toDateTimeString();
         }
 
+        if ($time === null || $time === 'now') {
+            $dateTime = new DateTime('now', static::safeCreateDateTimeZone($tz));
+            $time = $dateTime->format('Y-m-d H:i:s').'.'.substr(microtime(), 2, 6);
```
lsloan: Some time passes between the `DateTime` call on line 262 and the `microtime` call on 263.  It may be fractions of a second or whole seconds, depending on the speed of the system that executes the code.  It would be better to get a `DateTime` objects that includes fractional seconds, then use the `u` format to include it in the string:

```php
$dateTime = \DateTime::createFromFormat('U.u', strval(microtime(true)));
$time = $dateTime->format('Y-m-d H:i:s.u'); // include μs
```

The `u` format gives the fraction as microseconds (six decimal places).  To get only milliseconds (three decimal places), use `substr` to strip off the last three characters:

```php
$time = substr($dateTime->format('Y-m-d H:i:s.u'), 0, -3); // truncate μs to ms
```

src/Carbon/Carbon.php
```diff
@@ -524,6 +553,9 @@ public function __get($name)
             case $name === 'quarter':
                 return (int) ceil($this->month / 3);
 
+            case $name === 'milliSeconds':
+                return (int) ceil($this->micro / static::MICROSECONDS_PER_MILLISECOND);
```
lsloan: As a Carbon user, I'd rather see a way to get the fractions of the second, however many decimal places it may be.  I don't want to be limited to `milliSeconds` (Is the capitalization intentional?), microseconds, etc.  Better yet, it's a part of the seconds, so return seconds as a floating point number.

lsloan:
Ha!  Only _now_ do I notice the PR is closed.

PR #617, opened 2016-02-18
Use 'now' as default value in constructor

lucasmichot:
Use `'now'` as default value in constructor

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -214,7 +214,7 @@ class Carbon extends DateTime
      * @param string|null               $time
      * @param \DateTimeZone|string|null $tz
      */
-    public function __construct($time = null, $tz = null)
+    public function __construct($time = 'now', $tz = null)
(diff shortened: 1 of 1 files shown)


src/Carbon/Carbon.php
```diff
@@ -211,10 +211,10 @@ protected static function safeCreateDateTimeZone($object)
      * Please see the testing aids section (specifically static::setTestNow())
      * for more on the possibility of this constructor returning a test instance.
      *
-     * @param string|null               $time
+     * @param string                    $time
      * @param \DateTimeZone|string|null $tz
      */
-    public function __construct($time = null, $tz = null)
+    public function __construct($time = 'now', $tz = null)
```
dereuromark: Usually you do not want to type now every time you just want to pass the timezone.
So default null is not a bad thing IMO.

lucasmichot: Indeed

PR #620, opened 2016-02-25
Tr translate fix

murat-yuksel:
Meaning of "from now" on dictionary is not suitable for the usage of package like the example; "3 yıl andan itibaren" (translation of 3 years from now) doesn't make sense. It should be "sonra" as both Google translate and jenssegers/date use it.


lucasmichot:
@murat-yuksel, it seems that you should also update the tests

murat-yuksel:
Sorry, i think its ok now

lucasmichot:
:+1:

PR #629, opened 2016-03-10
choose the format when transforming the date to a string

djug:


lucasmichot:
@djug , it's looking a bit hacky, because you could then do `$carbon->toDateString('H')`.
This function, supposed to represent a date, would return the hour

lucasmichot:
Ping @djug

djug:
@lucasmichot  I don't see any problem with that

lucasmichot:
There is no tests

briannesbitt:
I don't see the need for this.  Just use the format() method directly if
you need something specific.

If you need something specific often then you can use
Carbon::setToStringFormat().

PR #646, opened 2016-04-14
Add isCurrentYear, isCurrentMonth, isSameYear, isSameMonth methods

MaxGiting:
Return boolean if the date is in this current year.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1576,6 +1576,16 @@ class Carbon extends DateTime
     }
 
     /**
+     * Determines if the instance is in the current year
+     *
+     * @return bool
+     */
+    public function isCurrentYear()
+    {
+        return $this->year === static::now($this->tz)->year;
+    }
+
+    /**
      * Checks if the passed in date is the same day as the instance current day.
      *
      * @param \Carbon\Carbon $dt
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1632,11 +1632,8 @@ class Carbon extends DateTime
      */
     public function isSameMonth(Carbon $dt = null, $ofSameYear = false)
     {
-        if ($ofSameYear) {
-            return $this->isSameAs('Ym', $dt);
-        } else {
-            return $this->isSameAs('m', $dt);
-        }
+        $format = $ofSameYear ? 'Y-m' : 'm';
+        return $this->isSameAs($format, $dt);
     }
 
     /**
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1579,7 +1579,7 @@ class Carbon extends DateTime
      * Compares the formatted values of the two dates.
      *
      * @param string              $format The date formats to compare.
-     * @param \Carbon\Carbon|null $dt The instance to compare with or null to use current day.
+     * @param \Carbon\Carbon|null $dt     The instance to compare with or null to use current day.
      *
      * @return bool
      */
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -1576,6 +1576,16 @@ public function isLeapYear()
     }
 
     /**
+     * Determines if the instance is in the current year
+     *
+     * @return bool
+     */
+    public function isCurrentYear()
+    {
+        return $this->year === static::now($this->tz)->year;
```
lucasmichot: I think you can directly call `format` as for `\Carbon\Carbon::isBirthday`: this avoids calling `\Carbon\Carbon::__get`

MaxGiting: You're right. I liked the readability and didn't realise it took an indirect route to `format`

lucasmichot:
@MaxGiting I think there should be a companion function to this PR : `isSameYear` (on the model of `isBirthday`
Then the `isCurrenYear` could make use of `IsSameYear`...
Just my opinion

MaxGiting:
@lucasmichot I think you are right. The original solution arose from a specific scenario I had, but with `isSameYear` it could be applied to a much broader set. Plus why not have the same for the months while I am at it! Thanks. The joys of open source collaboration :+1:

src/Carbon/Carbon.php
```diff
@@ -1576,6 +1576,50 @@ public function isLeapYear()
...
+    /**
+     * Checks if the passed in date is in the same year as the instance year.
+     *
+     * @param \Carbon\Carbon $dt The instance to compare with.
+     *
+     * @return bool
+     */
+    public function isSameYear(Carbon $dt)
+    {
+        return $this->format('Y') === $dt->format('Y');
```
lucasmichot: You should also accept `null` as an argument - see `isBirthday`

src/Carbon/Carbon.php
```diff
@@ -1576,6 +1576,50 @@ public function isLeapYear()
...
+
+    /**
+     * Checks if the passed in date is in the same month as the instance month.
+     *
+     * @param \Carbon\Carbon $dt The instance to compare with.
+     *
+     * @return bool
+     */
+    public function isSameMonth(Carbon $dt)
+    {
```
lucasmichot: You should also accept `null` as an argument - see `isBirthday`

src/Carbon/Carbon.php
```diff
@@ -1576,6 +1576,50 @@ public function isLeapYear()
...
+        return $this->format('m') === $dt->format('m');
```
lucasmichot: In make sense, but I think passing a bool `$ofSameYear` would be useful: so that we can check if this the same month of the same year.

lucasmichot:
Some comments, but :+1:  for me

src/Carbon/Carbon.php
MaxGiting: agreed for both

src/Carbon/Carbon.php
MaxGiting: Sounds like a useful tidbit. Feels like you could keep on going with format comparison.

lucasmichot: Yep. Stg like `$format = $ofSameYear ? 'Y-m' : 'm';`

MaxGiting:
I agreed with your comments @lucasmichot I also felt that the isSame comparison methods using `format` were all so similar they could be extracted out into `isSameAs`, but I am not 100% happy with that method name. Plus leaves a lot more flexibility in the future.

PR #647, opened 2016-04-14
Use new Composer OR syntax

lucasmichot:
Use new Composer OR syntaxt : `||`


GrahamCampbell:
Personally it's a :-1: from me. :trollface: There's nothing wrong with the old one, and it's not deprecated.

lucasmichot:
So if this ever turns to be deprecated, it will be done

PR #653, opened 2016-04-21
Update Carbon.php

saruman:
using my setlocale(LC_TIME, 'es') and formating my date with "formatLocalized", the output is not well (mi�rcoles, 06 de enero de 2016), so I googled and found this solution:
http://stackoverflow.com/questions/8993971/php-strftime-french-characters

Now is working properly: miércoles, 06 de enero de 2016

This problem will occur for all those languages with special characters.

A better solution will be using setlocale(LC_TIME, 'es.UTF-8') but strftime don't understand that.  Any suggestions?

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1038,7 +1038,7 @@ class Carbon extends DateTime
             $format = preg_replace('#(?<!%)((?:%%)*)%e#', '\1%#d', $format);
         }
 
-        return strftime($format, strtotime($this));
+        return utf8_encode(strftime($format, strtotime($this)));
     }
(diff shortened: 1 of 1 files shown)


lucasmichot:
@saruman
Can you provide the code you use to confirm the bug:
Which date are you using ?
I cannot reproduce it

saruman:
Hey, thanks for reply.

Here is the code:
setlocale(LC_TIME, 'es');
echo Carbon::parse('2016-01-06')->formatLocalized('%A, %d \\de %B \\de %Y');

briannesbitt:
Can you create a test for this?

PR #661, opened 2016-05-10
Update fa.php

hafezdivandari:
Farsi translations of 'after' and 'before' switched to correct the meanings.


lucasmichot:
Thanks @hafezdivandari , can you also update the localization tests for Farsi ?

hafezdivandari:
Done.

PR #664, opened 2016-05-17
Implement new method to compare only the date

luqmanrom:


lucasmichot:
Can you add tests ?

src/Carbon/Carbon.php
```diff
@@ -1260,6 +1260,19 @@ public function equalTo(Carbon $dt)
...
+     * Determines if the instance date is equal to another
+     * instance date
+     *
+     * @param Carbon $dt
+     *
+     * @return bool
+     */
+    public function eqDate(Carbon $dt)
+    {
+        return $this->toDateString() == $dt->toDateString();
```
lucasmichot: We use strict comparisons in this project !

src/Carbon/Carbon.php
```diff
@@ -1260,6 +1260,19 @@ public function equalTo(Carbon $dt)
...
 
     /**
+     * Determines if the instance date is equal to another
+     * instance date
+     *
+     * @param Carbon $dt
+     *
+     * @return bool
+     */
+    public function eqDate(Carbon $dt)
```
lucasmichot: I would name this function `isDate` instead

src/Carbon/Carbon.php
lucasmichot: `toDateString` calls `format`
Why not calling `format` directly ?

lucasmichot:
@luqmanron , see https://github.com/briannesbitt/Carbon/pull/664#issuecomment-219750442

PR #665, opened 2016-05-24
Updated PT locale. Correct semantics.

gdiasdasilva:


gdiasdasilva:
This was implemented in the pt_BR version. It should be like this in the portuguese PT, too.

PR #668, opened 2016-05-26
Soft Fail on createFromFormat Function

john-main-croud:
There are two reasons for this:
Firstly to match the core PHP createFromFormat behaviour
Secondly to work better with Laravel when MySQL timestamp columns have '0000-00-00 00:00:00' values.
Currently if you put a date column in the Eloquent model's 'dates' array and try to access it then it will automatically call Carbon's createFromFormat function to parse that column into a Carbon object. As such when the column has a '0000-00-00 00:00:00' value you get get an exception thrown, merely from trying to access the value. A soft fail in this case is a much more useful and proportionate response.

I have added the ability to return the original error(s) via a pass-by-reference argument.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -426,12 +426,11 @@ class Carbon extends DateTime
      * @param string                    $format
      * @param string                    $time
      * @param \DateTimeZone|string|null $tz
+     * @param array                     $errors
      *
-     * @throws InvalidArgumentException
-     *
-     * @return static
+     * @return static|false
      */
-    public static function createFromFormat($format, $time, $tz = null)
+    public static function createFromFormat($format, $time, $tz = null, &$errors = [])
     {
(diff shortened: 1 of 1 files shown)


lucasmichot:
@john-main-croud :

- this is a BC
- this change should then target Laravel, not Carbon
- the coding style is not respected
- the tests break

john-main-croud:
I dispute your first two points. As this class extends PHP's DateTime object, it should return in the same manner i.e. return false rather than throw an exception. I think everyone would agree that throwing an exception is rather an overreaction in this case.

I really just put the code in as a talking point - if there is support for this I'll tidy it up and sort the tests.

lucasmichot:
Nevertheless this is a BC, thousands of developper using this library expect to get `false` returned instead of an exception; but it could be an idea for version `2.0`

Up to @briannesbitt

PR #669, opened 2016-05-26
Add short time units for diffForHumans()

EazyServer:
This is an amazing work but was surprised doesn't support short tags! you would please add this feature :) thanks


lucasmichot:
@EazyServer your PR is broken

PR #671, opened 2016-05-31
Add support for short time units for human

EazyServer:
Can you please accept this PR (re-submitted) to add support for short time units for human? 
I have modified the english version and added the support for all other languages. 
Other languages need to be checked and updated by a developer speaking the intended language.
Thanks


lucasmichot:
@EazyServer, the tests fail.

EazyServer:
When I run the tests on local machine it shows no error (All tests pass successfully)!
carbon_test_pass

lucasmichot:
- Your IDE is not triggering the proper phpunit bin. It should be `./vendor/bin/phpunit` as described in the travis file
- there are no new tests with calls to `diffForHumans` with the short flag set to true
- the coding style is not respected

GrahamCampbell:
> (All tests pass successfully)!

No tests even ran...

EazyServer:
@lucasmichot 

- Not sure how changing phpunit bin would make a difference! but I changed it to the one in the vendor folder anyway and still show All passed!
- Could you please point where the coding styling is not respected?
- Will add the tests with short tag on later on.
Thanks
carbon_test_pass

lucasmichot:
@yarob-union 

> Not sure how changing phpunit bin would make a difference! but I changed it to the one in the vendor folder anyway and still show All passed!
The fact that youd don't execute any test for the moment as @GrahamCampbell mentionned
> Could you please point where the coding styling is not respected?
Please review StyleCI status
> Will add the tests with short tag on later on
Please also fix the failing ones, see TravisCI logs for details

EazyServer:
@lucasmichot 
The tests are being executed and they do pass, see when I try to fail one of the tests it does fail! which indicates the tests are actually being executed!!

carbon_test_pass

additionally, I did not add any substantial code logic apart from `$short?'a':'b'` and duplicated values in translation arrays for year, month etc. for short units.
I don't see how that would fail the tests! there are even no tests for short units activated yet! so it shouldn't be called anyway!

Do you think re-submitting the PR might resolve this bizarre case?

EazyServer:
@GrahamCampbell since your are the creator of @StyleCI could you please tell me why the formating is failing? I don't seem to recognize the issue!

Thanks

EazyServer:
Finally!

lucasmichot:
Can you squash your commits ?

src/Carbon/Carbon.php
```diff
@@ -2350,7 +2350,7 @@ public function secondsUntilEndOfDay()
      *
      * @return string
```
lucasmichot: `$short`argument is missing in the docblock

EazyServer: Oh I just seen this! will add the argument and then squash commits.

EazyServer:
commits squashed :)

Issue #182, opened 2014-09-21
not working as expected with `json_encode()`


icylace:
I thought that a Carbon date object would automatically convert into a datetime string when using `json_encode()` on it.  It doesn't because it needs to implement the `JsonSerializable` interface to make it work.  I don't see an advantage to JSON-encoding the Carbon object itself rather than its standard string form.

briannesbitt:
The plan is to drop 5.3 support since its end of life.  Then we would add the JsonSerializable implementation along with a 2.0 release.

PR #673, opened 2016-06-08
Json serializable carbon objects

shooka:
I know that `JsonSerializable` is 5.4+ only, but I think it's time to put 5.3 to rest.

Here's how it works:

Normally when you `json_encode($carbon)`, it returns the entire `DateTime` object:

```
{
    "date":"2016-06-08 13:55:28.000000",
    "timezone_type":3,
    "timezone":"Europe\/Copenhagen"
}
```

The idea is that by setting the json format, the `jsonSerialize` method returns json in the specified format defaulting to the `DateTime` object.

The format is set like this:

```
Carbon::setToJsonFormat(Carbon::W3C);
```

Which makes `json_encode($carbon)` return `"2016-06-08T13:55:28+02:00"`

It is of course possible to reset the format using `Carbon::resetToJsonFormat()`


jycr753:
+1

lucasmichot:
Duplicates https://github.com/briannesbitt/Carbon/pull/600

shooka:
@lucasmichot No. This is serializing to json whereas #600 is serializing to PHP's internal format.

lucasmichot:
This is a BC and should target a V2.

jycr753:
version what? Carbon version 2?

shooka:
@lucasmichot You're right. We should probably create a new v2 alpha branch. What do you think?

lucasmichot:
I think so too, up to @briannesbitt

dereuromark:
> This is a BC and should target a V2.

Someone still confuses the two things. You probably mean BC breaking for the sentence to make sense :)

briannesbitt:
Yes a v2 thing.  As per #182 comments.

lucasmichot:
> Someone still confuses the two things. You probably mean BC breaking for the sentence to make sense :)

Indeed !

> Yes a v2 thing. As per #182 comments.

What about creating a V2 branch ?

shooka:
@GrahamCampbell Do you have any idea why the PR fails in StyleCI? I didn't change anything in the localization tests.

shooka:
@briannesbitt Have you made a decision regarding the v2 branch? :)

MrCrankHank:
Any news here? This would be really useful for things like boostraptable or datatable...

briannesbitt:
For the 1.x branch we can add this if you don't force the 5.3 and use the interface.  That way users on 5.3 can just call the function directly?  We would add those things back in for the 2.x branch.

Thoughts?

shooka:
@briannesbitt I get what you mean, but I don't see how this would benefit 5.3 developers. As you mention, developers using 5.3 would have to call the method directly. I actually think the existing solution is prettier.

Today:

```
json_encode((string) $carbon);
```

After merging this PR:

```
json_encode($carbon->jsonSerialize());
```

The two solutions are roughly the same. Don't you agree?

briannesbitt:
True... forgot you would still need to call to encode it.  2.x it is.

shooka:
@briannesbitt @lucasmichot Are we any closer to merging this? I opened the PR one year ago today.

jycr753:
@shooka is one year, don't worry much about it.

kylekatarnls:
Sorry for delay, it had been tagged for 2.x because it couldn't be merged in 1.x, dropping PHP 5.3 is a breaking change.

Moreover a lot of Carbon users are Laravel users, so I prefer to do it the same way Laravel does and with a fallback to not let it break PHP 5.3: #1207

So I close this issue because we'll start 2.0 from last version 1.

Thanks for your help, sorry it couldn't be merged. But now Carbon is about to be json-able on PHP 5.4+ and the rest of Carbon will continue to work on PHP 5.3.

kylekatarnls:
@jycr753 Maybe I was unclear. Now Carbon objects are serializable as JSON without having to drop PHP 5.3 compatibility. Is there something bothering you?

PR #681, opened 2016-06-22
Fix phpdoc to return Carbon instead of static

nicholasnet:
Currently Carbon returns static in terms of PhpDoc which causes an issue with resolved method in IDE.

```
class Foo
{
    /**
     * @var \Carbon\Carbon
     */
    protected $foo;

    public function __construct()
    {
        $this->foo = \Carbon\Carbon::now();
    }
}

class Bar extends Foo
{
    public function test()
    {
        $this->foo->format('Y-m-d'); // This won't resolve in IDE like PhpStorm
    }
}
```

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -242,7 +242,7 @@ class Carbon extends DateTime
      *
      * @param \DateTime $dt
      *
-     * @return static
+     * @return Carbon
      */
     public static function instance(DateTime $dt)
     {
@@ -263,7 +263,7 @@ class Carbon extends DateTime
      * @param string|null               $time
      * @param \DateTimeZone|string|null $tz
      *
-     * @return static
+     * @return Carbon
      */
     public static function parse($time = null, $tz = null)
     {
@@ -275,7 +275,7 @@ class Carbon extends DateTime
      *
      * @param \DateTimeZone|string|null $tz
      *
-     * @return static
+     * @return Carbon
      */
     public static function now($tz = null)
     {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This doesn't look correct to me?

lucasmichot:
Add an argument to format and this will just work fine

briannesbitt:
The opposite of this request was done before #124

lucasmichot:
It's a just à bad call of format function

GrahamCampbell:
Yeh, this PR is wrong and can be closed. It does NOT fix the IDE.

nicholasnet:
The intention was not to fix the IDE and I fixed the format function call though that is not the issue that this pull request is trying to resolve. @briannesbitt I saw the reverse of this PR was merged but I think the use case is still valid. Anyway it's your call.

shooka:
As the methods actually return `new static()...`, I think it would be wrong to change the return type in the docblocks to `Carbon`. This should be fixed by the IDE developers and in the meantime we just have to live with it.

See https://www.phpdoc.org/docs/latest/guides/types.html#keywords

lucasmichot:
Indeed @shooka @GrahamCampbell
Plus we are PSR-4, this should be `\Carbon\Carbon`

briannesbitt:
I think the return static are fine for the static methods.  What about moving to `@return $this` for the fluent interface methods?

lucasmichot:
@briannesbitt I feel TravisCI shall be re-enabled ASAP ;-)

PR #686, opened 2016-06-27
Fix 2 typos

lucasmichot:
Fix 2 typos


tests/Carbon/AddTest.php
```diff
@@ -257,7 +257,7 @@ public function tesAddQuarter()
         $this->assertSame(8, Carbon::createFromDate(1975, 5, 6)->addQuarter()->month);
     }
 
-    public function tesAddQuarterNegatve()
+    public function tesAddQuarterNegative()
```
juliangut: Still a typo, should be "test...()" `testAddQuarterNegative`

briannesbitt:
Good catch... there are 4 of them actually :-)

PR #694, opened 2016-07-04
Simplify createSafe

lucasmichot:
Simplify `\Carbon\Carbon::createSafe`


shooka:
I would still change the name to `safeCreate` as it is a safe creation and not creating safes 😉 

Also to match the existing `safeCreateDateTimeZone`

lucasmichot:
The goal is to simplify the function.
Concerning the name of the function itself, it should probably be in another PR

shooka:
You are right. I just thought that since it hasn't been released yet and you were the creator of the method, you would have special privileges 😄

PR #706, opened 2016-07-06
Created extensions adding business day functions

rmblstrp:
I created a set of extension methods which provides a flexible set of
methods that allow the determination if a date is a business day as
well as determining the next or previous business day. Custom holiday
and business schedules can be supplies to the methods to allow an
end-user to determine what holidays, if any, are to be observed and the
days that fit operating schedule of the business being modeled.

**NOTE:** I know that this functionality has been asked for before but it seemed that you didn't want to include the functionality into the Carbon class. I felt that extending Carbon would be a bad idea because it would lock implementations to a certain subclass and created a hierarchical mess if there were multiple subclasses with some users wanting functionality that was contained in 2 or more subclasses. Because of those reasons I opted to create an extension class where other contributors could add in functionality to extend Carbon while keeping the Carbon class pure. If that functionality is deemed worthy enough to be promoted to the Carbon class it can be done easily without breaking backwards compatibility and without having duplication of code.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -45,6 +45,22 @@ echo Carbon::now()->subMinutes(2)->diffForHumans(); // '2 minutes ago'
 // rolling up to seconds, minutes, hours, days, months, years
 
 $daysSinceEpoch = Carbon::createFromTimestamp(0)->diffInDays();
+
+// Check to see if a date is a business day defaulting to a Monday
+// through Friday work week
+$today = Carbon::now();
+CarbonExtensions::isBusinessDay($today);
+
+// Use a custom operating schedule by implementing the BusinessSchedule
+// interface
+$schedule = new OperatingSchedule();
+CarbonExtensions::isBusinessDay($today, $schedule);
+
+// Create custom holiday schedules by implementing the HolidaySchedule
+// interface so that observed holidays can be excluded as business days
+$holidays = new Holidays($today->year);
+CarbonExtensions::isBusinessDay($today, $holidays);
+
 ```
 
 ## Installation
--- a/src/Carbon/CarbonExtensions.php
+++ b/src/Carbon/CarbonExtensions.php
@@ -1,0 +1,160 @@
+<?php
+/*
+ * This file is part of the Carbon package.
+ *
+ * (c) Brian Nesbitt <brian@nesbot.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Carbon;
+
+/**
+ * Class CarbonExtensions
+ */
+class CarbonExtensions
+{
+    /**
+     * Checks the date to see if it is a business day
+     *
+     * @author Christopher McGinnis <cmcginnis@tuition.io>
+     *
+     * @param Carbon                                $date
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return bool
+     */
+    public static function isBusinessDay(Carbon $date, $schedule = null)
+    {
+        $isHoliday = $schedule instanceof HolidaySchedule
+            ? $schedule->isHoliday($date)
+            : false
+        ;
+
+        $isBusinessDay = $schedule instanceof BusinessSchedule
+            ? $schedule->isBusinessDay($date)
+            : $date->isWeekday()
+        ;
+
+        return $isBusinessDay && !$isHoliday;
+    }
+
+    /**
+     * Sets the date to the next business day
+     *
+     * @author Christopher McGinnis <cmcginnis@tuition.io>
+     *
+     * @param Carbon                                $date
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return Carbon
+     */
+    public static function nextBusinessDay(Carbon $date, $schedule = null)
+    {
+        do {
+            $date->addDay();
+        } while (!self::isBusinessDay($date, $schedule));
+
+        return $date;
+    }
+
+    /**
+     * Sets the date to the current or next business day
+     *
+     * @author Christopher McGinnis <cmcginnis@tuition.io>
+     *
+     * @param Carbon                                $date
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return Carbon
+     */
+    public static function currentOrNextBusinessDay(Carbon $date, $schedule = null)
+    {
+        if (!self::isBusinessDay($date, $schedule)) {
+            self::nextBusinessDay($date, $schedule);
+        }
+
+        return $date;
+    }
+
+    /**
+     * Sets the date to the previous business day
+     *
+     * @author Christopher McGinnis <cmcginnis@tuition.io>
+     *
+     * @param Carbon                                $date
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return Carbon
+     */
+    public static function previousBusinessDay(Carbon $date, $schedule = null)
+    {
+        do {
+            $date->subDay();
+        } while (!self::isBusinessDay($date, $schedule));
+
+        return $date;
+    }
+
+    /**
+     * Sets the date to the current or previous business day
+     *
+     * @author Christopher McGinnis <cmcginnis@tuition.io>
+     *
+     * @param Carbon                                $date
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return Carbon
+     */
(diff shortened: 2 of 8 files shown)


readme.md
```diff
@@ -45,6 +45,22 @@ echo Carbon::now()->subMinutes(2)->diffForHumans(); // '2 minutes ago'
 // rolling up to seconds, minutes, hours, days, months, years
 
 $daysSinceEpoch = Carbon::createFromTimestamp(0)->diffInDays();
+
+// Check to see if a date is a business day defaulting to a Monday
+// through Friday work week
+$today = Carbon::now();
+CarbonExtensions::isBusinessDay($today);
```
lucasmichot: `\Carbon\Carbon::isWeekday` already exists for that

rmblstrp: isWeekday is not the same as isBusinessDay.  Weekdays are Monday-Friday whereas a business might only be open Friday-Sunday. Additionally, isBusinessDay checks to see if the day is a holiday if a holiday schedule is passed in as a parameter.

For example: A business operates on a Monday - Friday schedule. The current date is on a Friday and the following Monday is a holiday. You need to know what the date is going to be for 3 business days from now. Since the following Monday is a holiday the date would be for the following Thursday.

While the basic functionality of isBusinessDay does the same as isWeekday, the difference is that it allows for custom business and holiday schedules.  If you take a look at the unit tests you'll find examples of how this is used.

lucasff:
+1. Using setWeekdays and afterwards using addWeekdays seems to not have an effect at all.

lucasmichot:
@briannesbitt whats your opinion on that?
I feel this should probably target `2.x`

lucasmichot:
@rmblstrp It seems the test are failing...
Can you squash and rebase ?

src/Carbon/CarbonExtensions.php
```diff
@@ -0,0 +1,160 @@
...
+ * (c) Brian Nesbitt <brian@nesbot.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Carbon;
+
+/**
+ * Class CarbonExtensions
```
lucasmichot: Can you remove this class docblock ?

src/Carbon/CarbonExtensions.php
```diff
@@ -0,0 +1,160 @@
...
+ */
+class CarbonExtensions
+{
+    /**
+     * Checks the date to see if it is a business day
+     *
+     * @author Christopher McGinnis <cmcginnis@tuition.io>
```
lucasmichot: Can you also remove the author docblock

src/Carbon/CarbonExtensions.php
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @param Carbon                                $date
```
lucasmichot: for param, return and throw docblocks we use FQN class name

src/Carbon/CarbonExtensions.php
```diff
@@ -0,0 +1,160 @@
...
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return Carbon
+     */
+    public static function nextBusinessDay(Carbon $date, $schedule = null)
+    {
+        do {
+            $date->addDay();
+        } while (!self::isBusinessDay($date, $schedule));
```
lucasmichot: Can you call with `static::` instead ?

src/Carbon/CarbonExtensions.php
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @param Carbon                                $date
+     * @param int                                   $days
+     * @param HolidaySchedule|BusinessSchedule|null $schedule
+     *
+     * @return Carbon
+     */
+    public static function addBusinessDays(Carbon $date, $days, $schedule = null)
+    {
+        while ($days != 0) {
```
lucasmichot: Any reason not to use a stricter comparison `!==` ?

src/Carbon/HolidaySchedule.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Carbon;
+
+interface HolidaySchedule
+{
+    public function isHoliday(Carbon $date);
```
lucasmichot: Docblock?

tests/CarbonExtensions/Fixtures/FederalHolidays.php
```diff
@@ -0,0 +1,206 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Tests\CarbonExtensions\Fixtures;
+
+use Carbon\Carbon;
+use Carbon\HolidaySchedule;
+
+/**
+ * Class FederalHolidays
```
lucasmichot: Can you remove the class name declaration in this docblck ?

tests/CarbonExtensions/Fixtures/FederalHolidays.php
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @property-read int $year;
+ * @property int $observedHolidays;
+ * @property-read Carbon $newYearsDay;
```
lucasmichot: `\Carbon\Carbon` instead

tests/CarbonExtensions/Fixtures/FederalHolidays.php
```diff
@@ -0,0 +1,206 @@
...
+        $this->thanksgivingDay = $this->getThanksgivingDay();
+        $this->christmasDay = $this->getChristmasDay();
+    }
+
+    /**
+     * @param Carbon $date
+     *
+     * @returns bool
+     */
+    public function isHoliday(Carbon $date)
```
lucasmichot: Carbon $carbon instead

tests/CarbonExtensions/Fixtures/FederalHolidays.php
```diff
@@ -0,0 +1,206 @@
...
+ * @property-read Carbon $independenceDay;
+ * @property-read Carbon $laborDay;
+ * @property-read Carbon $columbusDay;
+ * @property-read Carbon $veteransDay;
+ * @property-read Carbon $thanksgivingDay;
+ * @property-read Carbon $christmasDay;
+ */
+class FederalHolidays implements HolidaySchedule
+{
+    const NEW_YEARS_DAY = 0x0001;
```
lucasmichot: Missing docblocks for const and properties

tests/CarbonExtensions/Fixtures/FederalHolidays.php
```diff
@@ -0,0 +1,206 @@
...
+     * @param Carbon $date
+     *
+     * @returns bool
+     */
+    public function isHoliday(Carbon $date)
+    {
+        // created a copy of the datetime object to we can remove the time values without affecting
+        // the passed in instance
+
+        return ($this->newYearsDay->isSameDay($date) && $this->isObserved(self::NEW_YEARS_DAY))
```
lucasmichot: I would separate this huge conditions in smaller conditions

tests/CarbonExtensions/Fixtures/FederalHolidays.php
```diff
@@ -0,0 +1,206 @@
...
+    {
+        return ($holiday & $this->observedHolidays) == $holiday;
+    }
+
+    /**
+     * @return Carbon
+     */
+    protected function getNewYearsDay()
+    {
+        return Carbon::create($this->year, 1, 1, 0, 0, 0);
```
lucasmichot: Why not calling Carbon::now()->startOfYear()?

briannesbitt:
Seems possible, but waiting to v2 might be a good idea.

Not sure why "CarbonExtensions" is so generic where the interface is specific.  Seems like a more specific class name makes more sense ?

rmblstrp:
@briannesbitt Agreed. I put this together because we needed this functionality in a work projects.  I'll retool it a bit so it makes more sense and get it ready for the 2.x release.

ashilr:
222

lucasff:
up?

kylekatarnls:
Hi, I started this library (compatible with the coming soon 1.26 version) based on @rmblstrp work:
https://github.com/kylekatarnls/business-day

You're all invited to test, contribute, add holidays lists, etc.

kylekatarnls:
I close this issue as kylekatarnls/business-day should fulfill the need. If something is missing please open an issue here: https://github.com/kylekatarnls/business-day/issues

PR #717, opened 2016-07-21
Fix more interval tests

lucasmichot:
Follow up #643


briannesbitt:
Ha... jumped the gun on the other one.  Thanks.

PR #722, opened 2016-07-22
Fix \Carbon\Exceptions\InvalidDateException

lucasmichot:
Fix `\Carbon\Exceptions\InvalidDateException`

---

**Note:**
Not sure there should be setters either @briannesbitt ? I feel like just settings the properties in the constructor would be fine.


lucasmichot:
Ping @briannesbitt

briannesbitt:
Ya, I am good with no setters as the usage is typically `throw new InvalidDateException($field, $value);`

PR #723, opened 2016-07-22
PHP-CS-Fixer is unneeded and will just conflict with StyleCI

GrahamCampbell:
One of StyleCI's key goals is to eliminate messing about with local cs dev tools.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,6 @@
         "symfony/translation": "~2.6|~3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "~1.11",
         "phpunit/phpunit": "~4.0|~5.0"
     },
     "autoload": {
@@ -44,8 +43,6 @@
         }
     },
     "scripts": {
-        "test": "./vendor/bin/phpunit; ./vendor/bin/php-cs-fixer fix -v --diff --dry-run;",
-        "phpunit": "./vendor/bin/phpunit;",
-        "phpcs": "./vendor/bin/php-cs-fixer fix -v --diff --dry-run;"
+        "test": "./vendor/bin/phpunit;"
     }
 }


GrahamCampbell:
Also, PHP-CS-Fixer may yield different results to StyleCI meaning you just keep going backwards and forwards with cs fixes.

lucasmichot:
I disagree, this repository uses Travis-CI, does not mean we should not test locally too

jbrooksuk:
@lucasmichot the point of StyleCI is not to need to fix your code locally as it's fixed at the PR level.

lucasmichot:
@jbrooksuk yes I know. 
This repository is the perfect example of relying on php-cs-fixer AND styleci.
People were fixing StyleCI errors and thus many conflicts occured.

I've already had this talk with Graham before, and we disagree on this point.

I feel the reason why projects such as laravel/framework still provides php-cs-fixer support are mostly the same as the ones I give.

kylekatarnls:
My two cents: do StyleCI use php-cs-fixer? Else they can check different style rules and do not replace each other. Another point is: with a local check you can work offline and push directly a cleaner code before waiting the remote check to be complete.

jbrooksuk:
As Graham said though, there are issues with supporting both:

1. PHP-CS-Fixer does one thing, StyleCI does another. You can't have them both simply because they'll spend time fixing each others fixes.
2. StyleCI takes care of any syntax issues for you, the whole point is that you don't need to worry about that.

GrahamCampbell:
> do StyleCI use php-cs-fixer

Not exactly, no. The output is not the same as php-cs-fixer 1.11.x.

GrahamCampbell:
> My two cents: do StyleCI use php-cs-fixer? 

Not exactly, no. The output is not the same.

PR #728, opened 2016-07-22
Simplify __set

lucasmichot:
And no need to call `__get` twice.


briannesbitt:
I'll merge it but I am not sure this actually simplifies it ;-)  Less lines of code though is a positive.

PR #729, opened 2016-07-23
Factorize previous- and next- week- or weekend day

lucasmichot:
Factorize previous- and next- week- or weekend day

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2986,7 +2986,7 @@ class Carbon extends DateTime
      */
     private function nextOrPreviousDay($weekday = true, $forward = true)
     {
-        $step = $forward ? 1 : - 1;
+        $step = $forward ? 1 : -1;
 
         do {
             $this->addDay($step);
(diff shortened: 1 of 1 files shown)


src/Carbon/Carbon.php
```diff
@@ -2881,31 +2881,40 @@ public function next($dayOfWeek = null)
...
-     * Go forward to the next weekday.
+     * Go forward or backward to the next week- or weekend-day.
+     *
+     * @param bool $weekday
+     * @param bool $forward
      *
      * @return $this
      */
-    public function nextWeekday()
+    private function nextOrPreviousDay($weekday = true, $forward = true)
```
kylekatarnls: Shouldn't be "protected" to allow to use it when we extends Carbon?

briannesbitt:
I'd make the helper method private / protected again.  Don't want anyone public calling it.

lucasmichot:
Fixed with https://github.com/briannesbitt/Carbon/pull/729/commits/cd0433b8d585805297b952c8bca9b369a44c8644

PR #746, opened 2016-08-05
German Short Texts according to DIN 1355-1

BlueTeck:
https://de.wikipedia.org/wiki/DIN_1355-1


lucasmichot:
Tests?

BlueTeck:
There aren't any tests for shorttexts, should I create them in https://github.com/briannesbitt/Carbon/blob/master/tests/Localization/DeTest.php ?

BlueTeck:
Tests added

PR #752, opened 2016-08-11
New is methods for next / last week, month and year

arjasco:
Added the new methods `isNextWeek, isLastWeek, isNextMonth, isLastMonth, isNextYear, isLastYear`

Use case was to has a quick method when sorting data into last, current and next week segments, so seemed sensible to have it for month and year also.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1650,6 +1650,66 @@ class Carbon extends DateTime
     }
 
     /**
+     * Determines if the instance is within the next week
+     *
+     * @return bool
+     */
+    public function isNextWeek()
+    {
+        return $this->weekOfYear === static::now($this->getTimezone())->addWeek()->weekOfYear;
+    }
+
+    /**
+     * Determines if the instance is within the last week
+     *
+     * @return bool
+     */
+    public function isLastWeek()
+    {
+        return $this->weekOfYear === static::now($this->getTimezone())->subWeek()->weekOfYear;
+    }
+
+    /**
+     * Determines if the instance is within the next month
+     *
+     * @return bool
+     */
+    public function isNextMonth()
+    {
+        return $this->month === static::now($this->getTimezone())->addMonth()->month;
+    }
+
+    /**
+     * Determines if the instance is within the last month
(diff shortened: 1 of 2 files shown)


lucasmichot:
Shouldn't `previous` be used instead of `last` ?

arjasco:
That makes sense to me, I did actually put in the comment above the method previous! I will change.

src/Carbon/Carbon.php
```diff
@@ -1650,6 +1650,66 @@ public function isTomorrow()
     }
 
     /**
+     * Determines if the instance is within the next week
```
lucasmichot: All methods descriptions should finish with a point:
`Determines if the instance is within the next week.` ;-)

src/Carbon/Carbon.php
```diff
@@ -1650,6 +1650,66 @@ public function isTomorrow()
...
+    }
+
+    /**
+     * Determines if the instance is within the next month
+     *
+     * @return bool
+     */
+    public function isNextMonth()
+    {
+        return $this->month === static::now($this->getTimezone())->addMonth()->month;
```
lucasmichot: Unfortunatelly this is not always working:
```php
Carbon::setTestNow(Carbon::create(2016, 1, 31));
$this->assertTrue(Carbon::create(2016, 2, 27)->isNextMonth());
```

Months comparison should use `addMonthsNoOverflow` and `subMonthsNoOverflow`.

tests/Carbon/IsTest.php
```diff
@@ -56,6 +56,66 @@ public function testIsTodayTrue()
         $this->assertTrue(Carbon::now()->isToday());
     }
 
+    public function testIsNextWeekTrue()
+    {
+        $this->assertTrue(Carbon::now()->addWeek()->isNextWeek());
```
lucasmichot: All these tests should should handle overflow.

lucasmichot:
Ping @arjasco

arjasco:
@lucasmichot Sorry, it's been a while! really appreciate you fixing this up and getting it merged :thumbsup:

PR #753, opened 2016-08-13
Add ukraine language, create file ua.php

Greenelf:
Add Ukraine Language


lucasmichot:
@Greenelf , can you:
- Add the short notations
- Fix the coding style
- Add tests

PR #771, opened 2016-09-13
Added a "setUtf8" method for localization

Max13:
Completes #653 

There is not much other way to "simulate" UTF8 than actually "expecting" an UTF8 encoded string.

Ah, and because of the comments in `StringsTest::testToLocalizedFormattedDateString()`, I think my test will also have the same issue with Travis.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -171,6 +171,13 @@ class Carbon extends DateTime
     private static $lastErrors;
 
     /**
+     * Is UTF8 needed to print localized date/time ?
+     *
+     * @var bool
+     */
+    protected static $utf8 = false;
+
+    /**
      * Creates a DateTimeZone from a string, DateTimeZone or integer offset.
      *
      * @param \DateTimeZone|string|int|null $object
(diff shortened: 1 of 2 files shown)


lucasmichot:
Can you solve the conflicts @Max13 ?

Max13:
@lucasmichot Done, sorry I didn't checkout `HEAD` first.

src/Carbon/Carbon.php
```diff
@@ -1122,6 +1129,16 @@ public static function setLocale($locale)
         return false;
     }
 
+    /**
+     * Set if UTF8 will be used for localized date/time
+     *
+     * @param bool $useUtf8
+     */
```
lucasmichot: Missing `@return void`

src/Carbon/Carbon.php
```diff
@@ -1142,6 +1159,10 @@ public function formatLocalized($format)
             $format = preg_replace('#(?<!%)((?:%%)*)%e#', '\1%#d', $format);
         }
 
+        if (static::$utf8) {
+            return utf8_encode(strftime($format, strtotime($this)));
```
lucasmichot: `strftime($format, strtotime($this))` is repeated twice - maybe extract this to a `$value` variable ?

src/Carbon/Carbon.php
```diff
@@ -1122,6 +1129,18 @@ public static function setLocale($locale)
...
     }
 
+    /**
+     * Set if UTF8 will be used for localized date/time
+     *
+     * @param bool $useUtf8
+     *
+     * @return void
+     */
+    public static function setUtf8($useUtf8)
```
lucasmichot: I would rename this to `useUtf8()`

Max13: You're the boss

lucasmichot:
Needs rebase @Max13

briannesbitt:
Already made these changes... did see this PR at that point.

PR #780, opened 2016-10-05
Use DateTimeInterface instead of DateTime

luismulinari:


lucasmichot:
Tests failing @luismulinari : https://travis-ci.org/briannesbitt/Carbon/builds/165118572

dereuromark:
Isn't that a dup of https://github.com/briannesbitt/Carbon/pull/233 and others?
Currently that is not possible without BC breaking, as it requires newer PHP versions.
If you want to use interfaces here, you probably want to take a look at [Chronos](https://github.com/cakephp/chronos/blob/master/src/ChronosInterface.php#L21).

luismulinari:
You are right. Sorry about my mistake. I did not check the composer.json to see the minimum version required of PHP.

I'll take a look at Chronos. Thanks :)

PR #786, opened 2016-11-04
Removed duplicate array key

nurolopher:
There were two "m" array keys. Removed one of them.


src/Carbon/Lang/bg.php
```diff
@@ -13,7 +13,6 @@
     'year' => '1 година|:count години',
     'y' => '1 година|:count години',
     'month' => '1 месец|:count месеца',
-    'm' => '1 месец|:count месеца',
```
nurolopher: @lucasmichot ,  compare this line with 23rd line.

PR #787, opened 2016-11-07
Also test PHP 7.1 with Travis

lucasmichot:
Also test PHP 7.1 with Travis


lucasmichot:
@briannesbitt , Travis CI seems disabled, that would be awesome to re-enable it !

PR #819, opened 2016-11-25
Add Montenegrin language translation

milose:

PR #825, opened 2016-12-05
Allow Carbon::setLocale to accept long localization variants

milose:
As discussed in #819 there are language variants that allow script type to be passed in the format `ll-Ssss-RR` (`l` being language code, `s` being script used, `r` being the region).

According to http://symbolcodes.tlt.psu.edu/bylanguage/serbocroatian.html website, there are three variants for Montenegrin locale:
- `sr-ME` (Montenegrin, script unspecified)
- `sr-Cyrl-ME` (Montenegrin in Cyrillic)
- `sr-Latn-ME` (Montenegrin in Latin Script)

This reflects the situation in real life, as both scripts (Latin and Cyrillic) are equally used in Montenegrin, Serbian, Croatian and Bosnian languages - as far as I know.

This is a non-breaking change.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1138,8 +1138,8 @@ class Carbon extends DateTime
      */
     public static function setLocale($locale)
     {
-        $locale = preg_replace_callback('/([a-z]{2})[-_]([a-z]{2})/', function ($matches) {
-            return $matches[1].'_'.strtoupper($matches[2]);
+        $locale = preg_replace_callback('/\b([a-z]{2})[-_](?:([a-z]{4})[-_])?([a-z]{2})\b/', function ($matches) {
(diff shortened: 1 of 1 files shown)


lucasmichot:
Tests ?

milose:
@lucasmichot Given the current language coverage this cannot be tested.

lucasmichot:
@milose I feel this new locales should be at least tested in `\Tests\Carbon\LocalizationTest::providerLocales`

milose:
@lucasmichot That cannot be tested because `Carbon::setLocale($locale)` asserts dataProvider `\Tests\Carbon\LocalizationTest::providerLocales` feeds existing localization files. In order for file to exist, this should be merged first, then we can add tests for individual languages that are added.

PR #842, opened 2017-01-08
Integrating PHPStan & fixes of found issues

ondrejmirtes:
This is a PR for integrating static analysis tool [PHPStan](https://github.com/phpstan/phpstan) that focuses on finding bugs and running in your CI build. It found a few issues that I put in separate commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,9 @@ install:
   - if [[ $setup = 'stable' ]]; then travis_retry composer update --prefer-dist --no-interaction --no-suggest --prefer-stable; fi
   - if [[ $setup = 'lowest' ]]; then travis_retry composer update --prefer-dist --no-interaction --no-suggest --prefer-stable --prefer-lowest; fi
 
-script: vendor/bin/phpunit --verbose --coverage-clover=coverage.xml
+script:
+  - vendor/bin/phpunit --verbose --coverage-clover=coverage.xml
+  - if [[ $PHPSTAN = 'true' ]]; then composer require --dev phpstan/phpstan:^0.6 && vendor/bin/phpstan analyse --configuration phpstan.neon --level 3 src tests; fi
 
 after_success:
   - if [[ $coverage = 'true' ]]; then bash <(curl -s https://codecov.io/bash); fi
@@ -55,12 +57,16 @@ matrix:
       env: setup=lowest
     - php: 7.0
       env: setup=stable
+    - php: 7.0
+      env: PHPSTAN=true
     - php: 7.1
       env: setup=lowest
     - php: 7.1
       env:
         - coverage=true
         - setup=stable
+    - php: 7.1
+      env: PHPSTAN=true
   allow_failures:
     - php: hhvm
   fast_finish: true
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,6 @@
+parameters:
+	ignoreErrors:
+		- '#Access to an undefined property Carbon\\Carbon::\$doesNotExit#'
+		- '#Call to an undefined static method Carbon\\CarbonInterval#'
+		- '#Property Carbon\\Carbon::\$timezone \(DateTimeZone\) does not accept string#'
+		- '#Property Carbon\\Carbon::\$tz \(DateTimeZone\) does not accept string#'
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1092,8 +1092,9 @@ class Carbon extends DateTime
     protected static function translator()
     {
         if (static::$translator === null) {
-            static::$translator = new Translator('en');
-            static::$translator->addLoader('array', new ArrayLoader());
+            $translator = new Translator('en');
+            $translator->addLoader('array', new ArrayLoader());
+            static::$translator = $translator;
             static::setLocale('en');
         }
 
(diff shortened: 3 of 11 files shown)


lucasmichot:
Hey @ondrejmirtes thanks for this PR (and also for this great PHPStan tool, that use on many of my other projects)
I am not totally sure that it make sense to break a build if unit tests pass though, or to integrate it in TravisCI

Nevertheless the fix are totally valuable

Thoughts on that @briannesbitt ?

ondrejmirtes:
If unit tests pass there's still something that can be broken in your code that isn't covered by tests. Take for example the Translator fixes - if `static::translator()` returned an instance that doesn't have the `addResource` method, the code would break. PHPStan enforces more type safety and as a result, you can even write less tests - you don't have to write tests for "boring code" that just calls other methods. For example, people usually don't write tests for getters and setters because that would be a waste of time. But PHPStan can find a mistake such as a typo in a setter, or point out a wrong typehint. Some teams that use PHPStan even run it before running their unit tests.

ondrejmirtes:
Also, PHPStan can find more errors during refactoring that you'd otherwise find during testing or even in production.

phpstan.neon
```diff
@@ -0,0 +1,7 @@
+parameters:
```
lucasmichot: Can you add this file not to be exported in `.gitattributes` too?

phpstan.neon
```diff
@@ -0,0 +1,7 @@
...
+	ignoreErrors:
+		- '#Access to an undefined property#'
```
lucasmichot: Can you sort error messages ?

src/Carbon/Carbon.php
```diff
@@ -1091,8 +1091,9 @@ public static function hasRelativeKeywords($time)
     protected static function translator()
     {
         if (static::$translator === null) {
-            static::$translator = new Translator('en');
-            static::$translator->addLoader('array', new ArrayLoader());
+            $translator = new Translator('en');
```
lucasmichot: 👍

tests/Carbon/ModifyTest.php
```diff
@@ -9,6 +9,8 @@
  * file that was distributed with this source code.
  */
 
+namespace Tests\Carbon;
```
lucasmichot: 👍

.travis.yml
```diff
@@ -28,7 +28,9 @@ install:
   - if [[ $setup = 'stable' ]]; then travis_retry composer update --prefer-dist --no-interaction --no-suggest --prefer-stable; fi
   - if [[ $setup = 'lowest' ]]; then travis_retry composer update --prefer-dist --no-interaction --no-suggest --prefer-stable --prefer-lowest; fi
 
-script: vendor/bin/phpunit --verbose --coverage-text --coverage-clover=coverage.xml
+script:
+  - vendor/bin/phpunit --verbose --coverage-text --coverage-clover=coverage.xml
+  - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.5.2 && vendor/bin/phpstan analyse -c phpstan.neon -l 3 src tests; fi
```
lucasmichot: Can you use long argument notation, config and level ?

ondrejmirtes:
I'll fix your suggestions over the weekend. I'd also like to release 0.6 soon that brings some fixes that will lower the number of ignored errors. Thanks 👍

.travis.yml
lucasmichot: Why not level 4?

ondrejmirtes: I will use the long arg notation.

Level 4 is not suitable for Carbon, it's better for PHP 7 codebases with scalar typehints. Take this example:

```
/** @var bool */
private $foo;

/**
 * @param bool $foo
 */
public function setFoo($foo)
{
    $this->foo = (bool) $fool
}
```

PHPStan will complain that you're casting `bool` to `bool`. But you can't remove the casting because something else can be passed there and to keep the integrity of the object, you need to be sure that only `bool` is assigned to `$this->foo`.

But in this case:

```
public function setFoo(bool $foo)
{
    $this->foo = (bool) $fool
}
```

The `(bool)` can be safely removed.

lucasmichot: Thanks for this explanations @ondrejmirtes , we can stick to level 3 then

lucasmichot:
Make sure to rebase and fix the conflict on `.travis.yml` @ondrejmirtes

ondrejmirtes:
Fixed, rebased, build passes :)

.travis.yml
```diff
@@ -55,12 +57,16 @@ matrix:
...
+    - php: 7.0
+      env: PHPSTAN=1
     - php: 7.1
       env: setup=lowest
     - php: 7.1
       env:
         - coverage=true
         - setup=stable
+    - php: 7.1
+      env: PHPSTAN=1
```
lucasmichot: Can you use `phpstan=true` (lowercase, pseudo boolean) instead - just to keep consistent ? ;-)

ondrejmirtes: @lucasmichot Updated :)

lucasmichot:
Thanks for the great work @ondrejmirtes !

ondrejmirtes:
@lucasmichot You're welcome 😊 I'd be glad if you considered using PHPStan on your other, maybe even commercial projects 😊 If you have any questions or suggestion for PHPStan, get in touch 😊

Issue #861, opened 2017-01-17
Why toIso8601String() change from +0000 to +00:00 ?


thanh-taro:
I just run the composer update today and all my unit tests are failed. Because of this change.

dereuromark:
Refs https://github.com/briannesbitt/Carbon/issues/451
In Chronos we were able to implement it correctly, the PHP implementation is wrong and not the actual ISO value. For BC reasons you cannot change this in in minors of course for Carbon usually.

ameliaikeda:
This is indeed a breaking change – it just KOed several test suites of ours, and our database :cry:

kylekatarnls:
Now explained in the documentation.

PR #862, opened 2017-01-17
Fix ISO8601 bug

lucasmichot:
Fix ISO8601 BC

---

FIx https://github.com/briannesbitt/Carbon/issues/861

### DIFF
--- a/tests/Carbon/StringsTest.php
+++ b/tests/Carbon/StringsTest.php
@@ -81,7 +81,7 @@ class StringsTest extends AbstractTestCase
          * $cache = setlocale(LC_TIME, 0);
          * $d = Carbon::create(2016, 01, 06, 00, 00, 00);
          * setlocale(LC_TIME, 'spanish');
-         * $this->assertSame(utf8_encode('miércoles 06 enero 2016'), $d->formatLocalized('%A %d %B %Y'));
+         * $this->assertSame(utf8_encode('miï¿½rcoles 06 enero 2016'), $d->formatLocalized('%A %d %B %Y'));
          * setlocale(LC_TIME, $cache);
          * Carbon::setUtf8(false);
          */
@@ -138,7 +138,7 @@ class StringsTest extends AbstractTestCase
     public function testToIso8601String()
     {
         $d = Carbon::create(1975, 12, 25, 14, 15, 16);
-        $this->assertSame('1975-12-25T14:15:16-05:00', $d->toIso8601String());
+        $this->assertSame('1975-12-25T14:15:16-0500', $d->toIso8601String());
     }
 
     public function testToRC822String()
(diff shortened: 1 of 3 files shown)


dereuromark:
@lucasmichot I think your changes revert the BC break, so they "could" make sense after all.

lucasmichot:
Yeah but is this really a BC - or more an expected behaviour as per your comment in #861 ? It's not that obvious
Ping @briannesbitt

briannesbitt:
Ya I mentioned awhile ago that its a breaking change but also a bug fix to implement correct behaviour.  I guess the issue is people are relying on the incorrect behaviour so thats why its a break.

I had changed it to use the atom format awhile ago but I guess it never got released.  I guess save it for 2.0 at this point to mimic PHP behaviour.

ameliaikeda:
It's probably worth noting that *technically* both formats are valid under ISO8601, but PHP's default handling of dates (`createFromFormat` being the biggest problem) doesn't allow parsing ISO8601 properly, as it's too rigid. Not everyone uses new/parse though, and interop for proper handling of iso8601 is terrible in general.

tests/Carbon/StringsTest.php
```diff
@@ -81,7 +81,7 @@ public function testToLocalizedFormattedDateStringWhenUtf8IsNedded()
          * $cache = setlocale(LC_TIME, 0);
          * $d = Carbon::create(2016, 01, 06, 00, 00, 00);
          * setlocale(LC_TIME, 'spanish');
-         * $this->assertSame(utf8_encode('mi�rcoles 06 enero 2016'), $d->formatLocalized('%A %d %B %Y'));
+         * $this->assertSame(utf8_encode('mi�rcoles 06 enero 2016'), $d->formatLocalized('%A %d %B %Y'));
```
ameliaikeda: potential encoding error in your client here, maybe?

lucasmichot:
@briannesbitt I think that even if PHP does not exactrly the ISO, it should be worth to fix this in `1.x`
Somehow it's globally accepted in the PHP community

ameliaikeda:
@lucasmichot PHP isn't actually wrong, however: ISO8601 is a very large standard. Both `Y-m-d\TH:i:sO` and `Y-m-d\TH:i:sP` are correct under ISO8601, as is `Y-m-d`, `Y-m-d H:i:s` and `Y-m`. PHP just doesn't have very flexible handling by default.

lucasmichot:
Thanks for making it clearer @ameliaikeda 
Neverheless we should maybe provide the default `\DateTime::ISO8601` format value

jaylinski:
I wrote a little blog post on this issue: https://dev.karriere.at/a/so-you-want-to-use-iso-8601-dates-in-php-and-browsers

jaylinski:
@Glavic You are right, this works in PHP and in browsers. I will update the article.

kylekatarnls:
Won't fix as we do not wish to align on a PHP documented bug. If you need the legacy ISO8601 format, just use `->format(Carbon::ISO8601)`

PR #864, opened 2017-01-18
iso8601 not atom dateString

honarkhah:


lucasmichot:
Thanks for that @honarkhah 
See https://github.com/briannesbitt/Carbon/pull/862 which is on this topic

PR #867, opened 2017-01-21
PHPDoc consistency and centralise common code

jonbaldie:
In this I have attempted to make the PHPDoc blocks more consistent, and correcting several incorrect `@return` types.

I have also centralised some common method calls into methods to improve maintainability and reduce complexity.

There are quite a lot of alterations so I would understand any reluctance to merge this in, however all the tests seem to still pass. Up to you lads.


lucasmichot:
Can you add tests for `nowWithSameTz` and `resolveCarbon` ?

jonbaldie:
Yep will do.

jonbaldie:
Looks like they work, the tests seem to have passed.

kylekatarnls:
Hi, it's very interesting. But I see multiple subject, one by one, some of them could be yet merged. But here we have to have them all checked to merge.

First, I suspect min()/max() to be slower than the comparison. Could you benchmark it to check and revert the min/max replacements if it's slower?

kylekatarnls:
I extracted your code for nowWithSameTz and resolveCarbon in #1164

Issue #883, opened 2017-02-02
modify after timezone issue


fedot01:
If I want to get midnight time for different timezones with initial time I got this problem. For example with timestamp `1485943200` I need to know midnight time for Europe/Moscow and Europe/Samara.

I do this:
```php
$initial = Carbon::createFromTimestamp(1485943200);
$moscow_midnight = $initial->copy()->timezone('Europe/Moscow')->modify('midnight')->getTimestamp();
$samara_midnight = $initial->copy()->timezone('Europe/Samara')->modify('midnight')->getTimestamp();
```

And got:
```php
$moscow_midnight = 1485907200
$samara_midnight = 1485907200
```

This is midnight time for `GMT 0` for that day not for those timezones.

Ovsyanka:
This behavior is dont expected for me too. And i founded that it was introduced while solving #88.

Ovsyanka:
In #88 described the problem, that if you modifying date by hours there 

I figured that the problem, raised in #88 relates only to units <= hour. There is examples what i mean: For example we have the date 2014-03-30 00:00:00 in Europe/London `new Carbon('2014-03-30 00:00:00, 'Europe/London')`, then if i want to increase date by 1 day, i expect '2014-03-31 00:00:00', but if want to increase date by 24 hours, then i expect '2014-03-31 01:00:00', because in this timezone there will be this time after 24 hours. The same for minutes and seconds.

So, in the end, i surely suppose, that fix implemented for #88 should be used only for seconds, minutes and hours modifications (i dont know, is there another modifications less then hour?). All other modifications should work like before.

Ovsyanka:
Issues i think related to this and, that may be fixed while fixing this:
#887, #863, #871

kaufholdr:
Can you state your systems timezone? [date_default_timezone_get](http://php.net/manual/de/function.date-default-timezone-get.php)

Ovsyanka:
I can, but why you want to know _ours_ timezones?

kaufholdr:
It's only interesting if one of the used timezone is your default one or not because the Carbon::modify method behaves differently if the calculation is done in your timezone.

kaufholdr:
I can answer the question myself. If you execute it with your local timezone it gives the correct result. So both are not your local timezone. ;-)

Ovsyanka:
[...]

Issue #871, opened 2017-01-28
Issue with addSeconds and RFC dates after upgrade to 1.22.1


jakelehner:
We use RFC3339 format in our API, and store dates in unix timestamp in our database.
After upgrading Carbon to 1.22.1, I'm seeing an bug with converting Carbon dates back to timestamp after doing any math on the object:

*Carbon 1.21*
```php
$startTime = "2017-01-28T12:45:20-07:00";
$startTimeObj = new Carbon($startTime);

$startTimeObj->toRfc3339String();  // 2017-01-28T12:45:20-07:00
$startTimeObj->timestamp;          // 1485632720
$startTimeObj->addSeconds(10);
$startTimeObj->toRfc3339String();  // 2017-01-28T12:45:30-07:00 <-- +10 seconds
$startTimeObj->timestamp;          // 1485632730 <-- +10 seconds
```

*Carbon 1.22.1*
```php
$startTime = "2017-01-28T12:45:20-07:00";
$startTimeObj = new Carbon($startTime);

$startTimeObj->toRfc3339String();  // 2017-01-28T12:45:20-07:00
$startTimeObj->timestamp;          // 1485632720
$startTimeObj->addSeconds(10);
$startTimeObj->toRfc3339String();  // 2017-01-28T12:45:30-07:00 <-- +10 seconds
$startTimeObj->timestamp;          // 1485607530 <-- +10 seconds - 7 Hours
```

It appears as if I'm losing the timezone reference after doing `addSeconds()`, but only for converting to UNIX timestamp. The RFC conversion still works as expected.

Can anyone else confirm or comment?

jackshakes:
I think this is related to, or caused by the same core issue as, #863

jackshakes:
The issue I'm experiencing seems to be coming from the new implementation of the `modify()` method. In 1.22, methods such as `addSeconds()` ended called DateTime's `modify()` method directly. In 1.21.1, The Carbon class contains the following method:

```php
    /**
     * Consider the timezone when modifying the instance.
     *
     * @param string $modify
     *
     * @return static
     */
    public function modify($modify)
    {
        if ($this->local) {
            return parent::modify($modify);
        }

        $timezone = $this->getTimezone();
        $this->setTimezone('UTC');
        $instance = parent::modify($modify);
        $this->setTimezone($timezone);

        return $instance;
    }
```

[...]

Issue #863, opened 2017-01-17
Timezone offset is added to the value after upgrading to 1.22


sisve:
I'm seeing the timezone value added into the result after`addDays(1)` and `subDays(1)`. The last value ends up at 11:23:45 while still keeping the same timezone. I expect it to end up at 01:23:45.

Output from Laravel's tinker tool:
```
>>> $input = '2017-01-01T01:23:45+10:00'
=> "2017-01-01T01:23:45+10:00"
>>> $value = \Carbon\Carbon::parse($input);
=> Carbon\Carbon {#1650
     +"date": "2017-01-01 01:23:45.000000",
     +"timezone_type": 1,
     +"timezone": "+10:00",
   }
>>> $value = $value->addDays(1)
=> Carbon\Carbon {#1650
     +"date": "2017-01-02 01:23:45.000000",
     +"timezone_type": 1,
     +"timezone": "+10:00",
   }
>>> $value = $value->subDays(1)
=> Carbon\Carbon {#1650
     +"date": "2017-01-01 11:23:45.000000",
     +"timezone_type": 1,
     +"timezone": "+10:00",
   }

```

Failing test:
```
<?php

class CarbonTest extends \PHPUnit\Framework\TestCase {

    public function testCarbonZomg() {
        $input = '2017-01-01T01:23:45+10:00';

        $value = \Carbon\Carbon::parse($input);
        $this->assertEquals('2017-01-01T01:23:45+10:00', $value->toIso8601String());

        $value = $value->addDays(1);
        $this->assertEquals('2017-01-02T01:23:45+10:00', $value->toIso8601String());

        $value = $value->subDays(1);
        $this->assertEquals('2017-01-01T01:23:45+10:00', $value->toIso8601String());
        /**
         * Failed asserting that two strings are equal.
         * --- Expected
         * +++ Actual
         * @@ @@
         * -'2017-01-01T01:23:45+10:00'
         * +'2017-01-01T11:23:45+10:00'
         */
    }

}
```

brianseitel:
I also ran into this problem, same issue. I do a lot of sensitive work with dates that, if wrong, could cost me a lot of money.

 Any chance of it getting fixed soon?

TobiasFroberg:
This can be reproduced over here as well, running 1.22.1

lucasmichot:
Can you check if this related to https://github.com/briannesbitt/Carbon/issues/861?

lucasmichot:
[...]

PR #884, opened 2017-02-03
Fixed issue #883

Ovsyanka:


lucasmichot:
Very nice @Ovsyanka - let me have a look at it later

Ovsyanka:
It also should fix #887, #863, #871

src/Carbon/Carbon.php
```diff
@@ -3326,16 +3326,20 @@ public function isBirthday(Carbon $dt = null)
      */
     public function modify($modify)
     {
-        if ($this->local) {
-            return parent::modify($modify);
-        }
+        $shouldBeModifiedInUtc = preg_match('/(sec|second|min|minute|hour)s?/', $modify);
```
lucasmichot: Is a variable really useful here?

tests/Carbon/ModifyTest.php
```diff
@@ -31,4 +31,36 @@ public function testTimezoneModify()
         $b->addHours(24);
         $this->assertSame(24, $a->diffInHours($b));
     }
+
+    /**
+     * @dataProvider getDealWithDayLightTests
+     */
```
lucasmichot: `@param` docblocks missing :-(

tests/Carbon/ModifyTest.php
```diff
@@ -31,4 +31,36 @@ public function testTimezoneModify()
...
+    public function testDealWithDayLight($dateString, $timezone, $modify, $expected)
+    {
+        $date = new Carbon($dateString, $timezone);
+        $date->modify($modify);
+        $this->assertSame($expected, $date->format('Y-m-d H:i:s P'));
+        $this->assertInstanceOf('Carbon\Carbon', $date);
```
lucasmichot: Feel free to use `assertInstanceOfCarbon`

src/Carbon/Carbon.php
```diff
@@ -3326,16 +3326,20 @@ public function isBirthday(Carbon $dt = null)
...
 
-        $timezone = $this->getTimezone();
-        $this->setTimezone('UTC');
-        $instance = parent::modify($modify);
-        $this->setTimezone($timezone);
+        if ($shouldBeModifiedInUtc) {
+            $timezone = $this->getTimezone();
```
lucasmichot: Maybe use a negated assertion and an early return here ?

Ovsyanka:
Okay! I will make done it very soon.
The first two remarks are more about code style. But i think it is not a question to discuss, so i will change in the way you suggested.
The last two very reasonable, i will fix it too.

lucasmichot:
👍

lucasmichot:
I wish all your next PRs could be as good as your first one then 👌!

You can simply add another commit to thir PR

Ovsyanka:
Actially i changed my mind a little, about early return. I think it is not good in this case. I think that it is fine to handle errors, when you just stopping the function and do not doung its main task. But if you doing useful tasks in different manner, i surely suppose, that you have to use if-else block. It is more clearly to understand.
Do you agree with me? May i let it in that manner?

lucasmichot:
Sure @Ovsyanka

Ovsyanka:
Okay, i already did this in the last commit before your comment.

kaufholdr:
First of all I like this fix.

In my opinion it would be nice to have a small extension in the documentation of modify about when DST is included and when it is ignored.

src/Carbon/Carbon.php
```diff
@@ -3326,16 +3330,21 @@ public function isBirthday(Carbon $dt = null)
      */
     public function modify($modify)
     {
-        if ($this->local) {
-            return parent::modify($modify);
-        }
+        // If we changing time in some timezone with daylight correction, there can be the case,
+        // that this correction influence to time in nonwanted way.
+        // So we have to make this modification in UTC timezone.
+        if (!preg_match('/(sec|second|min|minute|hour)s?/', $modify)) {
```
hinovana: case-insensitive search ?

Ovsyanka: Great suggestion! I will fix it.

Ovsyanka:
I rebased branch to actual master and made regexp case-insensitive.

lucasmichot:
Just restarting Travis

Ovsyanka:
@kaufholdr wrote:
> In my opinion it would be nice to have a small extension in the documentation of modify about when DST is included and when it is ignored.

I totally agree with you, but i think it should be in another issue (and pull-request) because this is not nesessary to apply this. And i dont know where to place such documentstion. I will create the issue soon.

Ovsyanka:
@lucasmichot, why you do not merging this pull-request? Should i do something else?

bensinclair:
Would love to see this merged. We are experiencing the same issue and have had to create workarounds.

oceanwap:
Is anyone going to merge this? It's pending for so long.

vlakoff:
Ouch, I just got notified of this issue too. Looks like this PR deserves a bump :)

src/Carbon/Carbon.php
```diff
@@ -3326,16 +3326,21 @@ public function isBirthday(Carbon $dt = null)
      */
     public function modify($modify)
     {
-        if ($this->local) {
-            return parent::modify($modify);
-        }
+        // If we changing time in some timezone with daylight correction, there can be the case,
+        // that this correction influence to time in nonwanted way.
+        // So we have to make this modification in UTC timezone.
+        if (!preg_match('/(sec|second|min|minute|hour)s?/i', $modify)) {
```
vlakoff: Technically, this regex is identical to `/sec|min|hour/i`

vlakoff: Also, what is the situation concerning the other modifiers ("day", "week", etc.)?

Ovsyanka: Technically - yes, but this is more explicit and obvious for human reading. I will move the long regexp into comments and replace it by short one in the code.

Others is OK. The timezone not influence to modifications with unit bigger than hour.

vlakoff: I had to dig a bit to find this documentation page: [Relative Formats](http://php.net/manual/en/datetime.formats.relative.php)

Ovsyanka: I realised that `sec` matches to "ordinal" `second` (like `first`, `second`, `third`). Hm. I am not sure about how to solve this. I have to think about that.

Ovsyanka: I think i will try something like this: `/[+-]?[0-9]+ (sec|min|hour)\w*( ago)?/i`. What do you think?

Ovsyanka: @vlakoff, Look at https://github.com/briannesbitt/Carbon/issues/883#issuecomment-339064288 plz.

lwille:
Any news?

Ovsyanka:
I am waiting for votes on my comment https://github.com/briannesbitt/Carbon/issues/883#issuecomment-339446962.

Ovsyanka:
I decide to rid off this pull-request because that is wrong approach to solve the problem. Look at the https://github.com/briannesbitt/Carbon/issues/883#issuecomment-357393640 for details.

Issue #885, opened 2017-02-03
php-cs-fixer and .gitattributes EOL inconsistent on Windows


Ovsyanka:
I have a Windows 7. When i run php-cs-fixer all files marked "changed" because theirs EOL was changed from CRLF to LF.

But why they have CRLF initially?
It is because `* text=auto` in `.gitattributes`. It means that all files that considered as text will have system-related EOL in working directory. On windows it is CRLF.

But php-cs-fixer dont bother what is the current system - it always make EOL = LF.

GawainLynch:
​Owing to the apparently "unmaintained" status of this project (last official update was Februrary 2017), a small group have forked this project, and merged this pull request.
​​
You can safely update to the new project by changing `nesbot/carbon` to `carbondate/carbon` in your project's `composer.json` file.

The new home for the project is https://github.com/CarbonDate/Carbon and we welcome continuing engagement from the open source community.

Please also note that Carbon owes its very existence to Brian Nesbitt's work, and for that, we are eternally grateful.

dereuromark:
@GawainLynch 
Why not contributing to https://github.com/cakephp/chronos as de facto successor?
Having a strong community behind a project is always a big plus here.

GawainLynch:
Compatibility would be the first. We have years of support to provide yet on components using `Carbon\Carbon`, so Chronos is just not an option.

kylekatarnls:
Now fixed.

PR #886, opened 2017-02-03
Fix issue #885

Ovsyanka:
> I have a Windows 7. When i run php-cs-fixer all files marked "changed" because theirs EOL was changed from CRLF to LF.

> But why they have CRLF initially?
> It is because `* text=auto` in `.gitattributes`. It means that all files that considered as text will have system-related EOL in working directory. On windows it is CRLF.

> But php-cs-fixer dont bother what is the current system - it always make EOL = LF.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-* text=auto
+* text=auto eol=lf
 
 .codecov.yml export-ignore
 .editorconfig export-ignore


Ovsyanka:
Why you don't merging this pull-request? This issue makes autoformatting a bit complicated for me. Do you have any objections?

PR #889, opened 2017-02-06
[docs] No approximation is used anymore.

jeroennoten:
I removed a statement in the docs about approximation in `diffForHumans()` that does not hold anymore.


lucasmichot:
Ver good catch @jeroennoten

PR #898, opened 2017-02-10
Provide RFC7231 formating

### LINKED ISSUE #897: Provide a RFC7231 constant to format dates for HTTP headers
Missing from PHP is actually a RFC7231 format for styling dates for headers.  The format is fairly simple and only a subset of the existing standard formats but it can be important in some systems when providing things like cache headers. Since carbon is an improved version of php's DateTime object it seems like a good place to provide this extended functionality.

The format is simply: `'D, d M Y H:i:s \G\M\T'`

PHP.net comment about this:
http://php.net/gmdate#25031
Drupal discussion and patch about fixing cache headers and providing similar functinoality:
https://www.drupal.org/node/1918820
RFC documentation. Original and most recent version.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3
http://tools.ietf.org/html/rfc7231#section-7.1.1.1
PHP.net bug:
https://bugs.php.net/bug.php?id=74080

neclimdul:
Fixes #897.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -112,6 +112,13 @@ class Carbon extends DateTime
     const SECONDS_PER_MINUTE = 60;
 
     /**
+     * RFC7231 DateTime format.
+     *
+     * @var string
+     */
+    const RFC7231_FORMAT = 'D, d M Y H:i:s \G\M\T';
+
+    /**
      * Default format to use for __toString method when type juggling occurs.
      *
      * @var string
@@ -1397,6 +1404,18 @@ class Carbon extends DateTime
     public function toW3cString()
     {
         return $this->format(static::W3C);
+    }
+
+    /**
+     * Format the instance as RFC7231
+     *
+     * @return string
+     */
+    public function toRfc7231String()
+    {
+        return $this->copy()
+            ->setTimezone('GMT')
+            ->format(static::RFC7231_FORMAT);
     }
 
     ///////////////////////////////////////////////////////////////////
--- a/tests/Carbon/StringsTest.php
+++ b/tests/Carbon/StringsTest.php
@@ -189,4 +189,13 @@ class StringsTest extends AbstractTestCase
         $d = Carbon::create(1975, 12, 25, 14, 15, 16);
         $this->assertSame('1975-12-25T14:15:16-05:00', $d->toW3cString());
     }
+
+    public function testToRfc7231String()
+    {
+        $d = Carbon::create(1975, 12, 25, 14, 15, 16, 'GMT');
+        $this->assertSame('Thu, 25 Dec 1975 14:15:16 GMT', $d->toRfc7231String());
+
+        $d = Carbon::create(1975, 12, 25, 14, 15, 16);
+        $this->assertSame('Thu, 25 Dec 1975 19:15:16 GMT', $d->toRfc7231String());
+    }
 }


lucasmichot:
Thanks for this PR @neclimdul , can you also add tests?

neclimdul:
Of course. Also I was handing the timezones wrong... so that's why we have tests right? :)

src/Carbon/Carbon.php
```diff
@@ -112,6 +112,11 @@ class Carbon extends DateTime
     const SECONDS_PER_MINUTE = 60;
 
     /**
+     * RFC7231 DateTime format.
```
lucasmichot: `@var string` maybe?

src/Carbon/Carbon.php
```diff
@@ -112,6 +112,11 @@ class Carbon extends DateTime
...
+     */
+    const RFC7231 = 'D, d M Y H:i:s \G\M\T';
```
lucasmichot: Could be probably renamed `RFC7231_FORMAT` for consistancy

src/Carbon/Carbon.php
```diff
@@ -1392,6 +1397,25 @@ public function toW3cString()
...
+        if (!$this->utc) {
+            $tz = $this->getTimezone();
+            $this->setTimezone('GMT');
+        }
+        $formatted = $this->format(static::RFC7231);
+        if (isset($tz)) {
+            $this->setTimezone($tz);
+        }
+
+        return $formatted;
```
lucasmichot: Why not just `return $this->copy()->setTimezone('GMT')->format(static::RFC7231_FORMAT);` ?

neclimdul: nice, i was actually looking for something like copy i was just using the wrong term. clone, etc.

src/Carbon/Carbon.php
neclimdul: That'd actually be inconsistent with the existing date constants in PHP. Your call on what to be consistent with there.

src/Carbon/Carbon.php
neclimdul: definitely.

src/Carbon/Carbon.php
lucasmichot: ;-)

src/Carbon/Carbon.php
lucasmichot: This is the case for all built-in `Datetime` formats.
The fomats which are related to `Carbon` only are suffixed with `_FORMAT`: see `DEFAULT_TO_STRING_FORMAT`

lucasmichot: Also you can simplify your format to `D, d M Y H:i:s T`; your tests will still pass.

tests/Carbon/StringsTest.php
```diff
@@ -188,4 +188,12 @@ public function testToW3cString()
         $d = Carbon::create(1975, 12, 25, 14, 15, 16);
         $this->assertSame('1975-12-25T14:15:16-05:00', $d->toW3cString());
     }
+
+    public function testToRfc7231String()
+    {
+        $d = Carbon::create(1975, 12, 25, 14, 15, 16, 'GMT');
+        $this->assertSame('Thu, 25 Dec 1975 14:15:16 GMT', $d->toRfc7231String());
+        $d = Carbon::create(1975, 12, 25, 14, 15, 16);
```
lucasmichot: Maybe a line break between the 2 tests?

src/Carbon/Carbon.php
lucasmichot: Yes, but as you've already set the `GMT` timezone in `$this->copy()->setTimezone('GMT')->format(static::RFC7231_FORMAT);` ;-)

Ovsyanka: IMHO @neclimdul's variant is more appropriate in this case. Because when you just look at the const value - you see there "timezone identifyer" whilst RFC says us that there should be "exactly GMT".

neclimdul: For what its worth PHP merged in the same implementation today. https://github.com/php/php-src/commit/618d97962a8ac59d981c8307cc704790f46090c8

Glavic:
@neclimdul: can you rebase?

neclimdul:
Should I squash while I'm at it?

kylekatarnls:
Yep :) Thanks.

neclimdul:
Done. Thanks coming back around to this :-D

PR #899, opened 2017-02-12
Add setter methods for default formats

martin-heralecky:
Several additional setter methods which can be used to setup the default formats.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -119,11 +119,81 @@ class Carbon extends DateTime
     const DEFAULT_TO_STRING_FORMAT = 'Y-m-d H:i:s';
 
     /**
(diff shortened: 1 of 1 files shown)


lucasmichot:
Thanks for this PR @martin-heralecky 
I feel that this can be totally make sens while extending Carbon, but not sure it makes sense to include this in the core

martin-heralecky:
Well, why not? Personally, I'd welcome this functionality because I can change the default formats of dates and times which saves me a lot of time in comparison that I'd do the same manually (using ->format() every time I need to display a date).

Ovsyanka:
I agree with Lucas. Btw i think that existed methods you changed in your PR don't needed in the core class at all =)
You can extend Carbon class and redefine this methods in a way you wish. And you more likely don't need all that static variables.
It is just my personal opinion.

PR #907, opened 2017-03-05
Adding support for not overflowing months when adding a year

robertpcontreras:
This adds the ability to set carbon to not overflow months when adding a year. For example, see the following:
```
Carbon\Carbon::parse('2016-02-29')->addYear()->toDateString(); // 2017-03-01
```
However, you can set the option to tell carbon to not overflow months like:
```
Carbon\Carbon::useMonthsOverflow(false);
```
This means:
```
Carbon\Carbon::parse('2016-01-31')->addMonth()->toDateString(); // 2016-02-28
```
This PR adds similar functionality so that:
```
Carbon\Carbon::parse('2016-02-29')->addYear()->toDateString(); // 2017-02-28
```

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2080,7 +2080,7 @@ class Carbon extends DateTime
      */
     public function subYearsNoOverflow($value)
     {
-        return $this->addMonthsNoOverflow(-1 * $value);
+        return $this->subMonthsNoOverflow($value * static::MONTHS_PER_YEAR);
     }
 
     /**
@@ -2104,7 +2104,7 @@ class Carbon extends DateTime
      */
     public function subYearsWithOverflow($value)
     {
-        return $this->addYears(-1 * $value);
+        return $this->subMonthsWithOverflow($value * static::MONTHS_PER_YEAR);
     }
 
     /**
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1968,11 +1968,7 @@ class Carbon extends DateTime
      */
     public function addYears($value)
     {
-        if (static::shouldOverflowMonths()) {
-            return $this->modify((int) $value.' year');
-        }
-
-        return $this->addMonthsNoOverflow($value * 12);
+        return $this->modify((int) $value.' year');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -1968,7 +1968,11 @@ public function isSaturday()
      */
     public function addYears($value)
     {
-        return $this->modify((int) $value.' year');
+        if (static::shouldOverflowMonths()) {
+            return $this->modify((int) $value.' year');
+        }
+
+        return $this->addMonthsNoOverflow($value * 12);
```
lucasmichot: Can you  use the constant instead of `12` ?

lucasmichot:
Thanks a lot @robertpcontreras 
I feel this is a very good idea but this would be a BC as some people expects also Carbon to behave like Datetime.

Why not going a bit further with this PR ?

I believe it would make sense to add the following functions instead:

- `addYearNoOverflow`
- `addYearsNoOverflow`
- `subYearNoOverflow`
- `subYearsNoOverflow`
- `addYearWithOverflow`
- `addYearsWithOverflow`
- `subYearWithOverflow`
- `subYearsWithOverflow`

robertpcontreras:
hey @lucasmichot,

made the changes you requested.

thanks

lucasmichot:
Tests?

tests/Carbon/AddTest.php
```diff
@@ -265,4 +265,68 @@ public function testSubCenturiesNegative()
         $this->assertSame(2075, Carbon::createFromDate(1975)->subCenturies(-1)->year);
         $this->assertSame(2175, Carbon::createFromDate(1975)->subCenturies(-2)->year);
     }
+
+    public function testAddYearNoOverflow()
+    {
+        $date = Carbon::createFromDate(2016, 2, 29)->addYearNoOverflow();
+        $this->assertSame(28, $date->day);
```
lucasmichot: I think you can use `assertCarbon` in all your test cases

src/Carbon/Carbon.php
```diff
@@ -1984,6 +1984,58 @@ public function addYear($value = 1)
     }
 
     /**
+     * Add years to the instance with no overflow of months
```
lucasmichot: Can you base yourself on `\Carbon\Carbon::addMonths` so that you make use of `\Carbon\Carbon::shouldOverflowMonths` too ?

robertpcontreras: Like this? 
```
public function addYearsNoOverflow($value)
{
    return $this->addMonths($value * static::MONTHS_PER_YEAR);
}
```
is it not already implied that the user requires the overflow if they call this method? It wouldn't make sense for this method to be called but the months to not overflow?

lucasmichot: Hmm, true. My bad

src/Carbon/Carbon.php
```diff
@@ -1984,6 +1984,58 @@ public function addYear($value = 1)
...
+     */
+    public function addYearsNoOverflow($value)
+    {
+        return $this->addMonthsNoOverflow($value * static::MONTHS_PER_YEAR);
+    }
+
+    /**
+     * Add year with overflow months set to false
+     *
+     * @param $value
```
lucasmichot: `@param int $value`

src/Carbon/Carbon.php
```diff
@@ -1984,6 +1984,58 @@ public function addYear($value = 1)
...
+     */
+    public function addYearsWithOverflow($value)
+    {
+        return $this->addYears($value);
+    }
+
+    /**
+     * Add year with overflow.
+     *
+     * @param $value
```
lucasmichot: `@param int $value`

tests/Carbon/AddTest.php
robertpcontreras: all tests refactored, thanks.

robertpcontreras:
@lucasmichot is this all good now?

robindirksen1:
Any ideas when this PR is merged?

kylekatarnls:
Note: this needs to add an useYearOverflow method for consistency with month. To do before the 1.24 release.

PR #912, opened 2017-03-10
Corrects improper handling of daylight savings time

jstiltner:
Corrects this error, which revealed a mishandling of daylight savings time:
```>>> use Carbon\Carbon
=> null
>>> $time = Carbon::now();
=> Carbon\Carbon {#1201
     +"date": "2017-03-10 13:34:52.000000",
     +"timezone_type": 3,
     +"timezone": "America/New_York",
   }
>>> $time->addHours(47);
=> Carbon\Carbon {#1201
     +"date": "2017-03-12 12:34:52.000000",
     +"timezone_type": 3,
     +"timezone": "America/New_York",
   }```

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -3323,7 +3323,6 @@ class Carbon extends DateTime
      */
     public function modify($modify)
     {
-
         $timezone = $this->getTimezone();
         $this->setTimezone('UTC');
         $instance = parent::modify($modify);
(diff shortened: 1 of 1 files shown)


Ovsyanka:
There already exists the pull-request fixing this: #884.

kylekatarnls:
Hi, @Ovsyanka will not "fix"this. This PR would apply the modify override on local and not local dates, yours  will remove it. But I we will not be able to do both in the same method, @jstiltner I invite you to argue on #884 to help us to take the right decision about this.

Else, we should not remove the 'local' getter even if modify() do not use it any longer since as it's public, it would be a breaking change.

kylekatarnls:
Hi modify is now removed.

PR #914, opened 2017-03-15
Change createFromFormat for compatibilty with php DateTime::createFromFormat()

jimrobinson-croud:
Carbon::createFromFormat() has been changed to return false for compatibility with php DateTime::createFromFormat() method

http://php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-returnvalues

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -561,9 +561,7 @@ class Carbon extends DateTime
      * @param string                    $time
      * @param \DateTimeZone|string|null $tz
      *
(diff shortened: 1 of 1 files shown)


kylekatarnls:
It would be fair but many methods use createFromFormat and it means all try-catch will be broken. We could consider this for 2.x but not for 1.x version for backward compatibility.

kylekatarnls:
First I merge this separately to fix unit tests and rebase it.

PR #927, opened 2017-03-31
Define User timezone

thiagocordeiro:
this feature allow developers to define user timezone, when this is set, Carbon will apply given tz to the date before parsing into strings

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -3399,6 +3399,7 @@ class Carbon extends DateTime
         if (!is_null(static::$userTimeZone)) {
             $this->timezone(static::$userTimeZone);
         }
+
         return parent::format($format);
     }
 }


Glavic:
This is really bad solution. You can see that in failing test. On other hand, people do not expect for `format()` method to change object timezone. Timezone should be changed on cloned object, not on calling one. Based on that, I don't see this being implemented, since it is too user specific. If you have any new feature, please don't hesitate to comment.

kylekatarnls:
Thanks, your code will be added as macro example:
https://github.com/briannesbitt/Carbon/compare/gh-pages...kylekatarnls:gh-pages-1.26

efraindx:
Thank you very much @thiagocordeiro for your approach. I've implemented the function name in my local project of this way: setDestinationTimezone(), because I needed both things, convert times from application to users and from users to application. IMO this way is better because of that.

kylekatarnls:
I cannot be sure as I don't see your code, but I don't think your approach is right. You should not convert from user timezone to application timezone. The browser can send UTC dates (`toISOString`), and that's the only reliable data to send. And then, the server always receive UTC dates no matter from where it's sent. So you can reconvert it easily to anything with no risk of mistake (DST, timezone change, timezone bad detection). But it should not be needed at all as you can simply set your server timezone as UTC. And you should.

Here are some articles to explain deeply how to handle properly timezone front-side/back-side:
https://medium.com/@kylekatarnls/always-use-utc-dates-and-times-8a8200ca3164
https://medium.com/@kylekatarnls/handle-dates-the-right-way-in-php-c215650fc4fa

PR #932, opened 2017-04-17
Add compare method to CarbonInterval

Horat1us:
Adds simple compare method to compare current interval with another (because php does now have build-in functions for it, like for DateTime)


Horat1us:
Any reason not to merge pr?

kylekatarnls:
Nop, I find it good, I just wonder if it's consistent with the Carbon methods (lt, gt, lte, gte and eq), maybe we should have some shorter/longer methods, or maybe having a compare too in Carbon.

kylekatarnls:
We'll first merge your PR as is in the 1.24 release. Then we could go further with other PRs eventually.

PR #946, opened 2017-05-10
Seconds per hour, day, week

fr05t1k:


anze:
Your code will not work on days when DST switches, for example: 26 march with locale "uk". Also tests will fails only on these days.
`Carbon::setLocale('uk');
        $this->assertEquals(
            Carbon::createFromFormat('d/m/Y H:i:s', '26/03/2017 00:00:00')->addDay(),
            Carbon::createFromTimestamp(1490479200 + Carbon::SECONDS_PER_DAY)
        );`

fr05t1k:
@anze you're right. Thanks!

kylekatarnls:
Hi, if we follow this logic, we should provide every possible factors: MONTHS_PER_CENTURY, HOURS_BY_DECADE, MINUTES_PER_WEEK, etc. and if you want to set a Jupiter time object using SECONDS_PER_MINUTE, you will also have to change all second-related constants.

That's why in my opinion, those constants should be used as settings giving a factor for each unit relative to the unit under it, and all other values should not be hard-coded but calculated from them.

You could do some extend/macro method:
```php
class MyCarbon extends Carbon
{
  public function secondsPerHour()
  {
    return static::SECONDS_PER_MINUTE * static::MINUTES_PER_HOUR;
  }
}
```

PR #950, opened 2017-05-18
Add __set_state() method, fixes issue with Laravel's config cache

rdarcy1:
Laravel uses __set_state() to set the value of config items when retrieving them from its config cache. Without the method being implemented in Carbon, the DateTime::__set_state() is used and so any Carbon instances defined in the config are converted to DateTime if the cache is used.


kylekatarnls:
Thanks, it's not Laravel specific, var_export on Carbon object should produce a code that re-create a Carbon instance. Good fix!

PR #961, opened 2017-06-03
fix: succeed init a instance from DateInterval

awonwon:
add a not operator in condition

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -226,7 +226,7 @@ class CarbonInterval extends DateInterval
      */
     public static function instance(DateInterval $di)
     {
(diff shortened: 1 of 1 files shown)


awonwon:
Another found that CarbonInterval's days property isn't correct when instance method returning. The result as following.

image

I think line 235 isn't work in CarbonInterval Class.

https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/CarbonInterval.php#L235

kylekatarnls:
Hi, this is probably not the solution. Can you provide the code you use to generate the DateInterval?

kylekatarnls:
I close this, if you can provide more details I'll reopen.

PR #963, opened 2017-06-13
Fix traditional Chinese language, extra space removed and '周' shall be '週'

luoenzhen:
There should be no space between :count and the Chinese character.
eg. the right one is "最近2月" not "最近2 月".


ymhuang0808:
Helpful!

PR #970, opened 2017-06-21
Zulu ISO8601 String Helper

bretterer:
Many times, I prefer to store my times in ISO8601 Zulu format

>Y-m-d\TH:i:s\Z

Right now, when doing `->toIso8601String()` it returns it in the current timezone.  

This PR adds a `->toIso8601ZuluString()` that will convert the current time instance to UTC and format it in the proper ISO8601 Zulu Formatted String.

A test has been written to check functionality.

reference for Zulu: https://tools.ietf.org/html/rfc3339#section-2

### DIFF
--- a/tests/Carbon/StringsTest.php
+++ b/tests/Carbon/StringsTest.php
@@ -142,6 +142,12 @@ class StringsTest extends AbstractTestCase
         $this->assertSame('1975-12-25T14:15:16-05:00', $d->toIso8601String());
     }
 
+    public function testToIso8601ZuluString()
+    {
+        $d = Carbon::create(1975, 12, 25, 14, 15, 16);
+        $this->assertSame('1975-12-25T19:15:16Z', $d->toIso8601ZuluString());
+    }
+
     public function testToRC822String()
     {
         $d = Carbon::create(1975, 12, 25, 14, 15, 16);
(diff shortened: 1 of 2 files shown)


Glavic:
I think we should not change current object (converting to different TZ).
Maybe something like this would be better?

```php
public function toIso8601ZuluString()
{
	$cloned = clone $this;
	return $cloned->setTimezone('UTC')->format('Y-m-d\TH:i:s\Z');
}
```

bretterer:
Good Point @Glavic .  I updated this to copy first.  thanks for your suggestion.

Glavic:
MHO: name with `***ISO8601***` is to extensive. Would rather see `toRfc3339ZuluString()`.

tests/Carbon/StringsTest.php
```diff
@@ -142,6 +142,12 @@ public function testToIso8601String()
         $this->assertSame('1975-12-25T14:15:16-05:00', $d->toIso8601String());
     }
 
+    public function testToIso8601ZuluString()
+    {
+        $d = Carbon::create(1975, 12, 25, 14, 15, 16);
```
Glavic: I would also add last (timezone) parameter. It will make test more robust.

PR #977, opened 2017-07-10
Update Carbon.php

autumndev:
Fix ISO8601 string format


Glavic:
Same issue as here #862

PR #978, opened 2017-07-24
Create SkTest.php

MarekAdamicky:


kylekatarnls:
Thanks for this, we used a more complete test instead. But if you see any error in it:
https://github.com/briannesbitt/Carbon/blob/master/tests/Localization/SkTest.php

PR are welcome

PR #993, opened 2017-08-14
Added isDayOfWeek(int) method

wilcarjose:
This is a simple function to verify if a date is a specific day of the week, this way it is easier to use comparisons like` if ($dt->isDayOfWeek(5))` instead of: `if ($dt->dayOfWeek === 5)`


Glavic:
@kylekatarnls: 1.24 ?

kylekatarnls:
Yep :) I approve.

wilcarjose:
Great!

PR #994, opened 2017-08-14
fix typo in from_now property and 'week' word

atefBB:
In arabic we say ':time from now' (see: https://github.com/jenssegers/date/commit/efa4cbee3d7a71509ef49e8090677fb8a948b476), and 'week' begins with 'أ' not 'إ'.


Glavic:
@atefBB: thx for the typo fix.

PR #996, opened 2017-08-14
Add Armenian Language Localization

armen-gabrielyan:


kylekatarnls:
Hi, thanks for your help. Please note that we have to change arm into hy to use the 2-letters ISO 639-1 code.

PR #1001, opened 2017-08-30
Fix comparison of dates in php >=7.1

### LINKED ISSUE #785: Microseconds PHP 7.1
Hi

I'm currently testing my application with PHP 7.1 RC5 on my local dev machine (MacOS Sierra) and Travis CI.
My tests are failing because I'm comparing dates.

After debugging I see following:
Creating a new Carbon instance with Carbon::now() on PHP < 7.1 sets the microseconds to 000000. On PHP 7.1 it sets the real microseconds. Therefore all `->eq()` are failing.

Is that a change Carbon should handle?

Regards

akovalyov:
Switch from comparison of instances to comparison of epoch seconds.
Closes #785

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1412,7 +1412,7 @@ class Carbon extends DateTime
      */
     public function eq(self $dt)
     {
-        return $this == $dt;
+        return $this->format('U') === $dt->format('U');
     }
 
     /**
--- a/tests/Carbon/ComparisonTest.php
+++ b/tests/Carbon/ComparisonTest.php
@@ -245,4 +245,20 @@ class ComparisonTest extends AbstractTestCase
         $farthest = $instance->farthest($dt1, $dt2);
         $this->assertSame($dt2, $farthest);
     }
+
+    public function testEqualDatesCreatedViaInstance()
+    {
+        $dt1 = Carbon::instance(new \DateTime());
+        usleep(5);
+        $dt2 = Carbon::instance(new \DateTime());
+        $this->assertTrue($dt1->eq($dt2));
+    }
+
+    public function testEqualDatesCreatedViaNow()
+    {
+        $dt1 = Carbon::now();
+        usleep(5);
+        $dt2 = Carbon::now();
+        $this->assertTrue($dt1->eq($dt2));
+    }
 }


akovalyov:
@lucasmichot @briannesbitt PHP7.2 is near and PHP 7.0 active support ends in 2 months. I would say, it is quite important to merge this one, if you don't have any objections of course.

GawainLynch:
​Owing to the apparently "unmaintained" status of this project (last official update was Februrary 2017), a small group have forked this project, and merged this pull request.
​​
You can safely update to the new project by changing `nesbot/carbon` to `carbondate/carbon` in your project's `composer.json` file.

The new home for the project is https://github.com/CarbonDate/Carbon and we welcome continuing engagement from the open source community.

Please also note that Carbon owes its very existence to Brian Nesbitt's work, and for that, we are eternally grateful.

Glavic:
@akovalyov: hi, could you rebase this PR? TY

akovalyov:
@Glavic done.

kylekatarnls:
Hi, I would not merge this one because the unit tests you propose already pass without using `->format('U')`

Could you please provide a unit test showing the bug you want to fix (a unit tests where `==` cause a problem that `->format('U')` solve).

kylekatarnls:
Hi, I can confirm, this tests fail on PHP 7.1 with `$this == $dt` and not before.

But it's because PHP 7.1 introduced microseconds. So I wonder if it would be right to say that Carbon instance created in the same second are equal. This would be more the responsibility of a `->isSameSecond` method or something like that.

Moreover, your test has 1 chance over 200000 to fail. It's few but it means neither the test nor the ->format('U') comparison is 100% reliable.

It redefines what equality between date is, however I think we should rely on PHP for that. Dealing with PHP versions changes is a user concern.

@Glavic @briannesbitt other opinions?

Glavic:
I agree, versions bellow <7.1 do not have micro seconds when creating DateTime object (only version 7.1.3). This can be nicely seen here: https://3v4l.org/ThbMJ

@akovalyov: what you are trying to do is just ignore micro seconds part, which I don't see correct, because value `2000-01-01 11:22:33.xxx` cannot be same as `2000-01-01 11:22:33.yyy`.

Like Kyle already said, I would not change this method. Maybe add other method for comparing without micro seconds?

kylekatarnls:
In fact, this method exists, it's just:
```php
$date1->isSameAs('U', $date2)
```
Please use this to compare dates ignoring microseconds (compatible way for different PHP versions).

But you raise a good point, we should talk about the PHP 7.1 difference in `->eq` method and documenting it accordingly.

I close this PR and open a documentation issue.

briannesbitt:
Having DateTime do the heavy lifting underneath was always the goal !

akovalyov:
@kylekatarnls that's probably not even a documentation issue, but the definition issue -- what does the equality of dates mean? Which dates are considered equal? One might even argue that dates cannot be equal, but the same only, in this case the whole `eq` method is superfluous. I.e. the current implementation of the `eq` method cannot handle the cases like https://3v4l.org/fWPs7.

kylekatarnls:
PHP evolves and keeping everything working in PHP 7.2 like it was in PHP 5.3 is not the goal of the library. Equality between any values is already defined by PHP itself `==` is equality, `===` is strict equality. PHP added microseconds for good reasons and we will follow this direction, not moving back. If something would be consider by most people as a PHP bug, it's not our responsibility to workaround with a hack, if so, we should just open a bug ticket to PHP but for me `date != date + 5µs` is an expected behavior.

What I propose to document is warn users about the new sensibility (second sensibility became microsecond sensibility from PHP 7.1).

Is everyone OK with that?

PR #1005, opened 2017-09-07
Fix issue #883, add/sub seconds/minutes/hours should handle DST correctly

knutdx:
I know there already is a pull request on this (https://github.com/Ovsyanka/Carbon/commit/eb6581b3fdddd9e5110539f5f49726127ccb4ebe) but I think my solution may be cleaner. Seems to me that `DateTime::modify` is really bad when it comes to DST, so it is better to do these `addXxx` and `subXxxx` operations with timestamps instead. Since timestamps doesn't have timezone and doesn't care about DST.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2426,6 +2426,7 @@ class Carbon extends DateTime
     public function addHours($value)
     {
         $seconds = $value * self::SECONDS_PER_MINUTE * self::MINUTES_PER_HOUR;
+
         return $this->setTimestamp($this->getTimestamp() + $seconds);
     }
 
@@ -2476,6 +2477,7 @@ class Carbon extends DateTime
     public function addMinutes($value)
     {
         $seconds = $value * self::SECONDS_PER_MINUTE;
+
         return $this->setTimestamp($this->getTimestamp() + $seconds);
     }
 
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2425,7 +2425,8 @@ class Carbon extends DateTime
      */
     public function addHours($value)
     {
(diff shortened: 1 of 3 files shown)


KnutHelland:
The failing tests now are not because of code, but because of misconfiguration of the CI

Glavic:
@knutdx: can you please rebase? TY

Ovsyanka:
I don't think it is good approach because it will be inconsistent with the DateTime default behaviour. And you just came to decision we already tagged as wrong. Look at https://github.com/briannesbitt/Carbon/issues/883#issuecomment-357266063

KnutHelland:
Fyi: when I submitted this pull request, I altso added a bug report and a merge request to phps timelib to fix this issue: https://bugs.php.net/bug.php?id=75167 there is a really ugly (and easy to fix) bug in timelib here, and it is annoying that noone has fixed it. Almost feels like the datetime library in php is abondoned.

KnutHelland:
Pull request to the timelib used in php source: https://github.com/derickr/timelib/pull/24

KnutHelland:
So at least I think carbon could do a hack to compensate for that bug

KnutHelland:
Sorry for all the comments, but I also have to say that I don't longer work with PHP and don't have a environment to test these things any more (and don't have access to that old work-account on github neiter), so I'll not rebase. But seems like you don't need it after all...

kylekatarnls:
modify finally removed, we will kept the PHP native behavior and adopted it in the diffInSeconds/Minutes/Hours method all using intervals now.

So this could be done in other methods? For now, I have no better ideas than the following names:
```
addRealSeconds
addRealMinutes
addRealHours
diffInRealSeconds
diffInRealMinutes
diffInRealHours
```

So I will open a new issue for this and close this one.

PR #1008, opened 2017-09-24
previous with timezone fix

BostjanOb:
When you create a new date with specific timezone and then use `previous()` function it returns the wrong date. 

Switching the `startOfDay()` call solves the problem.

For example `$d = Carbon::create(2017, 9, 23, 19, 00, 00, 'Europe/Berlin')->previous(Carbon::FRIDAY);` should return 2017-09-22 but it returns 2017-09-15. This PR fixes that.


BostjanOb:
Found out that it checks local timezone, if not the same it uses UTC

PR #1017, opened 2017-10-17
Optimize the magic properties getter

vlakoff:
I think the optimization is worth it, as this method may be called thousands of times.

* Use the language construct `isset()` instead of the function `array_key_exists()`
* Use a static variable

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -673,22 +673,24 @@ class Carbon extends DateTime
      */
     public function __get($name)
     {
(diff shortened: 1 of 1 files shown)


Glavic:
@vlakoff: nice catch. Can you please rebase? TY

vlakoff:
Yep, it has been rebased. Thanks @Glavic and @kylekatarnls for coming to the project :)

Glavic:
TY for your PR!

PR #1021, opened 2017-10-25
Use namespaced TestCase

MarioBlazek:


Glavic:
@MarioBlazek: can you please rebase? TY

MarioBlazek:
@Glavic rebased.

Glavic:
Should we add latest 4.8.x version to composer? `"phpunit/phpunit": "^4.8 || ^5.7"`

kylekatarnls:
Yes we should it enable the namespaced classe. And this class "prepare" the v6.

kylekatarnls:
Then, phpunit is a dev tool, allowing ^4.8 is only there to run tests on 5.3, but we can use PHPUnit version as height as our PHP compatibility range allow us. It has no impact on the final package.

kylekatarnls:
Thanks. Sorry for the delay.

PR #1029, opened 2017-11-07
Fix typo in fromSerialized() method comments

ironpilot:
Change the word `form` to `from` in method comments


Glavic:
@ironpilot: TY for the typo fix.

PR #1032, opened 2017-11-15
Fix for formatLocalized

jeroen-dof:
When you change the default format for the __toString method using setToStringFormat, the formatLocalized will use this format as input for the strtotime function. When you are using a format which is not supported by strtotime (example: d,m,Y), strtotime will return false, resulting in a Carbon object on 1970-01-01.


kylekatarnls:
Seems totally fair. Could you please rebase your PR?

briannesbitt:
Fixed in #1107 using `toDateTimeString()` instead as the default string format could be anything.

Issue #976, opened 2017-07-10
ISO8601 Date retusn Atom Date format


autumndev:
Hey,

I have upgraded our carbob instance and found some of our tests failing. They are failing as we are expecting a carbon ISO8601 string as per [docs](http://carbon.nesbot.com/docs/#api-commonformats):
```php
$dt->toIso8601String();   // 1975-12-25T14:15:16-0500
```

 but we are getting an ATOM string:
```php
$dt->toAtomString();      // 1975-12-25T14:15:16-05:00
```

looking through the code I can see the following:
```php
       /**
     * Format the instance as ATOM
     *
     * @return string
     */
    public function toAtomString()
    {
        return $this->format(static::ATOM);
    }
...
    /**
     * Format the instance as ISO8601
     *
     * @return string
     */
    public function toIso8601String()
    {
        return $this->toAtomString();
    }
```
My Question is why is the ISO8601 string just calling the Atom function? the extra ':' (colon) is breaking my tests and whilst a quick fix (use base format() function) I wish to understand why this has changed

autumndev:
I just saw https://github.com/cakephp/chronos/pull/45 and https://github.com/briannesbitt/Carbon/issues/861

which explains the php constant is wrong. 

I would suggest docs update with explanation as docs show ISO8601 with php constant formatting, which is not true.

I had put a pull request in for the fix: https://github.com/briannesbitt/Carbon/pull/977

this can be excluded if required.

autumndev:
also from what i can see [wiki](https://en.wikipedia.org/wiki/ISO_8601)

the iso format can be +1100 or +11:00

autumndev:
ignore previous comment, w3 states it should be +1100:
https://www.w3.org/TR/NOTE-datetime

acorncom:
@autumndev I must be reading something different here ...

>    Complete date plus hours and minutes:
          YYYY-MM-DDThh:mmTZD (eg 1997-07-16T19:20+01:00)

>   Complete date plus hours, minutes and seconds:
      YYYY-MM-DDThh:mm:ssTZD (eg 1997-07-16T19:20:30+01:00)

> Complete date plus hours, minutes, seconds and a decimal fraction of a
second
      YYYY-MM-DDThh:mm:ss.sTZD (eg 1997-07-16T19:20:30.45+01:00)

Looks like it should be `+01:00`?

autumndev:
[...]

PR #1042, opened 2017-11-29
ISO8601 String format display update

autumndev:
update documentation to show that toIso8601String() returns an toAtomString() and is basically a pass through function.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -426,7 +426,7 @@ echo $dt;                                          // 1975-12-25 14:15:16
 // $dt->toAtomString() is the same as $dt->format(DateTime::ATOM);
 echo $dt->toAtomString();      // 1975-12-25T14:15:16-05:00
 echo $dt->toCookieString();    // Thursday, 25-Dec-1975 14:15:16 EST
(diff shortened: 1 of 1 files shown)


autumndev:
as per issue #976

kylekatarnls:
I am OK to document these more clearly.

Our documentation is auto-generated (clone locally the project, checkout the gh-pages branch, install dependencies with `composer update`, then the script `php generate.php` allow you to generate .html files), so if you manually change docs/index.html, it will be erased by the next run of generate.php, you need to edit manually index.src.html and run generate.php before to commit the whole thing.

Now the ISO8601 result is right. But if you're still OK, it would be interesting to have a note about ->toIso8601String and the PHP DateTime format mistake (with link to the PHP doc). This note could be just after code examples at the end of the Common Formats section.

Please tell me if you have enough time to update this PR according to that.

Thanks.

neclimdul:
nit: typo in commit message and PR title.

kylekatarnls:
Thanks. Commit messages (and even PR title) are not a big problem, we squash most of PR and rewrite them at this moment.

kylekatarnls:
Closed in favor of #1121

PR #1045, opened 2017-12-02
update uk.php

AucT:
Word "назад" is not so popular nowadays. Most people use "тому" after counted time


Glavic:
Tnx for the fix.

PR #1052, opened 2017-12-14
Create a Carbon instance from the easter date of a year

andreisena:
Get the easter date based on the **Computus** calculation. See https://en.wikipedia.org/wiki/Computus

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -628,6 +628,37 @@ class Carbon extends DateTime
     }
 
     /**
+     * Create a Carbon instance from the easter day of a year.
+     *
+     * @param int|null $year
+     *
+     * @return static
+     */
+    public static function easterDate($year = null)
+    {
+        $instance = static::create($year);
+
+        $a = $instance->year % 19;
+        $b = floor($instance->year / 100);
+        $c = $instance->year % 100;
+        $d = floor($b / 4);
+        $e = $b % 4;
+        $f = floor(($b + 8) / 25);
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Nice! But it seems to me a bit too specific to be part of the core library. But I hope to put some day a list of custom macros, helpers and other Carbon extensions in our documentation, and this one could be relevant in this list.

kylekatarnls:
Will appear in the documentation:
https://github.com/briannesbitt/Carbon/compare/gh-pages...kylekatarnls:gh-pages-1.26#diff-eae600270579316beb8b8d9b9a5732d4R1433

36864:
PHP offers an `easter_days()` function that returns the number of days from March 21st to Easter. This can easily be used with carbon to get a year's easter date:
```
	public static function easterSunday($year){
		//returns easter date as number of days since march 21st
		//http://php.net/manual/en/function.easter-days.php
		//easter_date is an alternative but has issues with timezones and is limited to UNIX timestamp dates
		$easter = easter_days($year);

		$march21 = Carbon::createFromDate($year, 3, 21);

		return $march21->addDays($easter);
	}
```

kylekatarnls:
If it fails on >2038, I prefer to keep the current version, if not, we can update it.

36864:
easter_date() fails for year >2038. easter_days does not.
From the php docs: 
>This function can be used instead of easter_date() to calculate Easter for years which fall outside the range of Unix timestamps (i.e. before 1970 or after 2037).

kylekatarnls:
Thanks, changed.

PR #1054, opened 2017-12-15
Implement hasFormat() method

Propaganistas:
This PR adds a method to check if a string is in a given datetime format. Would be glad to see this merged.

Note: the regex patterns could perhaps be shortened but I preferred a readable pattern...

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2005,17 +2005,17 @@ class Carbon extends DateTime
     /**
      * Checks if the (date)time string is in a given format.
      *
+     * @param string $date
      * @param string $format
-     * @param string $time
      *
      * @return bool
      */
-    public static function isFormat($format, $time)
+    public static function hasFormat($date, $format)
     {
         try {
             // Try to create a DateTime object. Throws an InvalidArgumentException if the provided time string
             // doesn't match the format in any way.
-            static::createFromFormat($format, $time);
+            static::createFromFormat($format, $date);
 
             // createFromFormat() is known to handle edge cases silently.
             // E.g. "1975-5-1" (Y-n-j) will still be parsed correctly when "Y-m-d" is supplied as the format.
@@ -2025,7 +2025,7 @@ class Carbon extends DateTime
                 static::$regexFormats
             );
 
-            return (bool) preg_match('/^'.$regex.'$/', $time);
+            return (bool) preg_match('/^'.$regex.'$/', $date);
         } catch (InvalidArgumentException $e) {
         }
 
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -150,6 +150,54 @@ class Carbon extends DateTime
     );
 
     /**
+     * Format regex patterns.
+     *
+     * @var array
+     */
+    protected static $regexFormats = array(
+        'd' => '(3[01]|[12][0-9]|0[1-9])',
+        'D' => '([a-zA-Z]{3})',
+        'j' => '([123][0-9]|[1-9])',
+        'l' => '([a-zA-Z]{2,})',
+        'N' => '([1-7])',
+        'S' => '([a-zA-Z]{2})',
+        'w' => '([0-6])',
+        'z' => '(36[0-5]|3[0-5][0-9]|[12][0-9]{2}|[1-9]?[0-9])',
+        'W' => '(5[012]|[1-4][0-9]|[1-9])',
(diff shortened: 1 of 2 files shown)


Glavic:
Trying to guess what you are trying to do. Is ti a datetime validator, or if format checker so that you cannot enter format for example `H:i:š` ?

Can you also rebase this PR? TY

Propaganistas:
The suggested method validates if a string complies to a given date format.

E.g.
```php
Carbon::isFormat('Y-m-d', '1975-05-01'); // true
Carbon::isFormat('Y-m-d', '01-05-1975'); // false
Carbon::isFormat('d-m-Y', '01-05-1975'); // false
````

**Comment by Glavic** [comment]
So it is validator. Wouldn't method name `isValid` be more correct?

On other hand, why overkill with regex? 
What about this 2 line validator: http://php.net/manual/en/function.checkdate.php#113205 ?

**Comment by Propaganistas** [comment]
`isValid()` would seem to me like validating if a date exists, but this merely validates the format of the string.

The regex is necessary to handle cases like this:
```php
$format = 'D jS';
$date = 'Sun 21st';

Carbon::isFormat($format, $date); // true

$d = DateTime::createFromFormat($format, $date);
// At this very moment, there is no "February 21st on Sunday in 2018",
// so DateTime takes the next Sunday and hence $d represents 2018-02-25.
$d->format($format) == $date; // false (Sun 25th vs Sun 21st)
````

**Comment by kylekatarnls** [comment]
`isFormat` would mean comparing a "Y-m-d" to ISO8601 for example.

So it should rather be `isValid` or `matchFormat`, maybe `hasFormat`.

**Comment by Glavic** [comment]
Still cannot see any use for this :/
Why would I need a method that would validate my format?

Even if there is any use case, isn't this same as this oneliner ([example](https://eval.in/956252))?
```php
function isValidFormat($date, $format = 'Y-m-d H:i:s') {
    return is_object(DateTime::createFromFormat($format, $date));
}
```

Propaganistas:
The oneliner you suggested doesn't take into account edge cases like described in an inline comment:
```php
// createFromFormat() is known to handle edge cases silently.
// E.g. "1975-5-1" (Y-n-j) will still be parsed correctly when "Y-m-d" is supplied as the format.
// To ensure we're really testing against our desired format, perform an additional regex validation.
```

Propaganistas:
I have two use cases in my current project:
* Validate incoming strings from user input
* A legacy database where different date formats were used in a single column. Using the `parse()` method doesn't yield the right results in some cases

Glavic:
So we are back to validator, since this test: https://github.com/briannesbitt/Carbon/pull/1054/files#diff-cd3bb6154cd03a9e792e1946af8528f1R401

And in my book, date `$date = 'Sun 21st';` is invalid, don't see any reason why would method return `true` on that...

Note on both your case:
- for validation, [this](http://php.net/manual/en/function.checkdate.php#113205) can be used; like said before, on today, if you enter `Sun 21st`, it is invalid date.
- db stuff, like field with multiple/unknown date formats, is a horror :) it is no surprise that `parse()` doesn't know every possible combination some1 had in mind when entering date in the db...

Propaganistas:
> And in my book, date $date = 'Sun 21st'; is invalid, don't see any reason why would method return true on that...

Because it may be a string that was once constructed using `date('D jS')`, and the method merely checks if the string is in format `D jS`. Nothing more, nothing less. Hence it should return true, independently whether it applies to the current month/year.

Glavic:
True, `Sun 21st` is not even a date (maybe player name `Sun` was `21st` on his scoreboard?), since you don't know to which year and month it belongs. For that you need extra validation, but for all other date/time validations, those [2 lines](http://php.net/manual/en/function.checkdate.php#113205) a more than enough.

Is this too user specific problem, or it is a mass problem? Let see what others have to say about it.

Propaganistas:
I'd say its a nice addition to Carbon, not necessarily a mass problem.

briannesbitt:
This is about input validation rather than date creation.  For example if I actually try and create a date with the format "D jS" and the input of "Sun 21st" right not it gives me "2018-02-25 10:08:03" as a result.  The date would be considered invalid (because the closest Sunday is the 25th not 21st,) but the format according to my constraints is correct.

If we did include it I think it would be something like hasFormat()

Propaganistas:
Indeed @briannesbitt, you pinpointed the intents of the method.

I'll adapt the method's name accordingly.

PR #1055, opened 2017-12-20
Fix issue with mocking time on PHP >=7.1

### LINKED ISSUE #1033: Microseconds stripped when using `setTestNow()` and `now()`
Hello,

First, I love this library. Thanks to all for your hard work. You've certainly made things easier for me.

## Summary
It appears that the microseconds are being stripped off the object when calling the `Carbon::now()` method and using microseconds. 

## Steps to recreate
PHP version on my local machine is 7.1.8

screen shot 2017-11-16 at 4 27 49 pm

I placed the following code in a file called `test.php` within the `src` directory.

Then, I executed it via `php test.php` at the terminal.  

```php
use Carbon\Carbon;

require_once __DIR__.'/../vendor/autoload.php';

$now = new Carbon;
echo $now->format('Y-m-d H:i:s.u') . PHP_EOL;

$now->subDay();
echo $now->format('Y-m-d H:i:s.u') . PHP_EOL;

Carbon::setTestNow($now);

echo Carbon::getTestNow()->format('Y-m-d H:i:s.u') . PHP_EOL;
echo Carbon::now()->format('Y-m-d H:i:s.u') . PHP_EOL;

```

## Result

screen shot 2017-11-16 at 4 19 19 pm

As you can see, the microseconds are preserved when manipulating the date as well as after initially calling `Carbon::setTestNow($now)`. However, when calling `Carbon::now()` later on, the microseconds are reset to all zeroes.

I believe the issue occurs because of the `$time = $testInstance->toDateTimeString();` on [line 288](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php#L288) in `src/Carbon.php`. 

When substituting `$time = $testInstance->format('Y-m-d H:i:s.u')`, things work as expected, and all the tests pass.

screen shot 2017-11-16 at 4 24 39 pm

I'm happy to submit a PR for this, but would like a little guidance, first. 

A) Is this something that you see as needing to be fixed?
B) What is the preferred way to solve this (if not the method shown above)?

Thank you,
ds

jkrnak:
Since PHP 7.1 DateTime supports microseconds:
> 7.1 | From now on microseconds are filled with actual value. Not with '00000'.
http://php.net/manual/en/datetime.construct.php#refsect1-datetime.construct-changelog

The constructor when using a mocked time was converting date with the format string:  `Y-m-d H:i:s`. 
This meant that it was losing the microseconds part. 
When using a mocked time with comparison operators you could end up in some weird errors because the microseconds part was 0 on all mocked times.

```php
$c = new Carbon('2017-12-20 15:38:22.1234');
Carbon::setTestNow($c);

$mockedC = Carbon::now();
printf("%s\n", $c->format("Y-m-d H:i:s.u"));       // 2017-12-20 15:38:22.123400
printf("%s\n", $mockedC->format("Y-m-d H:i:s.u")); // 2017-12-20 15:38:22.000000
var_export($c == $mockedC);                        // false
```

This change adds microsecond to the mocked time.

Fixes #1033

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -119,6 +119,13 @@ class Carbon extends DateTime
     const DEFAULT_TO_STRING_FORMAT = 'Y-m-d H:i:s';
 
     /**
+     * Format for converting mocked time, includes microseconds.
+     *
+     * @var string
+     */
+    const MOCK_DATETIME_FORMAT = 'Y-m-d H:i:s.u';
+
+    /**
      * Format to use for __toString method when type juggling occurs.
      *
      * @var string
@@ -285,7 +292,7 @@ class Carbon extends DateTime
                 $tz = $testInstance->getTimezone();
             }
 
-            $time = $testInstance->toDateTimeString();
+            $time = $testInstance->format(static::MOCK_DATETIME_FORMAT);
         }
 
         parent::__construct($time, static::safeCreateDateTimeZone($tz));
--- a/tests/Carbon/ConstructTest.php
+++ b/tests/Carbon/ConstructTest.php
@@ -109,4 +109,15 @@ class ConstructTest extends AbstractTestCase
         $this->assertSame($timezone, $c->tzName);
         $this->assertSame(9 + $dayLightSavingTimeOffset, $c->offsetHours);
     }
+
+    public function testMockingWithMicroseconds()
+    {
+        $c = new Carbon(Carbon::now()->toDateTimeString().'.123456');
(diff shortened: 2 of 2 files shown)


Glavic:
@jkrnak: nice catch! Can you please rebase? +remove those travis fixes, since we already solved that.

jkrnak:
@Glavic I've rebased. 
Thanks for fixing the builds!

kylekatarnls:
Thanks for your PR!

PR #1058, opened 2017-12-25
Implemented CarbonRange to take work and weekend days count from a date range.

Denys-Bushulyak:
Example:
```php
$begin = Carbon::create(2017,12,1);
$end = Carbon::create(2017,12,31);
$range = new CarbonRange($begin, $end);
$range->getWeekendDays();
$range->getWorkingDays();
````

### DIFF
--- a/src/Carbon/CarbonRange.php
+++ b/src/Carbon/CarbonRange.php
@@ -1,0 +1,100 @@
+<?php
+/**
+ * User: denys
+ * Date: 25.12.17
+ * Time: 14:23
+ */
+
+namespace Carbon;
+
+
+use Carbon\Exceptions\UnknownIntervalException;
+
+class CarbonRange
+{
+    private $begin;
+    private $end;
+
+    public function __construct(Carbon $begin, Carbon $end)
+    {
+        $this->begin = $begin;
+        $this->end = $end;
+    }
+
+    /**
+     * @return Carbon
+     */
+    public function getEnd()
+    {
+        return $this->end;
+    }
+
+    /**
+     * @param Carbon $end
+     */
+    public function setEnd($end)
+    {
+        $this->end = $end;
+    }
+
+    /**
+     * @return Carbon
+     */
+    public function getBegin()
+    {
+        return $this->begin;
+    }
+
+    /**
+     * @param Carbon $begin
+     */
+    public function setBegin($begin)
+    {
+        $this->begin = $begin;
+    }
+
+    /**
+     * Return count of working days at interval.
+     * @param array $weekendDays Weekend days.
+     * @return int Count of working days.
+     * @throws \Exception
+     */
+    public function getWorkingDays($weekendDays = array(Carbon::SATURDAY, Carbon::SUNDAY))
+    {
+        $interval = $this->getIntervalAtDays();
+
+        $begin = clone $this->getBegin();
+
+        $count = 0;
+        for ($i = 0; $i < $interval; $i++) {
+
+            if (!in_array($begin->dayOfWeek, $weekendDays)) {
+                $count++;
+            }
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by kylekatarnls** [comment]
Hi, this seems nice. I notice you use `[]` array. For PHP 5.3 compatibility, you should use `array()` everywhere.

I also wonder if it should not be a dedicated project (not inside Carbon). I think we could add a note about CarbonRange in our documentation if you would create a package for this.

Other opinions? Rather a core class of Carbon, or a dedicated composer library?

**Comment by briannesbitt** [comment]
Thats a lot of extra code for methods we already have in Carbon ?  `diffInWeekdays()` and `diffInWeekdays()`

**Comment by Denys-Bushulyak** [comment]
> Other opinions? Rather a core class of Carbon, or a dedicated composer library?

When I created it I thinking about it as the core feature rather external library.

**Comment by imbrish** [comment]
Isn't anyone concerned that holidays are completely ignored? Maybe method should be renamed to `getWeekDays`... or strictly said `countWeekDays`.

**Comment by Denys-Bushulyak** [comment]
I update code. How to refresh pull request?

**Comment by imbrish** [comment]
Just push new commits to the branch you want to merge from (`Denys-Bushulyak:feature-carbon_range`).

**Comment by Denys-Bushulyak** [comment]
Thank you. Done.

**Comment by kylekatarnls** [comment]
Hi, the concept of the range is good but the only 3 methods in it getWorkingDays, getWeekendDays and getIntervalAtDays are too specific but we could imagine a magic `__call` method to call any diff method on $begin use $end as first argument, something like that:

```php
public function __call($method, $arguments)
{
    if (substr($method, 0, 4) !== 'diff') {
        $method = 'diff'.ucfisrt($method);
    }
    if (method_exists($this->begin, $method)) {
        array_unshift($arguments, $this->end);

        return call_user_func_array($this->begin, $arguments);
    }

    throw new \BadMethodCallException("Method {$method} does not exist.");
}
```

This way you can call any diff method in your range (and so also week and weekend days):
```php
$range->inWeekdays()
$range->inWeekendDays()
$range->inHours()
$range->diff() // return the native DateInterval
$range->forHumans(true, false, 6) // $begin->diffForHumans($end, true, false, 6)
...
```

What do you think of it?

Denys-Bushulyak:
Hi, Kyle. You right.

Denys-Bushulyak:
Guys, I am not exp. contributor: how I have to contribute my code? Merge my feature with master and make pull request to your master? Like now or somethings else?

imbrish:
After you submitted pull request now it's up to repository maintainer to merge it.

In my opinion your feature haven't been merged yet, because it is very specific to your use case, and you didn't seem to try to make it more generic by implementing Kyle's suggestions.

However now I started to work on more robust extension of built-in DatePeriod. When it's done it would make sense to re-submit your features there!

Denys-Bushulyak:
@kylekatarnls , I will be available at May and do that.

imbrish:
@kylekatarnls it could be possible to extend CarbonPeriod's `__call` with aliases for Carbon's `is*` methods that'll add automatic filters:
```php
$weekdaysInDesember = CarbonPeriod::create('2017-12-01', '2017-12-31')->isWeekday()->count();
```

kylekatarnls:
I would prefer something like: `$weekdaysInDesember = CarbonPeriod::create('2017-12-01', '2017-12-31')->map('isWeekday')->count();`
because `isSomething` should always return a boolean.

kylekatarnls:
Documented via: https://github.com/kylekatarnls/Carbon/commit/ef0fcae5af83ed0521c1588d729fa0d8de16679c

Will be in the 1.28 documentation on release.

PR #1060, opened 2018-01-05
Update NL lang

BackEndTea:
In dutch both seconden and secondes is valid.


jmarcher:
But if both are valid why having both?

PR #1063, opened 2018-01-10
Changed toIso8601String helper function 

afrojuju1:
Changed `toIso8601String` to take in a boolean which when false would remove the timezone from ISO8601 string.
This adds extra flexibility around ISO8601 string format.


Glavic:
If you need custom format, why don't you use `->format()` method?

kylekatarnls:
Each `->toX` method should be reserved for one format. I recommend you to extend the class (or use a macro if you work with Laravel) to create an other method for your custom format.

briannesbitt:
We would implement this as a separate method anyway, no boolean param.  But regardless, doesn't need to go into Carbon.  Sounds like a good place for your own class.

PR #1070, opened 2018-01-15
Reverted pull-request #548 about issue #88

Ovsyanka:
The realisation of initial issue #88 is wrong. The problem does not relies to `local` / `not local` timezone. Php handles date changes similar for both cases. So in fact changes was made 'fixes' non-local dates but local dates works same as before.
And it 'fixes' the problem and not actually fixes it because in fact the PHP handling of DST changing in non-UTC dates is quite right. If developer don't want to adjusting timestamp to DST change he should use UTC timezone.


Ovsyanka:
There is style check errors in the code I didnt channge. I don't knw why it is happens.

Glavic:
@Ovsyanka: could you please rebase + return those tests you removed?

Ovsyanka:
@Glavic It would be wrong to keep this tests because they assume wrong behavior.
Actually I can add another tests that will assure php default behavior. I will add it soon.

Glavic:
I kinda disagree, because those tests are [valid with `DateTime` ](https://eval.in/955872)and they should also be valid with `Carbon`.

Ovsyanka:
This tests isn't valid. Dates you printed have 23 hours difference, not 24. Look: https://eval.in/955877

Glavic:
I think there is a issue in `diffInSeconds()` method, where unix timestamp is subtracted from other unix timestamp... Also `DateTime::diff()` returns 1 day: https://eval.in/955883
If I add 24 hours to X, and then compare it with X, I need to get back 24 hours, not 23 or 25.
I would also fix `diffInSeconds()`: https://github.com/Glavic/Carbon/commit/97d26223f737b445903daead844b1a506dd46401

Ovsyanka:
I think we should not mixing different problems in one place. Initially I removed the tests exactly because that (just forgot the reason and rushed in to finish that PR). In this PR I just revert wrong change, that was implemented before. Now the Carbon acts like DateTime. Then, we can discuss `diff` methods. By now I think they doing good, but again - I suggest to discuss it in other place (new issue, for example). There I will explain my opinion.

I will meke another tests soon, that will not involve diff methods.

Glavic:
I see from your [example](https://eval.in/955877) that you get date `2014-03-30T00:00:00+00:00` and `2014-03-31T00:00:00+01:00` and that is 23 hour difference (php bug).

Add format `I` (capital i) to you format, and you will see if date is in DST or not, like [this](https://eval.in/956034). You will get `2014-03-30T00:00:00+00:00 0` and `2014-03-31T00:00:00+01:00 1`. Which is still wrong, because it should be `2014-03-31T01:00:00+01:00 1` (php bug).

For test, lets first add 1hour and then 23hour, see [here](https://eval.in/956094). We get correct result 24hour diff as `2014-03-31T01:00:00+01:00 1`. 
But [here](https://eval.in/956098) `DateTime::diff()` returns 25 hours (1day + 1hour)?

But ˙DateTime::diff()` will work if adding +24hours, see [here](https://eval.in/956235).

(php bug) This is probably [PHP bug 74274](https://bugs.php.net/bug.php?id=74274). I think you already mentioned that in issue: https://github.com/briannesbitt/Carbon/pull/1071.

I would suggest that carbon is acting as default php functions.

But `X + 24 = Y`, `Y - X` cannot be 23 :( And those tests that you deleted should work :/

Ovsyanka:
I think that is not the php bug but intentional behaviour. I can't be sure because I didn't saw that in the documentation, but in the suggested in the [74274](https://bugs.php.net/bug.php?id=74274) [RFC](https://wiki.php.net/rfc/datetime_and_daylight_saving_time) you can see, that timestamp adjusting is expected. And it has errors in some cases, but it is another story.

For example:
> 2010-03-14T04:30:00(DT) | 2010-03-13T04:30:00(ST) | P1DT0H

Here we see, that expected 1 day difference, but in fact there is 23 hours between this dates.

> I would suggest that carbon is acting as default php functions.
I agree.

> But X + 24 = Y, Y - X cannot be 23 :( And those tests that you deleted should work :/

I am not sure that they should. In the [RFC](https://wiki.php.net/rfc/datetime_and_daylight_saving_time) I referenced we can see, that factical diff expected. For example:

> 2010-03-14T03:30:00(DT) | 2010-03-13T04:30:00(ST) | P0DT22H

```
// according to RFC it is OK in cases of DST change.
$date - $date->addHours(24) === 23
// But add / sub the same time should be revertable (this is php bug)
$date === $date->addHours(24)->subHours(24) 
``

And again, I suggest not to discuss it here. This PR isn't about that =)

**Comment by kylekatarnls** [comment]
Hi, my  opinion on this is: [this case](https://eval.in/955877) is a PHP bug. It's not reliable to get a 23 hours difference after adding 24 hours. I agree to get 23h only if adding days units because it's DST and because we can consider 1 day = 25 hours on DST start and 25 hours at the end, 24 hours all other days. But if we add hours we should not get an other amount of hours as difference.

I understand that if we forget the DST, adding 24 hours to 00:00 and get 01:00 can lead to unexpected behaviour, but it's not our fault is DST exist and it's up to the developers to watch the timezone offset change (+/-1h) and handle it in their app.

So:
- I agree with your unit tests but right now, they pass with the current code. So your unit tests are not a sufficient reason to remove modify().
- I disagree to remove the modify unit tests, it would be a breaking change, so if you remove modify(), you must provide something instead to keep those unit tests working.

I expect a solution that don't remove any existing tests (a solution with no BC), so this can be a way: https://github.com/Glavic/Carbon/commit/97d26223f737b445903daead844b1a506dd46401

But first I also expect a unit test that exactly describe the bug (unexpected behavior) we're talking about. I expect this unit test to fail with the current code and work if we remove modify() or change something about it.

**Comment by Ovsyanka** [comment]
Here it is. The new tests.

**Comment by kylekatarnls** [comment]
Thanks, I cannot merge as is since removing tests means breaking change. But I will cherry-pick your tests and the @Glavic diff method fix into an other branch.

**Comment by Ovsyanka** [comment]
Yes, it is "breaking" change as any bug fixing. It always "breaks" uncorrect behavior and changes it to the correct one.
Seems like I misunderstood you or something. Could you explain in other words why you can't merge this PR?

**Comment by kylekatarnls** [comment]
I agree to break behavior we never document and never say it works that way. But I disagree to break something already tested and documented:
```php
$a = new Carbon('2014-03-30 00:00:00', 'Europe/London');
$b = $a->copy();
$b->addHours(24);
$this->assertSame(24, $a->diffInHours($b));
```
This **must** still pass. So I agree to remove modify(), if we fix accordingly diffInHours and other functions that rely on it. It's not OK to remove it and let things become inconsistent.

That's why I talked you about this fix for diffInHours: https://github.com/Glavic/Carbon/commit/97d26223f737b445903daead844b1a506dd46401

Ovsyanka:
I think i got you. But. This is just poor test for the changes maked in #88. It is not relevant. I will try to explain.

What this testcase ensures? That when we adding some hours we will have same hour difference. This should be achieved though changing `diffInHours` (and etc.) methods. But in the #88 this achieved by changind modify method and that leads to lot of errors (issues here). In fact #88 did not solve the problem and though the test is passing, claimed functionality didn't be implemented right.

This PR not making this functionality work, so yes, this test will not be passed, but it solves errors, that was implemented in trying to make this test pass in #88.

By the way, as I said before - I can't completely agree with you that in case of non UTC timezone this test should pass. It is differ from PHP default behavior and is questionable.

Ovsyanka:
Wait a moment, I made a quite mistake. I was wrongly confident about test you talking about is failing in my PR. But seems like it doesn't. I will investigate this some more.

kylekatarnls:
Yes it does but it pass if we use the @Glavic fix. And I really prefer to merge it too.

kylekatarnls:
This is what propose: #1129
I did not add any code, it's just yours and the @Glavic fix and it removes any tests and all of them pass.

Ovsyanka:
> I was wrongly confident that test you talking about is failing in my PR. But seems like it doesn't.

I was right, they fails indeed. So...

I personally don't like to mix these different things in one PR. But it is not important for me, so I will not argue about this any more. I suggest @Glavic to cherry-pick my tests because #1129 tests do not cover #88 implemented errors.

kylekatarnls:
#1129 is about #88, it reverts the first fix and apply another fix, so before and after the merged, #88 will remain fixed (that's my goal, avoid regression before redoing things better). And it does not contain any work-around or hack, it respects the native PHP DateTime behavior.

@Glavic pointed out the fact that date interval should not use timestamp and so he proposed a diffInSeconds method that do not use timestamp but date interval stuff to calculate seconds. And if someone want a timestamp difference, he still can use `->getTimestamp()` and go on its own with that how timestamp works (skip/jump seconds on DST), but for the Carbon diffInSeconds/Minutes/Hours method, I think the @Glavic method is better because it returns seconds that have really existed between true moments a contrario to `$b->getTimestamp() - $a->getTimestamp()`. I mean it's the exact number of clicks on your clock. But if you still want to get 23 after adding 24 hours, you can by getting the timestamp difference and divide it by 3600. But it's not something we should call "hours". That could be some methods named `timestampDiff` with some divider parameter:
```php
(int) $a->timestampDiff() / (static::MINUTES_PER_HOUR * static::SECONDS_PER_MINUTE);
```

I hope you find this analyse relevant because I really want to solve this in a way that respect your work and give working solution for everyone.

Ovsyanka:
> I hope you find this analyse relevant because I really want to solve this in a way that respect your work and give working solution for everyone.

I need quite a time to carefuly read and understand what you wrote about @Glavix's changes (I already can say that we need to clarify some base terms to talk on the "same language"). But I don't see nessesarity of discassion about Glavic's changes at this point. Why you made this analyse? Do you want me to include it in my PR? Or you just want to make me agree with you to be sure that you doing the right thing? Or what?

Sorry if my sentenses looks dry or injured - it is just my poor english.

What do you think about my suggestion to just use my tests? You said by yourself similar thought before:

> Theoretically, you should be able to cherry-pick this commit and so keep the modify unit test without breaking any test.

Ovsyanka:
We decided to include my tests to #1129 PR. Now this PR can be closed.

PR #900, opened 2017-02-13
Fixing #871, workaround of php bug [72338](https://bugs.php.net/bug.php?id=72338)

Ovsyanka:

PR #1071, opened 2018-01-16
Workaround about php bug 74274

Ovsyanka:
By now it is just tests that I want to discuss before trying to fix it. Questions and suggestions are appreciated. If anyone could help me with proper documentation it will be very nice.

I made tests to check this situations:

1. Each modification should be reverted. If you add one hour and then sub one hour the date shouldn't be changd. There is testAddHoursReversable test.

2. When you create date from timestamp - it should have this initial timestamp. There is testTimestampNotChanged test.

3. When you crossing DST change in the local timezones PHP adjusting timestamp  to looks like offset didn't changed. It works well. But if you change date into the skipped or repeated hour there is errors. There is testAddHours test for that.

4. If you creating Date from timestamp into repeated hour PHP always think that it is Standard Time and not the Daylight Time. There is getCreateFromTimestampTests for that.


Ovsyanka:
Sadly, by now I don't see effective way to fix it without fixing PHP itself.

Glavic:
@Ovsyanka: what is the status here?

Ovsyanka:
My last comment still actual. The problem is the DateTime can't represent local standard time in repeated hour wile backward transition.
Example:
When you try to set 1289109600, that mean `2010-11-07T02:00:00-04:00` or `2010-11-07T01:00:00-05:00` the php always changes timestamp to `1289106000`, that means `2010-11-07T01:00:00-04:00`.

The only workaround I see is to add some property to save right timestamp and then use it in all methods instead internal timestamp. I am not sure it worth it. I will be glad to know your thoughts about that.

I made the pull-request because I think my tests 1) shows the problems 2) someone can use it to try to solve the issue.

kylekatarnls:
This PR also need to replace `[]` with `array()` since Carbon supports PHP 5.3.

kylekatarnls:
Hi @Ovsyanka, it seems to me you could merge this with #1070, if not please tell me and I'll reopen.

Ovsyanka:
Hmm. This PR have only **failing** tests, those indicates the bug. I may not merge failing tests. Did I misunderstand you?

kylekatarnls:
This one couldn't be merged (failing tests, compatibility and style) but the source modification is the same than in #1070, so if there is something in this PR you still want to address, it seems better to me to put it in your other PR.

Ovsyanka:
Ah, it is just the same commits from another pull-requests because this one needs them to work. They are not the part of this PR actually and will be removed after rebase to the master. This PR **is** about the failing tests. I already wrote why I made it and if you think that it is right to close it - then OK.

Ovsyanka:
> I see, you created pr-WaPhpBug74274 from pr-revert-548-modify-errors. So this PR has nothing to be merged. Understood.

Yep, seems like you got it =)

> If PHP decided to adjust timestamp with DST, there's nothing we should do.

In case if it doing it right - yes. But it has errors (look at the my first commentary https://github.com/briannesbitt/Carbon/pull/1071#issue-163190434). And I think that it would be great if we could make this working right in Carbon library like there is no error in PHP. Like in #900. But by now I don't see good way to do it. And this PR is the only place where I imagine to put tests, that describes the bug. Now someone could try to fix that using my tests as goal. Personally I not suffer about this bugs - I prefere to use UTC dates most the time instead of local, but in general the errors exists so I decided to share what I already did in this matter.

PR #1084, opened 2018-02-11
Adds ability to pass a timestamp to the constructor

regularlabs:
This makes it possible to pass a timestamp to the Carbon class, like:
```
$date = new Carbon(1518358828);
```

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -265,7 +265,7 @@ class Carbon extends DateTime
      * Please see the testing aids section (specifically static::setTestNow())
      * for more on the possibility of this constructor returning a test instance.
      *
-     * @param string|null               $time
+     * @param string|int|null           $time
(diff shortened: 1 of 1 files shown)


Glavic:
Timestamp can already passed to the constructor with `@` prefix:

```php
$date = new Carbon("@1518358828");
```

regularlabs:
That assumes you are using UTC timestamps.
It also requires you to check what the format is of your date before creating the Carbon instance.
If you need to do that, then passing it to the constructor is not necessary at all. Then you can simply do:
```
$date = Carbon::createFromTimestamp(1518358828);
```

The idea behind this PR is that you can pass it any date/datetime value without needing to worry about what format it has.

Glavic:
Unix timestamp should always be UTC. If that is not true in your case, then you are doing something "wrong". On other hand, I would never ever mix `date()` functions with superior `DateTime` class. Just my 2 cents.

PR #1088, opened 2018-02-13
Document overflow methods

kylekatarnls:
Documentation for:
- static useMonthsOverflow
- static resetMonthsOverflow
- static shouldOverflowMonths
- addMonthsWithOverflow
- addMonthWithOverflow
- addMonthsNoOverflow
- addMonthNoOverflow
- subMonthsWithOverflow
- subMonthWithOverflow
- subMonthsNoOverflow
- subMonthNoOverflow

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -114,7 +114,7 @@ echo Carbon::parse('first day of December 2008')->addWeeks(2);    // 2008-12-15 
 <p>To accompany <code>now()</code>, a few other static instantiation helpers exist to create widely known instances.  The only thing to really notice here is that <code>today()</code>, <code>tomorrow()</code> and <code>yesterday()</code>, besides behaving as expected, all accept a timezone parameter and each has their time value set to <code>00:00:00</code>.</p>
 
 <p><pre><code class="php">$now = Carbon::now();
-echo $now;                               // 2018-02-13 07:06:49
+echo $now;                               // 2018-02-13 07:37:34
 $today = Carbon::today();
 echo $today;                             // 2018-02-13 00:00:00
 $tomorrow = Carbon::tomorrow('Europe/London');
@@ -183,14 +183,21 @@ echo $dt->copy()->micro;                               // 123456
 echo Carbon::minValue();                               // '1901-12-13 15:45:52'
 </code></pre></p>
 
+<p>Min and max value mainly depends on the system (32 or 64 bits).</p>
+
+<p>With a 32 bits OS system (you can check it in PHP with <code>PHP_INT_SIZE === 4</code>), the minimum value is the 0-unix-timestamp (1970-01-01 00:00:00) and the maximum is the timestamp given by the constant <code>PHP_INT_MAX</code>.</p>
+
+<p>With a 64 bits OS system, the minimum is 01-01-01 00:00:00 and maximum is 9999-12-31 23:59:59.</p>
+
 <h1 id="api-localization">Localization</h1>
 
 <p>Unfortunately the base class DateTime does not have any localization support.  To begin localization support a <code>formatLocalized($format)</code> method was added.  The implementation makes a call to <a href="http://www.php.net/strftime">strftime</a> using the current instance timestamp.  If you first set the current locale with PHP function <a href="http://www.php.net/setlocale">setlocale()</a> then the string returned will be formatted in the correct locale.</p>
 
 <p><pre><code class="php">setlocale(LC_TIME, 'German');
+echo $dt->formatLocalized('%A %d %B %Y');          // Mittwoch 21 Mai 1975
+setlocale(LC_TIME, 'English');
 echo $dt->formatLocalized('%A %d %B %Y');          // Wednesday 21 May 1975
-setlocale(LC_TIME, '');
-echo $dt->formatLocalized('%A %d %B %Y');          // Wednesday 21 May 1975
+setlocale(LC_TIME, ''); // reset locale
 </code></pre></p>
 
 <p><code>diffForHumans()</code> has also been localized.  You can set the Carbon locale by using the static <code>Carbon::setLocale()</code> function.</p>
@@ -228,7 +235,7 @@ echo Carbon::parse('now');                             // 2001-05-21 12:00:00
 var_dump(Carbon::hasTestNow());                        // bool(true)
 Carbon::setTestNow();                                  // clear the mock
 var_dump(Carbon::hasTestNow());                        // bool(false)
-echo Carbon::now();                                    // 2018-02-13 07:06:49
+echo Carbon::now();                                    // 2018-02-13 07:37:34
 </code></pre></p>
 
 <p>A more meaning full example:</p>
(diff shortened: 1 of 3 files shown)


kylekatarnls:
Shit, my locale is taken into account for the auto-generation. I will add somthing for that in the generate.php script.

docs/index.html
```diff
@@ -179,8 +179,8 @@ <h1 id="api-instantiation">Instantiation</h1>
 
 <p>Ever need to loop through some dates to find the earliest or latest date?  Didn't know what to set your initial maximum/minimum values to? There are now two helpers for this to make your decision simple:</p>
 
-<p><pre><code class="php">echo Carbon::maxValue();                               // '2038-01-18 22:14:07'
-echo Carbon::minValue();                               // '1901-12-13 15:45:52'
+<p><pre><code class="php">echo Carbon::maxValue();                               // '9999-12-31 23:59:59'
```
Glavic: That really depends on system OS (32 or 64bit).
https://en.wikipedia.org/wiki/Year_2038_problem
Should we also comment that?

docs/index.html
```diff
@@ -225,10 +225,12 @@ <h1 id="api-testing">Testing Aids</h1>
 echo Carbon::parse();                                  // 2001-05-21 12:00:00
 echo new Carbon('now');                                // 2001-05-21 12:00:00
 echo Carbon::parse('now');                             // 2001-05-21 12:00:00
-var_dump(Carbon::hasTestNow());                        // bool(true)
+var_dump(Carbon::hasTestNow());                        // C:\Users\bastien.miclo\Perso\Carbon\generate.php(44) : eval()'d code:1:
```
Glavic: `C:\Users\bastien.miclo\Perso\Carbon\generate.php(44)` should that be here?

docs/index.html
kylekatarnls: I was exactly thinking of that. :)

docs/index.html
```diff
@@ -225,10 +225,12 @@ <h1 id="api-testing">Testing Aids</h1>
...
+bool(true)
 Carbon::setTestNow();                                  // clear the mock
-var_dump(Carbon::hasTestNow());                        // bool(false)
-echo Carbon::now();                                    // 2016-06-24 15:18:34
+var_dump(Carbon::hasTestNow());                        // C:\Users\bastien.miclo\Perso\Carbon\generate.php(44) : eval()'d code:1:
```
Glavic: ... and here?

docs/index.html
kylekatarnls: Shit, xdebug must be disabled for the script.

docs/index.html
```diff
@@ -426,7 +458,7 @@ <h1 id="api-commonformats">Common Formats</h1>
 // $dt->toAtomString() is the same as $dt->format(DateTime::ATOM);
 echo $dt->toAtomString();      // 1975-12-25T14:15:16-05:00
 echo $dt->toCookieString();    // Thursday, 25-Dec-1975 14:15:16 EST
-echo $dt->toIso8601String();   // 1975-12-25T14:15:16-0500
+echo $dt->toIso8601String();   // 1975-12-25T14:15:16-05:00
```
Glavic: Before that, we should investigate: https://github.com/briannesbitt/Carbon/pull/977

kylekatarnls:
I cleaned it a bit but I must fix this: https://github.com/briannesbitt/Carbon/pull/1088/files#diff-eae600270579316beb8b8d9b9a5732d4L189

kylekatarnls:
In fact, the example assumes the default locale is en, however it depends on the system. So I will rather change the example.

kylekatarnls:
min and max documented according to 32/64 bits conditional

docs/index.src.html
```diff
@@ -136,14 +136,21 @@ <h1 id="api-instantiation">Instantiation</h1>
 {{minValue::exec(echo Carbon::minValue();/*pad(54)*/)}} // '{{minValue_eval}}'
 </code></pre></p>
 
+<p>Min and max value mainly depends on the system (32 or 64 bits).</p>
```
briannesbitt: The min and max values depend on the bit count of the system (32 or 64 bit).

docs/index.src.html
```diff
@@ -136,14 +136,21 @@ <h1 id="api-instantiation">Instantiation</h1>
...
+
+<p>With a 32 bits OS system (you can check it in PHP with <code>PHP_INT_SIZE === 4</code>), the minimum value is the 0-unix-timestamp (1970-01-01 00:00:00) and the maximum is the timestamp given by the constant <code>PHP_INT_MAX</code>.</p>
```
briannesbitt: With a 32-bit OS system (you can check it in PHP using <code>PHP_INT_SIZE === 4</code>), the minimum value is 0-unix-timestamp (1970-01-01 00:00:00) and the maximum is the timestamp given by the constant <code>PHP_INT_MAX</code>.

docs/index.src.html
```diff
@@ -136,14 +136,21 @@ <h1 id="api-instantiation">Instantiation</h1>
...
+
+<p>With a 64 bits OS system, the minimum is 01-01-01 00:00:00 and maximum is 9999-12-31 23:59:59.</p>
```
briannesbitt: With a 64-bit OS system, the minimum is 01-01-01 00:00:00 and maximum is 9999-12-31 23:59:59.

docs/index.src.html
```diff
@@ -532,6 +539,61 @@ <h1 id="api-addsub">Addition and Subtraction</h1>
 
 <p>P.S. Don't worry if you forget and use <code>addDay(5)</code> or <code>subYear(3)</code>, I have your back ;)</p>
 
+<p>By default, Carbon behave as the PHP DateTime do. That means adding or substracting months can overflow, example:</p>
```
briannesbitt: By default, Carbon relies on the underlying parent class PHP DateTime behavior.  As a result adding or subtracting months can overflow, example:

docs/index.src.html
```diff
@@ -532,6 +539,61 @@ <h1 id="api-addsub">Addition and Subtraction</h1>
...
+
+<p><pre><code class="php">{{::lint($dt = Carbon::create(2017, 1, 31, 0);)}}
+
+{{addoverflow1::exec(echo $dt->copy()->addMonth();/*pad(40)*/)}} // {{addoverflow1_eval}}
+{{addoverflow2::exec(echo $dt->copy()->subMonths(2);/*pad(40)*/)}} // {{addoverflow2_eval}}
+</code></pre></p>
+
+<p>You can't prevent the overflow with <code>Carbon::useMonthsOverflow(false)</code></p>
```
briannesbitt: You **can** prevent

docs/index.src.html
```diff
@@ -532,6 +539,61 @@ <h1 id="api-addsub">Addition and Subtraction</h1>
...
+$dt = Carbon::create(2017, 1, 31, 0);)}}
+
+{{addoverflow3::exec(echo $dt->copy()->addMonth();/*pad(40)*/)}} // {{addoverflow3_eval}}
+{{addoverflow4::exec(echo $dt->copy()->subMonths(2);/*pad(40)*/)}} // {{addoverflow4_eval}}
+
+// Call the method with true to allow overflow again
+{{reset1::exec(Carbon::resetMonthsOverflow();)}}} // same as Carbon::useMonthsOverflow(true);
+</code></pre></p>
+
+The method <code>Carbon::shouldOverflowMonths()</code> allow you to know if the overflow is currently enabled.
```
briannesbitt: allow**s**

PR #1092, opened 2018-02-14
Coverage

kylekatarnls:
- Add unit tests and assertions
- Remove non relevant comments
- Add coverage annotations for 32-bit and Windows specific stuff


tests/CarbonInterval/ConstructTest.php
```diff
@@ -235,4 +235,9 @@ public function testInstanceWithDaysThrowsException()
     {
         CarbonInterval::instance(Carbon::now()->diff(Carbon::now()->addWeeks(3)));
     }
+
+    public function testCallInvalidStaticMethod()
+    {
+        $this->assertNull(CarbonInterval::anything());
```
briannesbitt: I think this is fine but interesting looking back we don't throw an exception  or something.

PR #1095, opened 2018-02-14
Cover all languages

kylekatarnls:
Add unit tests for missing languages, then a unit test to check that a language has a matching test, so build will fail if a new language is added without its test.


Glavic:
Nice job!

PR #1098, opened 2018-02-15
Alias for `now()`

sagerman4:
- Carbon is absolutely the best way to use dates with PHP.
- As such, it is the authority.
- Authority can be daunting, to some.
- Carbon should lighten things up to avoid intimidation.
- Also, it is 16 years to the day since Super Troopers was released. So.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -344,6 +344,18 @@ class Carbon extends DateTime
     }
 
     /**
+     * Get a Carbon instance for the current date and time, for fun people.
+     *
+     * @param \DateTimeZone|string|null $tz
+     *
+     * @return static
+     */
+    public static function meow($tz = null)
+    {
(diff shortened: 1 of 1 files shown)


kylekatarnls:
Hi,

It make no sense if you don't create a SuperTrooper that extends the Carbon class, there `SuperTrooper::meow()` would make sense.

Then, each new method requires unit tests and documentation.

Then my personal opinion about this would be: we have more than 200 methods, I still have many of them to document, we have really enough work to do without starting to add new aliases. So I vote for: no merge because I have no sense of humour. You know, I'm the wowser of the team.

briannesbitt:
Sorry but I have to agree.  Thanks for the contribution though.

sagerman4:
I completely understand. This implementation would have made much more sense had I aliased Carbon as Farva.

That's on me, guys. 

Thanks for the consideration!

PR #1100, opened 2018-02-16
Fix relative modifiers list

briannesbitt:

PR #1101, opened 2018-02-16
Add docs for Carbon::setUtf8

briannesbitt:
Add docs for Carbon::setUtf8 (didn't bother to generate ... wait for ……#1100 )


kylekatarnls:
In fact it did not depend on the language, it depend if you use fr_FR.UTF8 or just fr_FR for example. I merge and I will complete and re-generate.

Issue #1081, opened 2018-02-04
formatLocalized doesn't work with custom toStringFormat


stueliueli:
I used to call `Carbon::setToStringFormat()` in my application.
Because of this, the `$carbon->formatLocalized()` does not work anymore.
This is because `strtotime()` takes a Carbon-Object, which then is converted to a string according to the [Specification](http://php.net/manual/en/function.strtotime.php)

> The function expects to be given a string containing an English date format and will try to parse that format into a Unix timestamp

A successful solution could look like this:
- User calls formatLocalized
- Current toStringFormat is being stored
- Default toStringFormat is set
- strtotime() is called
- Custom toStringFormat is re-set

Or `strtotime` could simply be called with `$carbon->format('Y-m-d H:i:s');`

ricardobucho:
Hey there,

I don't see how this is an issue.

I might be missing something here but the default should be the carbon object and not the string format.  You can just call `formatLocalized` whenever you wish or any other method for that matter.

You could just as easily extend the carbon class or add an helper function somewhere else that does something like` $carbonObject->toDateTimeString()`.

Hope I am not misreading the situation here, best of luck :)

stueliueli:
Yes, yes, you're right. Carbon itself works with carbon objects. But the carbon-method `formatLocalized` calls the PHP date method `strtotime` with `$this` as argument.

Relevant code from Carbon.php lines 1179-1190:

```php
/**
 * Format the instance with the current locale.  You can set the current
 * locale using setlocale() http://php.net/setlocale.
 *
 * @param string $format
 *
 * @return string
 */
 public function formatLocalized($format)
    {
        // Check for Windows to find and replace the %e
        // modifier correctly
        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
            $format = preg_replace('#(?<!%)((?:%%)*)%e#', '\1%#d', $format);
        }

        $formatted = strftime($format, strtotime($this));

        return static::$utf8 ? utf8_encode($formatted) : $formatted;
    }
```

The problem here is the line `$formatted = strftime($format, strtotime($this));` or specific the part `strtotime($this)`. 

[strtotime](http://php.net/manual/en/function.strtotime.php) expects the parameter to be a string. So php implicitly parses `$this` to a string. This, using the function `__toString()`. 
This function on a carbon object returns the time formatted according to the format provided by `Carbon::setToStringFormat()`. 
[...]

PR #1107, opened 2018-02-17
Fix issue [#1081](https://github.com/briannesbitt/Carbon/issues/1081)

stueliueli:


src/Carbon/Carbon.php
```diff
@@ -1227,8 +1227,13 @@ public function formatLocalized($format)
             $format = preg_replace('#(?<!%)((?:%%)*)%e#', '\1%#d', $format); // @codeCoverageIgnore
         }
 
+        $customToStringFormat = self::$toStringFormat;
+        self::resetToStringFormat();
+
         $formatted = strftime($format, strtotime($this));
```
Glavic: Why don't we just replace 
`$formatted = strftime($format, strtotime($this));` 
with 
`$formatted = strftime($format, $this->getTimestamp());`
?

Glavic:
Other issue I see here is timezone. `strftime()` will return formated date based on user's global timezone...

stueliueli:
@Glavic Yes, but that was already described in the PhpDoc method header: 

> Format the instance with the current locale.  You can set the current locale using setlocale() http://php.net/setlocale.

Would there be any other way this could be done?

Glavic:
What was already described in phpdoc? Missed that.

Only thing that I could think of now is that we would need to set global timezone, format, and then revert global tz...

src/Carbon/Carbon.php
stueliueli: Very good point. Knowing `getTimestamp` makes the use of `strtotime` obsolete.

stueliueli:
Ah, found the reason: getTimestamp returns the timestamp of the current user's timezone. 
According to unitTest `StringsTest::testToLocalizedFormattedTimezonedDateString`, the input is `1975-12-25 14:15:16` which corresponds to the timestamp `188748916`.
If I pass this timestamp to strftime as parameter, I get the output `1975-12-25 09:15:16`, because the carbon-object was created using timezone "Europe/London" and I'm currently in Timezone "Europe/Berlin". 

I don't know if there's any workaround to this @Glavic

src/Carbon/Carbon.php
stueliueli: See https://github.com/briannesbitt/Carbon/pull/1107#issuecomment-366469612

Glavic:
I was talking about the timezone, not locale. You mentioned phpdoc, so I thought you are referring to php.net for locale function :)

Like I said, we could save global timezone, set our timezone, format the date, revert the global timezone, like this: 

```php
$defaultTz = date_default_timezone_get();
date_default_timezone_set($this->getTimezone()->getName());
$formatted = strftime($format, $this->getTimestamp());
date_default_timezone_set($defaultTz);
```

Or we could just let `strtotime()` do the magic and use:
```php
$formatted = strftime($format, strtotime($this->format($this::DEFAULT_TO_STRING_FORMAT)));
```

briannesbitt:
I think just calling `toDateTimeString()` is safer because the `DEFAULT_TO_STRING_FORMAT` could in theory change.  But skipping messing with the `self::resetToStringFormat()` is definitely ideal.

stueliueli:
Oof, always something new to learn. Didn't know `toDateTimeString` but using it seems really the best way. Fixed in Commit https://github.com/briannesbitt/Carbon/pull/1107/commits/c2fdfd7b20190e9efa62168ecadc2facc5c57bdb

kylekatarnls:
I think we have a similar PR, now it has been changed Just have to check that.

briannesbitt:
#1032

Issue #974, opened 2017-07-05
isSame* methods should take DateTime instead of Carbon as argument


pzaj2:
I believe that
`isSame*` functions should accept `DateTime` instead of `Carbon`. There is no reason not to support this functionality, since it all goes down to `isSameAs` method which uses `format` method of `DateTime` class.

What do you think?

Glavic:
Sorry for late reply.
Problem I see is only `isSameDay()` method, because it calls `->toDateString()` on given object, and `DateTime` doesn't have that method, so fatal error will aprear.
Could you do a PR for a fix?

Glavic:
@pzaj2: did you mean something like this? => https://github.com/Glavic/Carbon/commit/cc7c98ed4306a256a0ce3ccb2ff749f3af30a0f9

Sorry for late reply :(

pzaj2:
Haha, I have completely forgotten about this issue, but yes, that seems about right.

Glavic:
I have opened a PR. We will see how people react to that. TY for the issue. Will close this now, since we have a PR.

PR #1109, opened 2018-02-18
Methods `isSame**()` should also accept \DateTime objects.

Glavic:
Issue #974.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -15,6 +15,7 @@ use Carbon\Exceptions\InvalidDateException;
 use Closure;
 use DatePeriod;
 use DateTime;
+use DateTimeInterface;
 use DateTimeZone;
 use InvalidArgumentException;
 use Symfony\Component\Translation\Loader\ArrayLoader;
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Hi,

Not ready to merge IMHO. Code like this: `$carbon->isSameDay('2010-01-03')` produced before a explicit error message like `isSameDate argument#1 must be a Carbon instance`, now it will produce: `Cannot call ->format on string` from inside ->isSameAs. It will be really less easy to debug.

This would need an expect method:
```php
protected function expectDateTime($dt, $message = '')
{
    if (!($dt instanceof DateTime) && !($dt instanceof DateTimeInterface) {
        throw new \InvalidArgument("Expect DateTime object$message");
    }
}
...
public function isSameDay($dt)
{
    $this->expectDateTime($dt, ' in isSameDay first argument');

    return $this->isSameAs('Y-m-d', $dt);
}
```

kylekatarnls:
`($dt instanceof DateTimeInterface)` is not enough because DateTimeInterface does not exist in PHP 5.3 :)
#issuecomment-366886826

briannesbitt:
Also a good time to add some tests for `isSameAs()` method as well ;-)

PR #1134, opened 2018-02-28
Space correction

akanksha1612:
The text alingment was very close a few spaces will be useful. It looks a lot cleaner.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -78,7 +78,7 @@ printf("Now: %s", Carbon::now());
 
 ### Without Composer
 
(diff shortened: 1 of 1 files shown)


Glavic:
I think there is no spaces before `.`, `!` or `?`, at least in my language.
To me, without space looks nicer.

kylekatarnls:
Yes, in English, it's always 1 space after, 0 spaces before. In French `!?;:` take also 1 space before for example.

The documentation of Carbon aims to be written in international English. If you see an inconsistency in the punctuation, please feel free to open other PRs for that.

Thanks.

PR #1137, opened 2018-02-28
Fixes #940 by extending the relative keyword list

### LINKED ISSUE #940: Carbon::setTestNow() has no effect on weekday relative time creation
After setting the current date value with `setTestNow()`, dates created with weekday relative strings yield values based on actual current date instead of test value.

**Example:**
```
<?php
require("vendor/autoload.php");
use Carbon\Carbon;

Carbon::setTestNow("2017-01-01 12:00:00");

$correct = new Carbon("next sunday"); // yields 2017-01-08
$incorrect = new Carbon("sunday"); // yields 2017-04-30, should yield 2017-01-01
```

Tested with version 1.22.

Evidence below:
image

DfKimera:
Fixes #940 by extending the relative keyword list to include bare weekdays; added test coverage for this edge case.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -101,6 +101,7 @@ class Carbon extends DateTime
         'sunday',
         'monday',
         'tuesday',
+        'wednesday',
         'thursday',
         'friday',
         'saturday',
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -98,6 +98,7 @@ class Carbon extends DateTime
         'today',
         'tomorrow',
         'yesterday',
+
         'sunday',
         'monday',
         'tuesday',
@@ -105,6 +106,18 @@ class Carbon extends DateTime
         'thursday',
         'friday',
         'saturday',
+
+        'january',
+        'february',
+        'march',
+        'april',
+        'may',
+        'june',
+        'july',
+        'august',
+        'september',
+        'october',
+        'december',
     );
 
     /**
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -98,6 +98,12 @@ class Carbon extends DateTime
         'today',
         'tomorrow',
         'yesterday',
+        'sunday',
+        'monday',
+        'tuesday',
+        'thursday',
+        'friday',
+        'saturday',
```
u01jmg3: `wednesday`?

DfKimera: Fixed!

kylekatarnls:
Hi, I think we could start by merging this but we would need a finest approach, here are some valid relative and absolute times:

```
first day of January 2008 > absolute
first day of next month > relative
sunday noon > relative
sunday nex week > relative
last sunday of January 2008 > absolute
```

And the unit test with `new Carbon` is good but a more atomic test with `hasRelativeKeywords` with more tricky sentences like this would be good too.

Do you want to complete your PR this way? Else I would just merge it and see later for those improvements.

DfKimera:
@kylekatarnls I've fixed those edge cases of non-relative dates with relative keywords (now testing for years in the time string too). Also added a bunch of test scenarios for these, and now also testing if the `hasRelativeKeywords` is matching correctly. Should be good for a merge now!

DfKimera:
Seems there's an inconsistency in PHP versions regarding the "sunday next week" scenario.
I removed the scenario, as this relates to issues in the underlying `DateTime` and not Carbon.

kylekatarnls:
Thanks! Good job.

PR #1140, opened 2018-03-01
Adding microtime on now call with PHP < 7.1

kylekatarnls:


src/Carbon/Carbon.php
```diff
@@ -343,12 +344,22 @@ public function __construct($time = null, $tz = null)
             $time = $testInstance->format(static::MOCK_DATETIME_FORMAT);
         }
 
+        // Get microseconds from microtime() if "now" asked and PHP < 7.1
+        $timezone = static::safeCreateDateTimeZone($tz);
+        // @codeCoverageIgnoreStart
+        if ($isNow && !isset($testInstance) && version_compare(PHP_VERSION, '7.1.0-dev', '<')) {
```
Glavic: Also add this for version 7.1.3 which always returns 0 of microseconds?

kylekatarnls: No, PHP 7.1.3 works well, see the travis unit tests logs (and my locale PHP 7.1.3 works too), what trouble do you have?

Glavic: Microseconds should be also added to 7.1.3, since it is the only version of 7.1.x that doesn't return microseconds > https://3v4l.org/Re0Ul

kylekatarnls: What's the heck? Give up to understand PHP changes., I will add that.

kylekatarnls: Too bad, it means we will keep this in 2.x just to fix 7.1.3

PR #1151, opened 2018-03-05
Extra helpers

kylekatarnls:
- toArray(): array
- startOfHour(): Carbon Modify to start of current hour, minutes and seconds become 0
- endOfHour(): Carbon Modify to end of current hour, minutes and seconds become 59
- startOfMinute(): Carbon Modify to start of current minute, seconds become 0
- endOfMinute() : Carbon Modify to end of current minute, seconds become 59
- midDay(): Carbon Modify to midday, default to self::$midDayAt

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -155,6 +155,13 @@ class Carbon extends DateTime
     const MOCK_DATETIME_FORMAT = 'Y-m-d H:i:s.u';
 
     /**
+     * Customizable PHP_INT_SIZE override.
+     *
+     * @var int
+     */
+    public static $PHPIntSize = PHP_INT_SIZE;
+
+    /**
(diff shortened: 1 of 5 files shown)


src/Carbon/Carbon.php
```diff
@@ -1618,6 +1668,28 @@ public function toRfc7231String()
             ->format(static::RFC7231_FORMAT);
     }
 
+    /**
+     * Get default array representation
+     *
+     * @return array
+     */
+    public function toArray()
```
Glavic: Should we also add microseconds part?

kylekatarnls: Indeed, it miss. You do the PR?

Glavic: Go ahead :)

PR #1161, opened 2018-03-08
diffForHumans creates the carbon instance if a timestamp is provided

Benjyclay:


kylekatarnls:
Hi, for exception to be compatible everywhere, please the comment syntax:

```php
    /**
     * @expectedException \InvalidArgumentException
     * @expectedExceptionMessage Failed to parse time string (2018-04-13-08:00:00) at position 16
     */
```

And please specify the type of exception rather than the generic \Exception class.

Thanks.

Benjyclay:
Sure,

This exception is thrown when it hits `DateTime::__construct()`. What Exception class would that be?

Here's an example of the error using a different Exception class `Failed asserting that exception of type "Exception" matches expected exception "\InvalidArgumentException".`

kylekatarnls:
Oh I thought this exception was typed, but it's not, so it's right `@expectedException \Exception`

PR #1180, opened 2018-03-10
Fix constructor bug

gitstashgithub:
The bug is if `version_compare(PHP_VERSION, '7.1.3-dev', '>=') && version_compare(PHP_VERSION, '7.1.4-dev', '<')` is true then the whole if statement is true.


kylekatarnls:
Thanks.

PR #1185, opened 2018-03-11
Fix unit-tests (testAddHourPassingArg, testMicroGetterWithDefaultNow) while running on Windows

nthachus:
* Make .php_cs.dist compatible with PHP 5.3
* Fix unit-tests (`testAddHourPassingArg`, `testMicroGetterWithDefaultNow`) while running on Windows
* Correct spelling for some unit-test methods and comments

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -234,6 +234,13 @@ class Carbon extends DateTime
     protected static $utf8 = false;
 
     /**
+     * Add microseconds to now on PHP < 7.1 and 7.1.3. true by default.
+     *
+     * @var bool
+     */
+    protected static $microsecondsFallback = true;
+
+    /**
      * Indicates if months should be calculated with overflow.
      *
      * @var bool
@@ -246,6 +253,28 @@ class Carbon extends DateTime
      * @var bool
      */
     protected static $yearsOverflow = true;
+
+    /**
+     * Add microseconds to now on PHP < 7.1 and 7.1.3 if set to true,
+     * let microseconds to 0 on those PHP versions if false.
+     *
+     * @param bool $microsecondsFallback
+     */
+    public static function useMicrosecondsFallback($microsecondsFallback = true)
+    {
+        static::$microsecondsFallback = $microsecondsFallback;
+    }
+
+    /**
+     * Return true if microseconds fallback on PHP < 7.1 and 7.1.3 is
+     * enabled. false if disabled.
+     *
+     * @return bool
+     */
+    public static function isMicrosecondsFallbackEnabled()
+    {
+        return static::$microsecondsFallback;
+    }
 
     /**
      * Indicates if months should be calculated with overflow.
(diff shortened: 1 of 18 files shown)


src/Carbon/Carbon.php
```diff
@@ -396,11 +396,9 @@ public function __construct($time = null, $tz = null)
             )
         ) {
             $dateTime = new DateTime('now', $timezone);
-            $microTime = microtime(true) * 1000000 % 1000000;
-            if ($microTime > 0) {
-                $microTime = str_pad(strval($microTime), 6, '0', STR_PAD_LEFT);
-                $time = $dateTime->format(static::DEFAULT_TO_STRING_FORMAT).'.'.$microTime;
-            }
+            $microTime = microtime(true);
```
Glavic: Can we just replace this 2 lines with: `$microTime = substr(microtime(false), 2, 6)` ?

src/Carbon/Carbon.php
```diff
@@ -396,11 +396,9 @@ public function __construct($time = null, $tz = null)
...
+            $microTime = str_pad(strval(($microTime - (int) $microTime) * 1000000 % 1000000), 6, '0', STR_PAD_LEFT);
```
Glavic: Can we just replace this 2 lines with code bellow?

```php
$microTime = substr(microtime(), 2, 6);
```

tests/Carbon/AddTest.php
```diff
@@ -183,7 +183,7 @@ public function testAddDayPassingArg()
 
     public function testAddHourPassingArg()
     {
-        $this->assertSame(2, Carbon::createFromTime(0)->addHour(2)->hour);
+        $this->assertSame(2, Carbon::createFromTime(0, 0, 0, 'UTC')->addHour(2)->hour);
```
Glavic: This will do the same, set minutes and seconds to zero:

https://github.com/briannesbitt/Carbon/blob/bba6c6e410c6b4317e37a9474aeaa753808c3875/src/Carbon/Carbon.php#L577

src/Carbon/Carbon.php
kylekatarnls: How could this prevent from having negative values? `-1.25 - -1 = -0.25`, still negative

Glavic: or

```php
$microTime = microtime() * 1e6;
```

yanwenwu:
Thanks

tests/Carbon/AddTest.php
nthachus: No, it's all about the TimeZone. When running this test on Windows, it's failed like:
```php
date_default_timezone_set('Asia/Ho_Chi_Minh');
$d = new \DateTime('2018-03-11 00:00:00', new \DateTimeZone('America/Toronto'));
echo $d->modify('+2 hour')->format('G'); // output: 3
```

Glavic: DST problem, I understand. Nice catch.

nthachus:
I'm not sure, but it's failed when running in PHP 5.3 Windows. I found out
that because of float rounding, so, sometimes, the $microTime is greater
than the $end.
Re-verify it please :-)

kylekatarnls:
I verified and either `/ 1000000` or `* 0.000001` can have the rounding bug, tested with:

```php
    public function testMicroGetterWithDefaultNow()
    {
        $now = Carbon::getTestNow();
        Carbon::setTestNow(null);

        for ($i = 0; $i < 10000; $i++) {
            $start = microtime(true);
            usleep(1000);
            $d = Carbon::now();
            usleep(1000);
            $end = microtime(true);
            $microTime = $d->getTimestamp() + $d->micro * 0.000001;
            $this->assertGreaterThan($start, $microTime);
            $this->assertLessThan($end, $microTime);
        }

        Carbon::setTestNow($now);
    }
```

By running the test 10000 times, it always fail.

Tell me if you find a fix for that.

nthachus:
In computer mathematics, $microTime = $d->getTimestamp() + $d->micro *
0.000001;
is different than 0.000001 * $d->micro + $d->getTimestamp();
So, retry with my code line please.

kylekatarnls:
Sorry, `$d->micro * 0.000001 + $d->getTimestamp()` also fails.

kylekatarnls:
Found the problem about micro, fixed here: https://github.com/briannesbitt/Carbon/pull/1197/files#diff-c060a4f07dfb6bce7015fa79a21b6963L398

nthachus:
It's wonderful :-)

kylekatarnls:
This PR is now a bit too polluted I close it in favor of #1203

PR #1186, opened 2018-03-12
Fix int overflow in 32bit in a better way (credit: Glavic)

kylekatarnls:


src/Carbon/Carbon.php
```diff
@@ -396,11 +396,7 @@ public function __construct($time = null, $tz = null)
             )
         ) {
             $dateTime = new DateTime('now', $timezone);
-            $microTime = microtime(true) * 1000000 % 1000000;
-            if ($microTime > 0) {
-                $microTime = str_pad(strval($microTime), 6, '0', STR_PAD_LEFT);
-                $time = $dateTime->format(static::DEFAULT_TO_STRING_FORMAT).'.'.$microTime;
-            }
+            $time = $dateTime->format(static::DEFAULT_TO_STRING_FORMAT).substr(microtime(), 1, 6);
```
Glavic: If you will use dot from a microtime(), then you should cut string 7 char long: 
```php
.......FORMAT).substr(microtime(), 1, 7);
```

kylekatarnls: Thanks^^ Strange that unit tests passed.

Glavic: `.12345` or `.123456` is both valid :)

Issue #1189, opened 2018-03-12
yesterday() unexpected result in US time zone after Daylight Savings transition


jason-klein:
Our systems use UTC timezone. We frequently need to read the Date value (YYYY-MM-DD) from yesterday($tz), where $tz is a local timezone.

The Daylight Savings transition occurred yesterday (03/11/2018). Today (03/12/2018), our calls to Carbon::yesterday($tz) are returning an unexpected date (03/10/2018 instead of 03/11/2018).

Please review. Is this a bug, or expected behaviour? Should we be reading the Date value from yesterday($tz) differently? 

Our temporary fix is as follows, but this doesn't seem like the correct approach.

```
$yesterday = yesterday('US/Central')->addHours(1);
```

Here is test code.
```
        for($d=11; $d<=13; $d++) {
            for ($h = 0; $h <= 23; $h++) {
                $knownDate = Carbon::create(2018, 3, $d, $h, 00);
                Carbon::setTestNow($knownDate);
                echo 'setTestNow:' . $knownDate . ' UTC'. "\n";

                $yesterday = Carbon::yesterday('US/Central');
                echo 'yesterday:' . $yesterday . ' US/Central' . "\n";
                echo 'yesterday:' . $yesterday->toDateString() . ' US/Central' . "\n";
                echo "\n";
            }
        }
```
Here is the applicable output from test code. The first output and last output are correct. The rest of the outputs are unexpected. The offset is minus 24 hours, instead of minus one day.

```
setTestNow:2018-03-12 04:00:00 UTC
yesterday:2018-03-10 00:00:00 US/Central
yesterday:2018-03-10 US/Central

setTestNow:2018-03-12 05:00:00 UTC
yesterday:2018-03-10 23:00:00 US/Central
yesterday:2018-03-10 US/Central

setTestNow:2018-03-12 06:00:00 UTC
yesterday:2018-03-10 23:00:00 US/Central
yesterday:2018-03-10 US/Central

setTestNow:2018-03-12 07:00:00 UTC
yesterday:2018-03-10 23:00:00 US/Central
yesterday:2018-03-10 US/Central

setTestNow:2018-03-12 08:00:00 UTC
yesterday:2018-03-10 23:00:00 US/Central
yesterday:2018-03-10 US/Central

setTestNow:2018-03-12 09:00:00 UTC
yesterday:2018-03-10 23:00:00 US/Central
yesterday:2018-03-10 US/Central

[...]

Issue #1187, opened 2018-03-12
Microseconds set incorrectly when calling Carbon::today() with PHP 7.0


halloei:
When creating an Carbon instance with `today()`, the milliseconds seem to be set to the current value instead to "000000".

This bug appears with Carbon >=1.24.0 and PHP 7.0.25.

unrealization:
Looking through the code I find this commit to be the culprit: https://github.com/briannesbitt/Carbon/commit/f863708fa1b5c777e80e8dafa0441683dff260ff

The problem lies with the way the date is being built in Carbon's constructor.
While attaching the microseconds to the time string in DateTime's constructor will allow the injection of microseconds into the DateTime object, also in PHP 7, there is no way to manipulate this value after initialization as setTime() only gained the microsecond-parameter in PHP 7.1, and setting the time without that parameter will not touch the microsecond value, unlike in PHP 7.1.

halloei:
Yes, but I cannot upgrade to PHP 7.1 for the moment and Carbon's version 1.24.0 introduced a breaking change (which your PR https://github.com/briannesbitt/Carbon/pull/1190 does not fix).

Example 1:
```
Carbon::now()->startOfDay()->format('Y-m-d H:i:s u');` //  returns '2018-03-12 00:00:00 258223'
```

Example2:
```
$from = Carbon::now()->startOfDay();
$to = Carbon::now()->startOfDay();
$date = Carbon::now()->startOfDay();

$date->between($from, $to); // returns false
```

As I'm using `startOfDay()` and `between()` often, you broke all my code with your "minor" update now.

unrealization:
While `today()` might be easy to fix, the microsecond value being completely immutable affects all methods that use `setTime()`. One of the most obvious candidates for problems is `endOfDay()`, as, inevitably, there will be a gap between the "calculated" and the expected value, which for `endOfDay()` needs to be a microsecond value of 999999.
Imagine checking against entries from a database, to see if an entry has been made today, like this:
`$between = $testDate->between($dayStart, $dayEnd);`
This, in many cases, will be false, depending solely on the microsecond value of `$testDate`.
Here's a bit of an example:
```php
$dayStart = Carbon::parse('today'); //to avoid the current "today() problem"
$dayEnd = Carbon::today()->endOfDay();

$randomMicro = str_pad(strval(mt_rand(0, 999999)), 6, '0');
[...]

PR #1190, opened 2018-03-12
Use the native way to get today/yesterday/tomorrow

kylekatarnls:
Can fix problem like this one: #1187

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -3322,7 +3322,7 @@ class Carbon extends DateTime
      */
     public function startOfDay()
     {
-        return $this->setTime(0, 0, 0);
+        return $this->modify('00:00:00.000000');
     }
 
     /**
@@ -3332,7 +3332,7 @@ class Carbon extends DateTime
      */
     public function endOfDay()
     {
-        return $this->setTime(23, 59, 59);
+        return $this->modify('23.59.59.999999');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


halloei:
Does not fix the issue with `startOfDay()`.

kylekatarnls:
No it has nothing to do with `startOfDay()`. When doing `startOfDay()` on a date with microseconds, microseconds will not be removed, it's not our fault, PHP 7.0 just does not allow to modify microseconds and `startOfDay` must be compatible from 5.3 to 7.2, the best we could do would be to remove microseconds  in startOfDay only for PHP >= 7.1.

Nevertheless, this PR allows to have proper microseconds to 0 when calling `today()`, `yesterday()` and `tomorrow()` and you said you used `today()`.

kylekatarnls:
If you have a better solution, please do not hesitate to propose your own.

halloei:
You're aware of the fact that before PHP 7.1, modifying microseconds is not fully supported. So why did you have to implement Carbon support for microseconds for those PHP versions? Espacially when Carbon 2 is about to be released.
You've released a breaking change as minor version. My solution would be to revert those commits and release them under a new major version.

kylekatarnls:
Sorry, we won't do that, consider replacing a false data (microsecond = 0) by a data not accurate in all cases as a breaking change is questionable (note that we're largely covered by tests and none of documented usages broke after this change), but removing a feature surely is a BC and removing it can bother a lot more people.

I will merge this PR since it fixes #1189.

I also propose to to enable/disable the µs adding with some method like `Carbon::useMicrosecondsFallback()`. It will not be prior on other stuff for 1.25 but if you're interested to do a PR for that, I would merge it.

Thanks for your feedback.

kylekatarnls:
You will be able to disable this feature on 1.25 thanks to #1197

PR #1191, opened 2018-03-12
get the all dates of week

meteguerlek:


src/Carbon/Carbon.php
```diff
@@ -1209,6 +1209,25 @@ public static function getWeekendDays()
...
+     *
+     * @return array
+     */
+    public static function getCurrentWeekDays()
+    {
+        $startOfWeek = Carbon::instance((new static)->startOfWeek());
+
+        $weekDays = [$startOfWeek->toDateTimeString()];
+
+        for ($i = 1; $i < 7; $i++)
```
briannesbitt: Why not just loop from 0 to < DAYS_PER_WEEK, no need to inject the first day of the week with a special case.

src/Carbon/Carbon.php
```diff
@@ -1209,6 +1209,25 @@ public static function getWeekendDays()
...
     }
 
+    /**
+     * Get the all dates of week
+     *
+     * @return array
+     */
+    public static function getCurrentWeekDays()
+    {
+        $startOfWeek = Carbon::instance((new static)->startOfWeek());
```
briannesbitt: If change this to Carbon::now()->startOfWeek()->subDay() you can then just loop from 0 to < DAYS_PER_WEEK and loop code will stay the same.

kylekatarnls:
Hi,

For me this method is too user-specific (mostly when returning datetime strings rather than a date copy users can format as they want after that).

Every feature in Carbon needs unit tests (that's why codecov does not pass).

You also need to follow the code style of the project (check StyleCI details).

It's also required to use constats (Carbon::DAYS_PER_WEEK) rather than hard-coded values, it allows to configure Carbon by extending it and make every method respect the user settings.

I agree with @Glavic for the macro label, this could go in our future list of macros.

meteguerlek:
Hi @kylekatarnls,

its ok, this could go in the future list. But 

```php
$weekDays[] = $startOfWeek->addDay();
```
not working.

kylekatarnls:
`$startOfWeek->addDay()->copy()`

Issue #997, opened 2017-08-22
[FEATURE] CarbonInterval from string


tflori:
I like the format of Jira to define time worked. It's very easy and intuitive.

`1w 3d 4h 32m 23s` will become 1 week, 3 days, 4 hours, 32 minutes and 23 seconds. I know you can use php's `new DateInterval('P10DT4H32M23S')`. One thing: it's not readable and the other you have to device time and date with `T`.

Of course it's not very complicated but a lot of text to have it in every project:
```php
<?php
function interval($interval)
{
  $years = 0;
  $months = 0;
  $days = 0;
  $hours = 0;
  $minutes = 0;
  $seconds = 0;

  foreach (explode(' ', $interval) as $v) {
    switch (substr($v, -1)) {
      case 'y':
        $years += intval($v);
        break;
      case 'M':
        $months += intval($v);
        break;
      case 'w':
        $days += intval($v) * 7;
        break;
      case 'd':
        $days += intval($v);
        break;
      case 'h':
        $hours += intval($v);
        break;
      case 'm':
        $minutes += intval($v);
        break;
      case 's':
        $seconds += intval($v);
        break;
    }
  }

  return CarbonInterval::instance(new DateInterval(
    'P' . // Period
    $years . 'Y' .
    $months . 'M' .
    $days . 'D' . 
    'T' . // Time
    $hours . 'H' .
    $minutes . 'M' .
    $seconds . 'S'
  ));
}

kylekatarnls:
I like it :)

imbrish:
Some time ago I wrote an extension of `Carbon\CarbonInterval` with similar functionality. It comes with a few more methods like cascading the interval (45 days becomes 1 month 15 days), converting it to period or to human readable string in current locale. 

You can take a look at it here https://gist.github.com/imbrish/fb09b3328c498873818b9ff0b56ea423.

If you think it will be an useful addition to the core Carbon i can draft a pull request with the whole ting of some parts of it.

kylekatarnls:
Cascading to months is a rounding maybe not expected. I personally would prefer to get 6 weeks and 3 days for 45 days.

imbrish:
Cascading is optional and can be done just when desired. You're right that what units we cascade to depends on use case/preferences. Adding some flexibility shouldn't be difficult though, it's just matter of an extra argument with some kind of unit whitelist.

[...]

PR #1199, opened 2018-03-14
create CarbonInterval::fromString(string) method

tflori:
Creates a CarbonInterval from string in JIRA format (solves #997)

Format:

Suffix | Unit    | Example | DateInterval expression
------ | ------- | ------- | -----------------------
y    | years   |   1y    | P1Y
M    | months  |   3M    | P3M
w    | weeks   |   2w    | P2W
d    | days    |  28d    | P28D
h    | hours   |   4h    | PT4H
m    | minutes |  12m    | PT12M
s    | seconds |  59s    | PT59S

e. g. `1w 3d 4h 32m 23s` is converted to 10 days 4 hours 32 minutes and 23 seconds.

Special cases:
 - An empty string will return a zero interval
 - Fractions are allowed for weeks, days, hours and minutes and will be converted
   and rounded to the next smaller value (caution: 0.5w = 4d)

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -251,10 +251,10 @@ class CarbonInterval extends DateInterval
         $minutes = 0;
         $seconds = 0;
 
-        $pattern = '/([\d.]+)\h*([a-z]+)/i';
+        $pattern = '/(\d+(?:\.\d+)?)\h*([^\d\h]*)/i';
         preg_match_all($pattern, $intervalDefinition, $parts, PREG_SET_ORDER);
         while ($match = array_shift($parts)) {
-            list(, $value, $unit) = $match;
+            list($part, $value, $unit) = $match;
             $intValue = intval($value);
             $fraction = floatval($value) - $intValue;
             switch (strtolower($unit)) {
@@ -313,7 +313,9 @@ class CarbonInterval extends DateInterval
                     break;
 
                 default:
-                    throw new InvalidArgumentException(sprintf('Invalid unit %s', $unit));
+                    throw new InvalidArgumentException(
+                        sprintf('Invalid part %s in definition %s', $part, $intervalDefinition)
+                    );
             }
         }
 
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -235,6 +235,7 @@ class CarbonInterval extends DateInterval
      *    and rounded to the next smaller value (caution: 0.5w = 4d)
      *
      * @param string $intervalDefinition
+     *
      * @return static
      */
     public static function fromString($intervalDefinition)
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -253,7 +253,7 @@ class CarbonInterval extends DateInterval
 
         $pattern = '/([\d.]+)\h*([a-z]+)/i';
         preg_match_all($pattern, $intervalDefinition, $parts, PREG_SET_ORDER);
-        foreach ($parts as &$match) {
+        while ($match = array_shift($parts)) {
             list(, $value, $unit) = $match;
             $intValue = intval($value);
             $fraction = floatval($value) - $intValue;
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -212,6 +212,98 @@ class CarbonInterval extends DateInterval
     }
 
     /**
+     * Creates a CarbonInterval from string in JIRA format
+     *
+     * Format:
+     *
+     * | Suffix | Unit    | Example | DateInterval expression |
+     * |--------|---------|---------|-------------------------|
+     * |   y    | years   |   1y    | P1Y                     |
+     * |   M    | months  |   3m    | P3M                     |
+     * |   w    | weeks   |   2w    | P2W                     |
+     * |   d    | days    |  28d    | P28D                    |
+     * |   h    | hours   |   4h    | PT4H                    |
+     * |   m    | minutes |  12m    | PT12M                   |
+     * |   s    | seconds |  59s    | PT59S                   |
+     * |--------|---------|---------|-------------------------|
+     *
+     * e. g. `1w 3d 4h 32m 23s` is converted to 10 days 4 hours 32 minutes and 23 seconds.
+     *
+     * Special cases:
+     *  - An empty string will return a zero interval
+     *  - Fractions are allowed for weeks, days, hours and minutes and will be converted
+     *    and rounded to the next smaller value (caution: 0.5w = 4d)
+     *
+     * @param string $intervalDefinition
+     * @return static
+     */
+    public static function fromString($intervalDefinition)
+    {
+        if (empty($intervalDefinition)) {
+            return new static(0);
+        }
+
+        $years = 0;
+        $months = 0;
+        $weeks = 0;
+        $days = 0;
+        $hours = 0;
+        $minutes = 0;
+        $seconds = 0;
+
+        foreach (explode(' ', $intervalDefinition) as $value) {
+            $intValue = intval($value);
+            $floatValue = floatval($value) - $intValue;
+            switch (substr($value, -1)) {
+                case 'y':
+                    $years += $intValue;
+                    break;
+
+                case 'M':
+                    $months += $intValue;
+                    break;
+
+                case 'w':
+                    $weeks += $intValue;
+                    if ($floatValue != 0) {
+                        $days += round($floatValue * Carbon::DAYS_PER_WEEK);
+                    }
+                    break;
+
+                case 'd':
+                    $days += $intValue;
+                    if ($floatValue != 0) {
+                        $hours += round($floatValue * Carbon::HOURS_PER_DAY);
+                    }
+                    break;
+
+                case 'h':
+                    $hours += $intValue;
+                    if ($floatValue != 0) {
+                        $minutes += round($floatValue * Carbon::MINUTES_PER_HOUR);
+                    }
+                    break;
+
+                case 'm':
+                    $minutes += $intValue;
+                    if ($floatValue != 0) {
+                        $seconds += round($floatValue * Carbon::SECONDS_PER_MINUTE);
+                    }
+                    break;
+
+                case 's':
+                    $seconds += $intValue;
+                    break;
+
+                default:
+                    throw new InvalidArgumentException(sprintf('Invalid unit %s', substr($value, -1)));
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


Glavic:
Where can I read more about jira format standard? I'm interested in floated numbers and rounding them up...
I also think method should be named something like `createFromJiraString($jira)` ? We have a lot of `createFrom***()` methods in Carbon.php

src/Carbon/CarbonInterval.php
```diff
@@ -211,6 +211,99 @@ public static function __callStatic($name, $args)
...
 
+    /**
+     * Creates a CarbonInterval from string in JIRA format
+     *
+     * Format:
+     *
+     * | Suffix | Unit    | Example | DateInterval expression |
+     * |--------|---------|---------|-------------------------|
+     * |   y    | years   |   1y    | P1Y                     |
+     * |   M    | months  |   3m    | P3M                     |
```
Glavic: Should example here be 3M ?

imbrish:
Well done!

1. Great addition with parsing floats. Why cutting it short though and not cascading all the way to seconds instead? This could be quickly done by a few changes:
```php
$parts = explode(' ', $intervalDefinition);
foreach ($parts as $value) {
// ...
case 'd':
$days += $intValue;
if ($floatValue != 0) {
     $parts[] = ($floatValue * Carbon::HOURS_PER_DAY) . 'h';
}
break;
```

2. It wouldn't hurt also to add some tests for repeated groups, like `2w 3w`, as it looks like this will be correctly parsed to equivalent of `5w`.

3. As @Glavic mentioned it's not so easy to find more about jira date period format. Maybe we shouldn't mention it altogether? For me biggest motivation in implementing this was shortening code, so maybe simple "parse" would be a good fit for method name, as it has some similarity with `Carbon::parse`.

4. I find it easier to remember about month's M if all parts related to date (years, months, weeks, days) are uppercased. It may again be just personal taste, but what do you think about it?

5. What about multiple spaces in the string, maybe we can be more forgiving in that respect and use `preg_split('/  +/', trim($intervalDefinition))` instead?

6. Should we error on empty string, or maybe return an empty interval instead?

kylekatarnls:
It's maybe not Jira specific, for example GitLab use this format too. But I'm pretty sure units over weeks (M and y are not part of such standard).

About @imbrish remarks:
1. +1, In GitLab 0.5w = 2d 4h (because 1d = 8h and 1w = 5d, all is in working hours)
2. +1
3. Have to find a matching standard not necessarily related to Jira
4. GitLab use `mo w d h m s` (mo = month is the biggest unit) and nothing is in uppercase and command are case insensitive (I like it insensitive so I would prefer "mo" over "M" and no case specific format).
5. +1
6. The if-empty return new static(0) current behavior seems fine to me.

Glavic:
What do you say for this tests? Should they be valid? https://github.com/Glavic/Carbon/commit/3aaaf19bdfaa327929ce0df12b5ec795f0d71c26

kylekatarnls:
New idea: what about a whole editable settings.

Jira (and GitLab in a next version) allow to configure days per weeks, hours per day to accord them to business needs (often used to log work).

Prefix used, ignored characters, case sensitiveness and list of units (and so the maximum unit) could be variables we could set with a method. I try to find an elegant way.

kylekatarnls:
@Glavic it could. In GitLab for example:
`1year 2moonth 3week` : KO
`1y2mo3w4d5h6m7s`: 🆗 
`1 w      2  w       3    w`: 🆗

kylekatarnls:
Possible settings method call:

```php
CarbonInterval::setStringFormat($settings, $caseSensitive = false, $format = 'default');
CarbonInterval::setStringFormat(array(
  CarbonInterval::MONTH_UNIT => array('mo', array(4, CarbonInterval::WEEK_UNIT)),
  CarbonInterval::WEEK_UNIT => array('w', array(5, CarbonInterval::DAY_UNIT)),
  CarbonInterval::DAY_UNIT => array('d', array(8, CarbonInterval::HOUR_UNIT)),
  CarbonInterval::HOUR_UNIT => array('h', array(60, CarbonInterval::MINUTE_UNIT)),
  CarbonInterval::MINUTE_UNIT => array('m'),
));
```

In this example, it would a week has 5 working days of 8 working hours, and years and seconds are ignored.

We could provide default settings and `fromString` could have a second argument that would select the format (to store and use multiple formats).

What do you think of it?

tflori:
1. For sure makes sense
2. Of course
3. That's a real problem. There is no "standard" except ISO 8601 what we don't want. For me it's useless to have more than weeks (in this case you should use two dates). I just tried to enter month in JIRA and it's just returning invalid value for (1M, 1mo, 1month and 1Mo).
4. See number 3 - I'm fine with removing it too. But to have mo and no case sensitivity is also a good idea.
5. Also a good idea
6. I see it the same as @kylekatarnls empty = zero interval

@kylekatarnls In general I like the idea of beeing able to configure it. But it should not get to complicated. Like `Carbon` it should only have some small switches. For example a `protected static $workingHoursPerDay` and `$workingDaysPerWeek` and a flag `const PARSE_WORK_TIME = 1`.

```php
CarbonInterval::setWorkingHoursPerDay(4);
$interval = CarbonInterval::fromString('1.25d'); // 1d 6h
$interval = CarbonInterval::fromString('1.25d', CarbonInterval::PARSE_WORK_TIME); // 1d 1h
```

To allow also names (week(s), day(s) and so on) will not be very complicated - at least if you don't include translations.

tflori:
The growing array is a problem. `each()` is deprecated since 7.2 and `foreach ($array as &$value)` is not working in php5.

kylekatarnls:
`&` is not needed in your case, you can simply do `foreach ($parts as $match) {`

imbrish:
Good point, I guess then we got to fallback to either `for` or `while` to do the looping.

imbrish:
One more idea, what about matching `/^ *-/` for creating an inverted interval?

Glavic:
I would not hardcode format names. Like Kyle suggested, it should be left to the user, so he can specify that as a setting...

tflori:
@imbrish yes, good point. next morning.. :smile:

@Glavic all the names are hard coded when you look at the magic methods `__call()` and `__callStatic()` for example. I'm not sure if this is not an endless story when we now implement something that is a user specific format. what is your proposal to have configuration?

kylekatarnls:
@imbrish nice adding (negative interval). In GitLab `-2w` just do nothing (ignored command), but I guess this is more a business rule. As we should be agnostic of such consideration. So yes, we should support negative intervals.

But here comes a ambiguity:
```
-2d 4h 1w
```
Is this `-` for the whole interval or just for `2d`? Or just for `2d 4h` (following smaller units)?

Glavic:
Just for `2h` :)

Every interval can be +/- ?
```
2d -4h -1w
```

kylekatarnls:
Fair enough.

imbrish:
@Glavic, I think this will overcomplicate the whole thing. We cannot just create an interval of `2d -4h -1w`, we would need to compute that it is actually negative `5d 4h`.

What's the use case for this anyway? I hoped just to be able to write:
```php
$interval = CarbonInterval::fromString('- 2w 5d 6h');
```
Instead of:
```php
$interval = CarbonInterval::fromString('2w 5d 6h');
$interval->invert = true;
```

@tflori, growing an array with `foreach ($array as &$value)` seems to work just fine in PHP 5.4 - 7.2, see https://3v4l.org/InWtK?

tflori:
What do you mean? `-1d 24h = new static(0)` or `(new static(0,0,0,2))->inverse ()`?

tflori:
@imbrish that's strange cause that's the only change in https://github.com/briannesbitt/Carbon/pull/1199/commits/dc5bb0a14732509bf23aebebbf6becea9d90ef60 and before they failed in all 5.x versions: https://travis-ci.org/briannesbitt/Carbon/builds/353698361

kylekatarnls:
OK, negative symbol will give us bugs. Let's have juste positive values for this version.

tflori:
@imbrish `/(\d+(?:\.\d+)?)\h*(\D*)/` is matching too much but not everything. So we will ignore some errors but not all. That's to unclear: `'text in front 2m'` is ok but `'2m text behind'` is not.

I've implemented now `[^\d\h]*` so we are stopping at a break or a new number.

@Glavic @kylekatarnls It's still an open question about configuration. I would rather not implement a configuration for format names but if you insist I can implement it. Otherwise: what about my proposal for configuring the hours per day, days per week and (maybe also) days per month with a flag if the real or the configured values should be used?

imbrish:
#### Parsing

Should we parse intervals within a string or the whole string? As far as I remember `DateTime` fails for unexpected characters in the format. If we choose to be strict, we could throw an exception if there's anything left in the definition after removing all matching parts:

```php
if (trim(preg_replace($pattern, '', $intervalDefinition))) {
    throw new InvalidArgumentException(sprintf('Invalid definition %s', $intervalDefinition));
}
```

#### Configuration

I think constants `DAYS_PER_WEEK`, `HOURS_PER_DAY` etc is the way to go. If users need Carbon to operate on work hours they can extend and configure it separately for the latter purpose. Config and flag would be confusing or error prone.

#### Format names

Whether we should have a configuration of format names depends entirely on what we think the use case is. If you guys think there's big enough demand for parsing user input, then we shouldn't force people to tackle translations themselves and we should instead rely on locale set for Carbon. 

I think we can replace all localized units before going into the loop, which would then handle just the simple `y`, `mo`, `w`, ... We could extract words from current localisations, to have a working solution. I imagine this will get improved later by users of particular locales with all the different forms or abbreviations.

@tflori if you think this is reasonable you can add me to your repo and I'll help you with that.

kylekatarnls:
I think configuration is absolutely needed for this because because we have time trackers as reference for this feature and 7 * 24 * 3600 is not relevant in a working days/hours context. And then if we want to support work duration its obviously variable, depends on the country (48h in India, 40h in USA, 35h in France), on the company.

If you want to log work and have 2 or more subsidiary companies, it will be annoying to change the setting each time you need to calculate a duration rahter than having 2 presets for CompanyA with 5 days of 8 hours and CompanyB with 6 days of 7 hours.

So methods to se individually `DAYS_PER_WEEK` and `HOURS_PER_DAY` seems not so good to me.

Also not letting choose the unit abbreviations and available units make the solution we give to the user very opinionated I think.

But we could merge this and keep this as the default settings and see in a next version how to add settings.

kylekatarnls:
Hi, other 1.25 features are ready to go to the documentation and this method seems stable enough to me. So I propose to merge it as is in 1.25 then add settings layer in 1.26.

@imbrish, @tflori I will open my Carbon GitHub fork to you so, we'll easily can continue this on a common branch.

PR #1201, opened 2018-03-14
Removed dead code.

Glavic:
1. You cannot set days on DateInterval object. > https://3v4l.org/M9lqa
2. Even if it could be set, DateInterval created with DateTime::diff() is not allowed (few lines before).


kylekatarnls:
I plan to refactor completly `instance` to use spec string instead for year, month, etc.

Issue #918, opened 2017-03-23
Allows setDate(DateTime) and similar


rentalhost:
Currently `setDate()` could receives the arguments: `year`, `month?` and `day?`. Should be great if it could receives a instance of `DateTime`.

```php
// Keeps time and replace date.
$datetime = $mydate->setDate(Carbon::today());
```

**It could be applied to following methods:**

* `setDate(DateTime)`: keeps time and replace date;
* `setTime(DateTime)`: keeps date and replace time;

The `setDateTime(DateTime)` should be avoided, because it seems no make sense - maybe if it should replace date and time and keep the timezone, but I still don't guess that it makes sense.

willrowe:
I was just looking for a simple way to copy just the date or just the time from another `DateTime`.
I like this as a possible solution, I would be willing to put together a PR for this if there is any interest.

kylekatarnls:
Yes please @willrowe your PR would be welcome :)

willrowe:
Great, I've been following the new development and was actually planning on submitting a PR for this anyways.

kylekatarnls:
Fixed via #1214

PR #1214, opened 2018-03-18
Add methods to set date or time from another instance

willrowe:
This adds the two new methods `matchDate` and `matchTime`, which sets the `year`, `month`, and `day` or `hour`, `minute`, and `second` respectively, based on the passed datetime.

This is based on discussions in #918. I decided to add these more explicitly named methods instead of overloading the existing `setDate` and `setTime` methods, but I'm happy to make any changes to naming etc., if deemed more appropriate.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1203,6 +1203,22 @@ class Carbon extends DateTime
     }
 
     /**
+     * Set the hour, day, and time for this instance to that of the passed instance.
+     *
+     * @param \Carbon\Carbon|\DateTimeInterface $date
+     *
+     * @return static
+     */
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Separate methods also have advantages but I'm not sure about the naming,I would `expect` `matchXYZ` to return a boolean, a matches count or eventually a match result. Those methods are definitely setters, it should start with `set`. Just my opinion, I'm open to what others will say about.

willrowe:
Maybe something like `setDateFrom` and `setTimeFrom`?

kylekatarnls:
Approved :)

willrowe:
Alright, done.

kylekatarnls:
Thanks!

PR #1237, opened 2018-04-03
Add months of the year because the days of the week already exists

lancepioch:
Perhaps this is unwanted, but I figured I'd give it a go, it'd be nice to have a constant array just like the days of the week


Glavic:
I think we already have that covered in setting locale: http://carbon.nesbot.com/docs/#api-localization

Also I think translations should not be in Carbon class, but in translation files: https://github.com/briannesbitt/Carbon/tree/master/src/Carbon/Lang

PR #1246, opened 2018-04-15
Simpler unit conversions in CarbonInterval

imbrish:
This PR adds two new `CarbonInterval` methods: `cascade()` and `total($unit)`, accompanied by `total<unit>` getters:

```php
echo CarbonInterval::months(76)->forHumans(); // 29 months
echo CarbonInterval::months(76)->cascade()->forHumans(); // 2 years 5 months
```

```php
$interval = CarbonInterval::fromString('3h 30m');
echo $interval->hours * 3600 + $interval->minutes * 60; // 12600

// even more useful when string can contain any units
echo CarbonInterval::fromString('3h 30m')->totalSeconds; // 12600
```

Methods functionality is slightly different, but both are based on the same `static::$cascades` mapping, thus I added both in the same pull request.

Let me know what you think!

I'm also open for suggestions about names, current may be a bit ambiguous.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -88,6 +88,7 @@ class Carbon extends DateTime implements JsonSerializable
     const MONTHS_PER_YEAR = 12;
     const MONTHS_PER_QUARTER = 3;
     const WEEKS_PER_YEAR = 52;
+    const DAYS_PER_MONTH = 30;
     const DAYS_PER_WEEK = 7;
     const HOURS_PER_DAY = 24;
     const MINUTES_PER_HOUR = 60;
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -88,7 +88,7 @@ class Carbon extends DateTime implements JsonSerializable
     const MONTHS_PER_YEAR = 12;
     const MONTHS_PER_QUARTER = 3;
     const WEEKS_PER_YEAR = 52;
-    const DAYS_PER_MONTH = 30;
+    const WEEKS_PER_MONTH = 4;
     const DAYS_PER_WEEK = 7;
     const HOURS_PER_DAY = 24;
     const MINUTES_PER_HOUR = 60;
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -787,7 +787,7 @@ class CarbonInterval extends DateInterval
                 $result += $this->$target * $cumulativeFactor;
             }
             else {
-               $result = ($result + $this->$source) / $factor;
+                $result = ($result + $this->$source) / $factor;
             }
         }
 
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -753,7 +753,7 @@ class CarbonInterval extends DateInterval
     }
 
     /**
-     * Get total amount of given unit.
+     * Get amount of given unit equivalent to the interval.
      *
      * @param string $unit
      *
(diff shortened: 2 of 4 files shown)


src/Carbon/Carbon.php
```diff
@@ -88,6 +88,7 @@ class Carbon extends DateTime implements JsonSerializable
     const MONTHS_PER_YEAR = 12;
     const MONTHS_PER_QUARTER = 3;
     const WEEKS_PER_YEAR = 52;
+    const DAYS_PER_MONTH = 30;
```
Glavic: I really don't like this part :(

kylekatarnls: Indeed I would better imagine a configurable matrix, it's what I started to imagine for business time duration strings:
https://github.com/briannesbitt/Carbon/compare/master...kylekatarnls:factors

Also I think the steps are not obvious.
 Some users could prefer cascading 7 days => 4 weeks => 1 month, other would prefer to cascade directly from days to month.

imbrish:
I was thinking about WEEKS_PER_MONTH instead, but it seemed more natural to have DAYS_PER_MONTH when `CarbonInterval` doesn't natively support weeks, but calculates them dynamically from days (they're always implicitly cascaded in `forHumans` method).

In any case there is currently no implicit way to get from days/weeks to months.

```
DAYS_PER_MONTH = (WEEKS_PER_YEAR / MONTHS_PER_YEAR) * DAYS_PER_WEEK // 30.333...
```

I don't think above is desirable, so we will need either DAYS_PER_MONTH or WEEKS_PER_MONTH.

---

I think `customizeCascades` method could be a good addition, then it should be possible to define:
```php
array(
    'seconds' => array('minutes', Carbon::SECONDS_PER_MINUTE),
    'minutes' => array('hours', Carbon::MINUTES_PER_HOUR),
    'hours' => array('dayz', Carbon::HOURS_PER_DAY),
    'dayz' => array('months', Carbon::DAYS_PER_MONTH),
    'months' => array('years', Carbon::MONTHS_PER_YEAR),
)
```
or let's say:
```php
array(
    // never cascade seconds, hours, minutes
    // dayz to weeks conversion does nothing in fact
    'dayz' => array('weeks', Carbon::DAYS_PER_WEEK),
    // plain old days to months conversion, but by factor DAYS_PER_WEEK * WEEKS_PER_MONTH
    'weeks' => array('months', Carbon::WEEKS_PER_MONTH),
    'months' => array('years', Carbon::MONTHS_PER_YEAR),
)
```

Unfortunately that could lead to unexpected results when counting total amount of unit (in the last example anything below day), so we would need to either:
- validate config when setting it and allow less flexibility,
- validate config when getting total and throw exception if cannot be counted,
- use separate config for both methods (this can snowball fast).

---

Which approach sounds best to you guys?

Failing tests don't seem to be related, but I'm curious to hear if you know what's the reason for them!

kylekatarnls:
This should fix unit tests: https://github.com/briannesbitt/Carbon/commit/a4ee1deb2b7b106b78f6b5ae5fa152e909b84986

I guess, tests was ran at exactly midnight on TravisCI time and as TravisCI sometimes freeze tests for load balancing, it reveals those kind of lacks in unit tests.

So this happens:
```php
$d = Carbon::createFromTimeString('12:20:30', 'Europe/London'); // created at 23:59:59.987, day is 14
$this->assertCarbon($d, Carbon::now()->year, Carbon::now()->month, Carbon::now()->day, 12, 20, 30, 0);
// Carbon::now()->day created at 00:00:123, day is 15
```

imbrish:
@kylekatarnls are you sure it was that random, even though update with c0b1b3a 10 minutes later resulted in same errors?

imbrish:
I thought about customization and tried to code some.

Unfortunately didn't go far. Allowing to customize every aspect of conversion quickly becomes way too abstract. There are just too many cases for that conversion array: changed order, missing or skipped units, confused factors.

All in all I think we should stick to single linear model of conversion like `y > mo > d > h > m > s` and let people change only the factors, the `X_PER_Y` constants.

Said that I updated the days to months conversion to use intermediary days to weeks.

src/Carbon/Carbon.php
```diff
@@ -88,7 +88,7 @@ class Carbon extends DateTime implements JsonSerializable
     const MONTHS_PER_YEAR = 12;
     const MONTHS_PER_QUARTER = 3;
     const WEEKS_PER_YEAR = 52;
-    const DAYS_PER_MONTH = 30;
+    const WEEKS_PER_MONTH = 4;
```
Glavic: Even this is not ok :(

src/Carbon/Carbon.php
```diff
@@ -88,6 +88,7 @@ class Carbon extends DateTime implements JsonSerializable
     const MONTHS_PER_YEAR = 12;
     const MONTHS_PER_QUARTER = 3;
     const WEEKS_PER_YEAR = 52;
```
Glavic: What about leap years?

src/Carbon/Carbon.php
kylekatarnls: If we want do be perfectly strict, 24 hours/day is not exact, 52 weeks/years is also an approximation.

It's OK for me to have approximated default values if they can be set by the user easily.

It's pretty obvious than with an interval of 3 months, getting totalDays or 3 weeks getting totalMonths, we cannot provide an exact value.

src/Carbon/Carbon.php
kylekatarnls: This constant is 4 years old, and I think anyone complained about it. 52 weeks is 364 days, so it's floor value for both leap years and other years.

PR #1247, opened 2018-04-15
Convert CarbonInterval to a DatePeriod

imbrish:
Add `toPeriod` method for creating a new DatePeriod from given CarbonInterval.

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -11,8 +11,8 @@
 
 namespace Carbon;
 
-use DatePeriod;
 use DateInterval;
+use DatePeriod;
 use DateTimeInterface;
 use InvalidArgumentException;
 use Symfony\Component\Translation\TranslatorInterface;
@@ -607,23 +607,23 @@ class CarbonInterval extends DateInterval
     /**
      * Convert the interval to a DatePeriod in given period.
      *
-     * @param  \DateTimeInterface  $start
-     * @param  \DateTimeInterface|int  $end
-     * @param  int  $options
-     * @return \DatePeriod
+     * @param \DateTimeInterface     $start
+     * @param \DateTimeInterface|int $end
+     * @param int                    $options
      *
      * @throws \InvalidArgumentException
+     *
+     * @return \DatePeriod
      */
     public function toPeriod($start, $end, $options = 0)
     {
         if ($this->spec() == 'PT0S') {
-            throw new InvalidArgumentException("Empty interval cannot be converted into period.");
+            throw new InvalidArgumentException('Empty interval cannot be converted into period.');
         }
 
-        if (! ($options & static::EXCLUDE_END) && $end instanceof DateTimeInterface) {
+        if (!($options & static::EXCLUDE_END) && $end instanceof DateTimeInterface) {
             $end = tap(clone $end)->modify('+1 second');
-        }
-        else if ($options & static::EXCLUDE_END && is_int($end)) {
+        } elseif ($options & static::EXCLUDE_END && is_int($end)) {
             $end--;
         }
 
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -607,9 +607,9 @@ class CarbonInterval extends DateInterval
     /**
      * Convert the interval to a DatePeriod in given period.
      *
-     * @param \DateTimeInterface     $start
-     * @param \DateTimeInterface|int $end
-     * @param int                    $options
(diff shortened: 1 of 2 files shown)


kylekatarnls:
There may exist a possible convergence with #1058

It seems to me this could returns a `CarbonPeriod` (sub-class fo `DatePeriod` similar to the idea in #1058) with some magic method like this one: https://github.com/briannesbitt/Carbon/pull/1058#issuecomment-373457499

And cherry on the cake, may be nice if CarbonPeriod could be constructed with only start and end (and would take `CarbonInterval::day()` as  the default interval).

Could be done in other PRs and the rest of the team may have other positions about that.

kylekatarnls:
Indeed, constructor could be kept as is and simply `CarbonPeriod::create()` could have more flexibility in its arguments list.

imbrish:
Native `DatePeriod` ignores `invert` flag, so there's not much I can do about negative intervals. Maybe apart from manually expanding recurrences before creating a new instance, see https://github.com/imbrish/Carbon/commit/35ad77491041e5f455435adae0cebac7477e10ce.

kylekatarnls:
Indeed, the date interval does not matter, but what happens when start > end?

imbrish:
Period will be empty, ie `iterator_to_array($period)` will return empty array.

src/Carbon/CarbonInterval.php
```diff
@@ -596,6 +605,36 @@ public function __toString()
...
+     */
+    public function toPeriod($start, $end, $options = 0)
+    {
+        if ($this->spec() == 'PT0S') {
+            throw new InvalidArgumentException('Empty interval cannot be converted into a period.');
+        }
+
+        if (!($options & static::EXCLUDE_END) && ($end instanceof DateTime || $end instanceof DateTimeInterface)) {
+            $end = clone $end;
+            $end = $end->modify('+1 second');
```
kylekatarnls: modify is not part of the DateTimeInterface. So we can either use `$end instanceof DateTime || $end instanceof DateTimeImmutable` could restrict some custom classes implementing DateTimeInterface and a compatible modify method. Or we can use `method_exists($end, 'modify')`, it's the most flexible but we have no guarantee of what this modify method can be but as DatePeriod constructor take DateTime|DateTimeInterface|int, it should not cause unexpected behavior.

Last option: fallback to `getTimestamp ` method that exists in DateTimeInterface and can allow to create a new object with 1 more second.

imbrish: `DateTimeInterface` can only be implemented by built in classes so changing it to `DateTimeImmutable` will be mostly cosmetic :)

Last option you suggest would be `$end = Carbon::createFromTimestamp($end->getTimestamp(), $end->getTimezone());`, is that right?

kylekatarnls: Yep, changing DateTimeInterface to DateTimeImmutable will be closest semantically, it means we handle both DateTime and DateTimeImmutable, because they both handle modify in known way, but for any future DateTimeInterface classes that can be created in PHP, as modify is not required by this interface, we cannot guarantee it will still work. In this particular case, I would prefer to have `$end instanceof DateTime || $end instanceof DateTimeImmutable`.

I also realize than with EXCLUDE_END, we get a object with `$start` referenced and `$end` copied. I have to do some check to see if it's reliable enough.

imbrish: At least on PHP 7.2.3 start and end instances seemed to be cloned when instantiating `DatePeriod`.

imbrish:
I was thinking to do:
```php
class CarbonPeriod implements Countable, IteratorAggregate
{
    public function getIterator()
    {
        return new IteratorIterator($this->period);
    }
}
```

If you open a pull request I can give you some more feedback. Here is my todo list from when I started to work on `CarbonPeriod`:
- When getting end date revert previously added +1 second.
- Use custom iterator so traversing returns `Carbon` objects.
- Implement filtering via `CallbackFilterIterator`.
- Allow registering user macros (custom filtering, for example `getWeekdays`).

kylekatarnls:
Exactly what we must do!

> When getting end date revert previously added +1 second.
Are you sure it's necessary?

> Use custom iterator so traversing returns Carbon objects.
Check carbonify there: https://github.com/briannesbitt/Carbon/compare/master...kylekatarnls:carbon-period

kylekatarnls:
Here is a proposal, don't hesitate to take parts you need, cherry-pick or if you prefer, I can grant you a write access on my fork. #1257

imbrish:
This can be now closed after #1270 is completed.

PR #1257, opened 2018-04-17
CarbonPeriod prototype

kylekatarnls:


imbrish:
Looks like I can already push to your repo :)
We can work together here, but I have a few questions:
1. What do you think about moving "exclude end" feature from #1247 here?
2. Shall we somehow agree on features first, or just start pushing ideas?

imbrish:
Added some basic filtering. I guess it will work as expected when given start and end dates, but when given recurrences we should probably always return given amount. Now we take x recurrences, and only return those, which are accepted.

Maybe just reinvent the wheel once again, ditch `DatePeriod` altogether, use `DateTime::add` in `Carbon::Interval::next`, and check current against end/recurrences in `valid`?

kylekatarnls:
Saved by symfony/polyfill: https://github.com/symfony/polyfill/blob/master/src/Php54/Resources/stubs/CallbackFilterIterator.php

kylekatarnls:
Now it works but, I don't know why 2018-04-22 is a week day on TravisCI, it passes on my locale machine.

imbrish:
I guess this one should be closed, so we can focus on #1270. Having 5 parallel pull requests for the same feature is a bit confusing.

kylekatarnls:
Right, closed in favor of #1270

PR #1259, opened 2018-04-19
StringsTest::testToLocalizedFormattedDateString should be uppercase Jeudi

ahuggins:


ahuggins:
For reference, the `%A` returns the Day of the week with Title case: http://php.net/manual/en/function.strftime.php

ahuggins:
Weird, I guess it depends on the version of PHP that is being used. Not sure then if this change is needed so much.

kylekatarnls:
It can depends on the locale file installed. Travis is here to ensure tests pass and it test it on all PHP version from 5.3 to 7.2, and as you can see it never causes trouble and neither on locale machines of the devs of our core team. So it's a problem on your machine, you can simply ignore it, or we can consider using `strtolower()` to make this test case insensitive. But we cannot break our Travis tests to them pass on your machine.

Moreover, I'm french and I can ensure you we do not put upper case to week days, in a sentence, the day will be lowercase:

`Je pars jeudi` is right in french.
`Je pars Jeudi` is wrong.

You should try to find why you have this customization in your french locale file. Maybe you have other errors that can impact your internationalization features.

PR #1260, opened 2018-04-20
Multiply CarbonInterval given number of times

imbrish:
With this pull request merged this will be possible (will be useful in `CarbonPeriod`):

```php
$ci = CarbonInterval::create(4, 3, 2, 5, 5, 10, 11); // 4 years 3 months 2 weeks 5 days 5 hours 10 minutes 11 seconds
$ci->times(2.75); // 11 years 8 months 7 weeks 3 days 14 hours 28 minutes 30 seconds
$ci->times(0.333); // 1 year 1 month 6 days 2 hours 3 minutes 4 seconds
$ci->times(-1); // Oops!
```

Interval with negative values seemed to work just fine when added to `DateTime` and stuff. So maybe `forHumans` needs a refresher? Now it generates empty string for such interval.

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -622,6 +622,25 @@ class CarbonInterval extends DateInterval
     }
 
     /**
+     * Multiply current instance given number of times
+     *
+     * @param float $factor
+     *
+     * @return static
+     */
+    public function times($factor)
+    {
+        $this->years = round($this->years * $factor);
+        $this->months = round($this->months * $factor);
+        $this->dayz = round($this->dayz * $factor);
+        $this->hours = round($this->hours * $factor);
+        $this->minutes = round($this->minutes * $factor);
+        $this->seconds = round($this->seconds * $factor);
+
+        return $this;
+    }
+
+    /**
      * Get the interval_spec string of a date interval
      *
      * @param DateInterval $interval
--- a/tests/CarbonInterval/TimesTest.php
+++ b/tests/CarbonInterval/TimesTest.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the Carbon package.
+ *
+ * (c) Brian Nesbitt <brian@nesbot.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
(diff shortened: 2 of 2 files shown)


src/Carbon/CarbonInterval.php
```diff
@@ -621,6 +621,30 @@ public function add(DateInterval $interval)
         return $this;
     }
 
+    /**
+     * Multiply current instance given number of times
+     *
+     * @param float $factor
+     *
+     * @return static
```
kylekatarnls: `@return $this` (static is for same Class but possible different instance)

src/Carbon/CarbonInterval.php
```diff
@@ -621,6 +621,30 @@ public function add(DateInterval $interval)
...
+     */
+    public function times($factor)
+    {
+        if ($factor < 0) {
+            $this->invert = !$this->invert;
```
kylekatarnls: `$this->invert = 1 - $this->invert;`

src/Carbon/CarbonInterval.php
imbrish: Agree, but `add`, `weeksAndDays` and `__call` return `static` and not a single method returns `$this` so it would be inconsistent to change it only here. Wouldn't it?

src/Carbon/CarbonInterval.php
imbrish: These two will be equivalent due to type casting. Is there some other reason than cosmetics? If no then what about stylish `1 ^ $this->invert`? ^^

36864: Is there a reason this should be an integer instead of a boolean?

kylekatarnls: @36864 Yes, DateInterval compatibility.
@imbrish Bitwise operator is OK.

src/Carbon/CarbonInterval.php
kylekatarnls: `@return $this` is used everywhere it's relevant in the Carbon class, if wrong in CarbonInterval, CarbonInterval should align on Carbon.

imbrish: Out of scope for this PR I guess :)

Issue #1248, opened 2018-04-16
Add second rounding methods


spawnia:
I am happily using Carbon as part of a Laravel project.

However, when comparing dates in unit testing, i run into an issue: I want to test the correct serialization of dates to the database. The format i store in the database only stores second precision, so the rehydrated Carbon instance does not contain microseconds.

My testing code looks like this:
````php
$now = now();

$myModel->some_date_field = $now;
$myModel->save();

$this->assertTrue($now->equalsTo($myModel->some_date_field)

var_dump($now);
var_dump($myModel->some_date_field);
````

The assertion fails and the output looks like this:
````
object(Illuminate\Support\Carbon)#3536 (3) {
  ["date"]=>
  string(26) "2017-04-16 11:36:08.769991"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(3) "UTC"
}
object(Illuminate\Support\Carbon)#3547 (3) {
  ["date"]=>
  string(26) "2017-04-16 11:36:08.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(3) "UTC"
}
````

Firstly, i think Carbon should round the timestamp to the next higher second.

Secondly, it would be nice to have a way of comparing the equality of those dates in a way that allows to ignore the precision. I see a few ways for doing that:

- Explicitely round the first date like this: `$carbon->roundSeconds()`
- Pass an extra parameters to the equality check, specifying precision: `$carbon->equalsTo($carbon2, Carbon::SECOND)`
- Have a seperate function for comparing equality with different precision `$carbon->equalsSecondsTo($carbon2)`

Which of those would be the most useful? I would happily submit a pull request, but would like to discuss this before getting to work.

I currently worked around it by calling `$carbon->second($carbon->second)` which effectively rounded the date instance.

kylekatarnls:
Hi,

> Firstly, i think Carbon should round the timestamp to the next higher second.

Why? I store microseconds in my DB in my app because I want a precise order of what happened, and I certainly not want to loose this information when comparing two dates. Carbon keep the maximum of precision the PHP version allow to and it's up to the user to do rounding he needs. The library should certainly not take such opinionated decision.

> Secondly, it would be nice to have a way of comparing the equality of those dates in a way that allows to ignore the precision.

There, is you can compare with any format you need:
```php
$date1->isSameAs('U', $date2) // second precision
$date1->isSameAs('Y-m-d H:i', $date2) // minute precision
[...]

PR #1261, opened 2018-04-21
Add methods for granular comparison (mentioned in #1248)

spawnia:
This adds 6 methods for more fine-granular comparison of hours, minutes and seconds.

- isSameHour / isCurrentHour
- isSameMinute / isCurrentMinute
- isSameSecond / isCurrentSecond

Tests are in place and ran successfully, i also unified the PHPDocs for related methods and added a comment mentioning planned changes to align the behaviour of `isSameMonth()` with the other methods.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2329,10 +2329,10 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Checks if the passed in date is in the same month as the instance´s month (and year if needed).
+     * Checks if the passed in date is in the same month as the instance´s month.
      *
-     * The second parameter $ofSameYear will be changed to default to true in 2.x to be consistent with
-     * the other comparison methods isSameDay, isSameHour. See Issue #1248
+     * Note that this defaults to only comparing the month while ignoring the year.
+     * To test if it is the same exact month of the same year, pass in true as the second parameter.
      *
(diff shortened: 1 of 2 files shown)


src/Carbon/Carbon.php
```diff
@@ -2329,9 +2329,12 @@ public function isCurrentMonth()
     }
 
     /**
-     * Checks if the passed in date is in the same month as the instance month (and year if needed).
+     * Checks if the passed in date is in the same month as the instance´s month (and year if needed).
      *
-     * @param \Carbon\Carbon|\DateTimeInterface|null $date       The instance to compare with or null to use current day.
+     * The second parameter $ofSameYear will be changed to default to true in 2.x to be consistent with
```
kylekatarnls: This is not an absolute certitude, it should be discussed further. Maybe we will rather split it into 2 distinct methods to better follow the principle of single responsibility. Or maybe it will be changed into a other way (native comparison, static method... PHPDoc comments does not need to detail future behavior. It should rather focus on current behavior.

spawnia:
You made a good point about not polluting the PHPDoc with future discussions. I updated it to describe the current behaviour in a bit more detail.

Your idea of adding different methods for fuzzy comparison sounds great. How about naming them `isEqualX`? This is similar to what PHPUnit does, they have `assertSame` and `assertEquals` which differ in strictness. I would hold off on implementing them before we can change `isSameMonth`, can we add such a proposal to a 2.0 milestone?

PR #1268, opened 2018-04-24
Allow to create from multiple formats

kylekatarnls:
This allow to create a Carbon instance from multiple possible formats. For example, if you have both date with and without microseconds, you can first try to get microsecond, then if the format does not match fallback to format without microseconds (see `testInstanceFromMultipleFormats`).

This can also allow some user multiple input such as: `array('Y-m-d', 'm/d/Y', 'n/j/Y', 'm/d/y')`

An alias `createFromFormats` allow to call with both singular and plural form.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -808,7 +808,7 @@ class Carbon extends DateTime implements JsonSerializable
     /**
      * Create a Carbon instance from a specific format or a format list.
      *
-     * @param string|array              $formats Datetime format or formats array to try
+     * @param string|array              $format Datetime format or formats array to try
      * @param string                    $time
      * @param \DateTimeZone|string|null $tz
      *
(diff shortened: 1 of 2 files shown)


36864:
I like the general idea behind what you're doing, but there's a lot of potential for misuse here.

If you want a function to allow creating dates with optional microseconds, make a function that explicitly makes microseconds optional on any input format string.

This approach would allow a dev to do do something like `createFromFormats(array('m/d/y', 'd/m/y', 'y/m/d')`, which could be interpreted as "any of these formats are valid", meaning someone might expect a `y/m/d` date to be perfectly valid to use here and be unpleasantly surprised when they get erratic behavior.
For example, the following snippet would bounce between formats based on the year and month that is retrieved (years 01-12 would use the first format, years 13-31 would use the second format, and only years after 32 would use the correct format).
```
> $date = get_formatted_date('y/m/d');
> Carbon::createFromFormats('m/d/y', 'd/m/y', 'y/m/d', $date);
```

While it does look silly that a dev would write this code, it's not too much of a stretch to think of the second line as an abstraction from a library that uses carbon.

```
> $date = LibDateHelper::createCarbonDate($date_string);
```

kylekatarnls:
@36864 I went in this limitation using the `protected $dateFormat;` of Laravel, so the point is to not have to change the method existing call (because you don't always control it if it's handled by a framework/third-party library).

I would say, the user can yet loop on `array('m/d/y', 'd/m/y', 'y/m/d')` and try-catch it. The ambiguity (mostly for `d/m` vs `m/d` which are both common depending on the user country) exist with or without this change, there is nothing the developer can do to prevent the user to use one format for an other except a explicit warning or replacing the input with a calendar ui control.

Maybe we could use our `$regexFormats` formats list to check collisions but it could be a bit heavy.

kylekatarnls:
After tests, this could cause regression, the case is bit exotic but an object implementing both __toString and ArrayAccess used as $format could be previously understand as 1 single format and could now iterate over the inner iterator's entries. So `createFromFormat` should rather be kept untouched and `createFromFormats` to be a different method calling createFromFormat for each format. #914 could ease to do that, so I prefer to wait for the version 2.

PR #1269, opened 2018-04-24
Allow to create from multiple formats

kylekatarnls:
This allow to create a Carbon instance from multiple possible formats. For example, if you have both date with and without microseconds, you can first try to get microsecond, then if the format does not match fallback to format without microseconds (see `testInstanceFromMultipleFormats`).

This can also allow some user multiple input such as: `array('Y-m-d', 'm/d/Y', 'n/j/Y', 'm/d/y')`

An alias `createFromFormats` allow to call with both singular and plural form.


kylekatarnls:
Merged on a temporary branch (`multiple-formats`) to make some tests

PR #1270, opened 2018-04-26
Carbon period from scratch

kylekatarnls:
Alternative of #1257 with no relying on DatePeriod (to complete)

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -935,6 +935,32 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
+     * Make a Carbon instance from given variable if possible.
+     *
+     * Always return a new instance. Parse only strings and only these likely to be dates (skip intervals
+     * and recurrences). Throw an exception for invalid format, but otherwise return null.
+     *
+     * @param mixed $var
+     *
+     * @return static|null
+     */
+    public static function make($var)
+    {
+        if ($var instanceof DateTime || $var instanceof DateTimeInterface) {
+            return static::instance($var);
+        }
+
+        if (is_string($var)) {
+            $var = trim($var);
+            $first = substr($var, 0, 1);
+
+            if (is_string($var) && $first !== 'P' && $first !== 'R' && preg_match('/[a-z0-9]/i', $var)) {
+                return static::parse($var);
+            }
+        }
+    }
+
+    /**
      * Get a copy of the instance.
      *
      * @return static
@@ -2118,7 +2144,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is a weekday
+     * Determines if the instance is a weekday.
      *
      * @return bool
      */
@@ -2128,7 +2154,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is a weekend day
+     * Determines if the instance is a weekend day.
      *
      * @return bool
      */
@@ -2138,7 +2164,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is yesterday
+     * Determines if the instance is yesterday.
      *
      * @return bool
      */
@@ -2148,7 +2174,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is today
+     * Determines if the instance is today.
      *
      * @return bool
      */
@@ -2158,7 +2184,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is tomorrow
+     * Determines if the instance is tomorrow.
      *
      * @return bool
      */
@@ -2168,7 +2194,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the next week
+     * Determines if the instance is within the next week.
      *
      * @return bool
      */
@@ -2178,7 +2204,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the last week
+     * Determines if the instance is within the last week.
      *
      * @return bool
      */
@@ -2188,7 +2214,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the next quarter
+     * Determines if the instance is within the next quarter.
      *
      * @return bool
      */
@@ -2198,7 +2224,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the last quarter
+     * Determines if the instance is within the last quarter.
      *
      * @return bool
      */
@@ -2208,7 +2234,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the next month
+     * Determines if the instance is within the next month.
      *
      * @return bool
      */
@@ -2218,7 +2244,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the last month
+     * Determines if the instance is within the last month.
      *
      * @return bool
      */
@@ -2228,7 +2254,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within next year
+     * Determines if the instance is within next year.
      *
      * @return bool
      */
@@ -2238,7 +2264,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is within the previous year
+     * Determines if the instance is within the previous year.
      *
      * @return bool
      */
@@ -2248,7 +2274,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is in the future, ie. greater (after) than now
+     * Determines if the instance is in the future, ie. greater (after) than now.
      *
      * @return bool
      */
@@ -2258,7 +2284,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is in the past, ie. less (before) than now
+     * Determines if the instance is in the past, ie. less (before) than now.
      *
      * @return bool
      */
@@ -2268,7 +2294,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is a leap year
+     * Determines if the instance is a leap year.
      *
      * @return bool
      */
@@ -2309,7 +2335,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is in the current year
+     * Determines if the instance is in the current year.
      *
      * @return bool
      */
@@ -2331,7 +2357,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is in the current month
+     * Determines if the instance is in the current month.
      *
      * @return bool
      */
@@ -2358,7 +2384,7 @@ class Carbon extends DateTime implements JsonSerializable
     }
 
     /**
-     * Determines if the instance is in the current month
+     * Determines if the instance is in the current month.
      *
      * @return bool
      */
@@ -2573,6 +2599,54 @@ class Carbon extends DateTime implements JsonSerializable
     public function isLastOfMonth()
     {
         return $this->day === $this->daysInMonth;
+    }
+
+    /**
+     * Check if the instance is start of day / midnight.
+     *
+     * @param bool $checkMicroseconds
+     *
+     * @return bool
+     */
+    public function isStartOfDay($checkMicroseconds = false)
+    {
+        return $checkMicroseconds
+            ? $this->format('H:i:s.u') === '00:00:00.000000'
+            : $this->format('H:i:s') === '00:00:00';
+    }
+
+    /**
+     * Check if the instance is end of day.
+     *
+     * @param bool $checkMicroseconds
+     *
+     * @return bool
+     */
+    public function isEndOfDay($checkMicroseconds = false)
+    {
+        return $checkMicroseconds
+            ? $this->format('H:i:s.u') === '23:59:59.999999'
+            : $this->format('H:i:s') === '23:59:59';
+    }
+
+    /**
+     * Check if the instance is start of day / midnight.
+     *
+     * @return bool
+     */
+    public function isMidnight()
+    {
+        return $this->isStartOfDay();
+    }
+
+    /**
+     * Check if the instance is midday.
+     *
+     * @return bool
+     */
+    public function isMidday()
+    {
+        return $this->format('G:i:s') === static::$midDayAt.':00:00';
     }
 
     /**
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -2604,7 +2604,8 @@ class Carbon extends DateTime implements JsonSerializable
     /**
      * Check if the instance is start of day / midnight.
      *
-     * @param bool $checkMicroseconds
+     * @param bool $checkMicroseconds check time at microseconds precision
+     *                                /!\ Warning, this is not reliable with PHP < 7.1.4
      *
      * @return bool
      */
@@ -2618,7 +2619,8 @@ class Carbon extends DateTime implements JsonSerializable
     /**
      * Check if the instance is end of day.
      *
-     * @param bool $checkMicroseconds
+     * @param bool $checkMicroseconds check time at microseconds precision
+     *                                /!\ Warning, this is not reliable with PHP < 7.1.4
      *
      * @return bool
      */
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -4448,8 +4448,6 @@ class Carbon extends DateTime implements JsonSerializable
      *
      * @param object $mixin
      *
-     * @throws \ReflectionException
-     *
      * @return void
      */
     public static function mixin($mixin)
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -355,7 +355,7 @@ class CarbonInterval extends DateInterval
     }
 
     /**
-     * Creates a CarbonInterval from string
+     * Creates a CarbonInterval from string.
      *
      * Format:
      *
@@ -501,7 +501,7 @@ class CarbonInterval extends DateInterval
     }
 
     /**
-     * Get the translator instance in use
+     * Get the translator instance in use.
      *
      * @return \Symfony\Component\Translation\TranslatorInterface
      */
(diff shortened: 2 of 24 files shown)


imbrish:
Ps. I didn't pay much attention to the ISO, so `toIso8601String` possibly produces malformed specification.

kylekatarnls:
Style and tests green again, now we need to cover all missing lines:
https://codecov.io/gh/briannesbitt/Carbon/src/carbon-period-from-scratch/src/Carbon/CarbonPeriod.php

kylekatarnls:
Note: I prefer to forbid negative recurrences, it would avoid `->spec()` to potentially provide strings that does not comply ISO-8601.

imbrish:
ISO 8601 may be partial (not implemented)

> For start/end expressions, if any elements are missing from the end value, they are assumed to be the same as for the start value including the time zone. This feature of the standard allows for concise representations of time intervals. For example, the date of a two-hour meeting including the start and finish times could be simply shown as "2007-12-14T13:30/15:30", where "/15:30" implies "/2007-12-14T15:30" (the same date as the start), or the beginning and end dates of a monthly billing period as "2008-02-15/03-14", where "/03-14" implies "/2008-03-14" (the same year as the start).

I guess that would justify a separate parser method instead of current naive "explode and pretend it wasn't ISO" implementation.

kylekatarnls:
`"2007-12-14T13:30/15:30"` can be done via:
```php
$startDate = Carbon::create('2007-12-14T13:30');
$endDate = $startDate->copy()->modify('15:30');
```
Maybe we can always create endDate from startDate, just compare to DatePeriod results and maybe, tricky things like: `2007-12-14T13:30/2007-12-15` or `2007-12-14T13:30:00.123456/13:30`.

imbrish:
New features:
- Backward iteration (inverted interval and end date before start date).
- Static create methods: `createFromArray`, `createFromISO`, `createBetween`.
- Static and instance aliases:
    ```php
    CarbonPeriod::days(3)->hours(5)->invert()
        ->sinceNow()->until('2010-01-10')
        ->filter(...);
    ```
- Support for changing parameters during iteration.
- Caching of current date validation result.
- Added `CarbonInterval` methods: `toPeriod`, `invert`.
- Reorganized and added more tests.

Discarded:
- Callback instead of interval — Just use `CallbackFilterIterator`.
- Automatically rewind after foreach — State before/during/after iterations should be consistent.

Remaining:
- Add `toArray`, `first`, `last`, `count` methods and implement `Countable` interface.
- Cache array of dates until rewind.
- Add macros (base tests on https://github.com/briannesbitt/Carbon/pull/1058).

imbrish:
Updates:
- Implement `toArray`, `first`, `last` and `count` methods (at almost no cost in performance!)
- Cache results of iteration until parameters are changed, `reset` method is called or `DISABLE_RESULTS_CACHE` option is toggled on. Reuse cached results in subsequent iterations.
- Support macros and mixins which take precedence over built-in aliases.

imbrish:
I'm a bit unsure if that `__call` magic for applying filters from `Carbon::is*` methods should be kept. It requires adding new methods to the switch and bloats the class comment even more. This could be fixed with a regex and some script for auto-generating the comment, but that seems like too much.

Alternative would be to assume all filters which do not pass `is_callable` test to be Carbon methods:

```php
CarbonPeriod::create('2017-12-01', '2017-12-31')->addFilter('isWeekday');
```

But that on the other hand would require two separate signatures of `addFilter`, so that using `is*` methods with arguments could work.

kylekatarnls:
Hum, this test pass on my locale PHP 7.0: https://travis-ci.org/briannesbitt/Carbon/jobs/378780071
Strange...

imbrish:
Which PHP 7.0.x version do you use? I tested it on https://3v4l.org and checked for PHP bugs. It was consistent behavior on PHP 7. Fix could be moved to CarbonInterval though.

kylekatarnls:
In fact I had a lot of review comments, but as CarbonPeriod.php is too big, they were hidden.

kylekatarnls:
PHP 7.0.10 windows 64b nts:
But this works on all versions: https://3v4l.org/mRgD2

imbrish:
You need to define magic `__get` method to trigger this bug, it's not about DateInterval.

kylekatarnls:
`isset` + `__isset` seems to work.

imbrish:
No wait, it was `__get` for CarbonInterval because it throws exception for `f`.
Your example works because PHP7 is not tested at all (check eol versions and compare 7.0.14 - 7.0.21 https://3v4l.org/mRgD2#v7014).

imbrish:
Yeah, just `isset` will fail.

I haven't seen your code reviews before. Could you check them and update these that are still relevant? It's hard to see what's going on with so many changes since.

kylekatarnls:
A fallback for native properties in the getter seems needed for PHP 7.0: https://3v4l.org/NOoYu

kylekatarnls:
PS: In fact, the bug only happen for PHP > 7.0.14 < 7.0.21, that's why, the latest 7.0.22 - 7.2.4 do not bug and why my locale old 7.0.10 is fine too.

kylekatarnls:
Ugly, but this works with or whitout __get and with or whitout __isset, so it handle DateInterval, CarbonInterval or any sub-class that can implement `f` or not: https://3v4l.org/G45mE

imbrish:
I think this workaround needs a comment, so someone won't refactor it into oblivion later.

src/Carbon/CarbonPeriod.php
```diff
@@ -0,0 +1,1724 @@
...
+
+        return implode('/', $parts);
+    }
+
+    /**
+     * Convert the date period into a string.
+     *
+     * @return string
+     */
+    public function toString()
```
kylekatarnls: @imbrish This should be language-agnostic or localized but not in English.

src/Carbon/CarbonPeriod.php
```diff
@@ -0,0 +1,1724 @@
...
+        $current = $this->prepareForReturn($this->current);
+
+        foreach ($this->filters as $tuple) {
+            list($filter) = $tuple;
+
+            $result = call_user_func($filter, $current->copy(), $this->key, $this);
+
+            if ($result === static::END_ITERATION) {
+                return static::END_ITERATION;
+            } elseif (!$result) {
```
kylekatarnls: @imbrish else and return are redundant.

imbrish:
@kylekatarnls do you mean moving `makeCarbon` and `makeCarbonInterval` to their respective classes?

I don't think that will work, because these methods need to be very specific:
1. Return a new instance even if Carbon is passed.
2. Parse string only if it looks appropriate.
3. Otherwise return null.

Meanwhile it would be preferable for native methods to parse anything including numbers (timestamps) and incorrect formats, throw exceptions on failure and return now on empty input.

kylekatarnls:
I simply mean move the function as is:
```php
class Carbon
{
  public function make($date)
  {
    if ($date instanceof DateTime || $date instanceof DateTimeInterface) {
      return Carbon::instance($date);
    }
    if (is_string($date) && !is_numeric($date)) {
      return new Carbon($date);
    }
  }
}
class CarbonInterval
{
  public function make($date)
  {
    if ($date instanceof DateTime || $date instanceof DateTimeInterface) {
      return CarbonInterval::instance($date);
    }
    if (is_string($var) && substr($var, 0, 1) === 'P') {
      return new CarbonInterval($date);
    }
  }
}
class CarbonPeriod
{
  ...
  if ($parsed = CarbonInterval::make($argument)) {
    $interval = $parsed;
  } elseif ($start === null && $parsed = Carbon::make($argument)) {
    $start = $parsed;
  } elseif ($end === null && $parsed = Carbon::make($argument)) {
    $end = $parsed;
  }
  ...
}
```

src/Carbon/CarbonPeriod.php
```diff
@@ -420,6 +420,9 @@ protected static function makeCarbonInterval($var)
      */
     protected static function intervalHasTime(DateInterval $interval)
     {
+        // array_key_exists + get_object_vars check actual microsecond support as public available property
+        // isset or property_exists will both fail on PHP 7.0.14 - 7.0.21 due to the following bug:
+        // https://bugs.php.net/bug.php?id=74852
```
imbrish: Same bug seems to happen as well on 7.1.0 - 7.1.7 but it is not apparent because `f` property exists - https://3v4l.org/tqJaM#v710.

kylekatarnls: Yep it's an other one: https://bugs.php.net/bug.php?id=74852
But the array_key_exists will not be impacted by this one.

imbrish:
Fixed:
- Typos, code style and comments.
- Filters for Carbon's `is*` methods can now be added via `addFilter` (no maintenance).
- Added support for simple `toString` translations and tweaked output slightly.
- Moved `make*` methods to their respective classes.

Added:
- New Carbon methods: `isStartOfDay`, `isMidnight`, `isEndOfDay` and `isMidday`.
- Added tests for non-closure filters.
- Support setting DateInterval by `fromString` format (`3d 4h 32m` etc).

src/Carbon/CarbonPeriod.php
```diff
@@ -0,0 +1,1623 @@
...
+        $parts = array();
+
+        if ($this->startDate && !$this->startDate->isStartOfDay() || $this->endDate && !$this->endDate->isStartOfDay()) {
+            $format = 'Y-m-d H:i:s';
+        } else {
+            $format = 'Y-m-d';
+        }
+
+        if ($this->recurrences !== null) {
+            $parts[] = $translator->transChoice('period_recurrences', $this->recurrences, array(':count' => $this->recurrences));
```
kylekatarnls: It's ambiguous, recurrences are occurrences + 1. Can be kept as is for now, but In fact, it would be something more like:
occurrences = recurrences + (exclude_start ? 1 : 0)
- occurrences times
- exclude_start ? from : after
- exclude_end ? to : until

But it can make things more confused.

src/Carbon/CarbonPeriod.php
```diff
@@ -0,0 +1,1623 @@
...
+    {
+        $translator = Carbon::getTranslator();
+
+        $parts = array();
+
+        if ($this->startDate && !$this->startDate->isStartOfDay() || $this->endDate && !$this->endDate->isStartOfDay()) {
+            $format = 'Y-m-d H:i:s';
+        } else {
+            $format = 'Y-m-d';
+        }
```
kylekatarnls: Ternary for just the hour part would be cleaner.

src/Carbon/Carbon.php
```diff
@@ -933,6 +933,27 @@ public static function createFromTimestampUTC($timestamp)
...
+     *
+     * @return static|null
+     */
+    public static function make($var)
+    {
+        if ($var instanceof DateTime || $var instanceof DateTimeInterface) {
+            return static::instance($var);
+        }
+
+        if (is_string($var) && preg_match('/[:-]/', $var)) {
```
kylekatarnls: It drops the support of `"first monday of january"` strings kind and RegExp is a bit more heavy.

src/Carbon/Carbon.php
```diff
@@ -2574,6 +2574,46 @@ public function isLastOfMonth()
...
+    }
+
+    /**
+     * Check if the instance is end of day.
+     *
+     * @return bool
+     */
+    public function isEndOfDay()
+    {
+        return $this->format('H:i:s.u') === $this->copy()->endOfDay()->format('H:i:s.u');
```
kylekatarnls: Could fail with object like this in PHP 7.0: `Carbon::create('2000-01-01 20:34:12.123456')->endOfDay()`, it's explicitly end of day. Then copy operation for `isStartOfDay` and `isEndOfDay` seems heavy. I think we can be more flexible by ignoring microseconds part and could simply be: `$this->format('H:i:s') === '00:00:00'` and `$this->format('H:i:s') === '23.59.59'`

imbrish: I was comparing format with string originally, but decided to rely on startOfDay/endOfDay to keep the "is" check always consistent with these methods. Could be overkill, but PHP got so much faster I rarely think about these micro-optimizations anymore.

Date only DateTime (like for MySQL date column) will have zeroed time part, microseconds included. Thus in my opinion dates with microseconds should not be considered date only. But I guess this depends on use case and could serve us better as separate `isDateOnly`, while we can discuss how to deal with microseconds in isStartOfDay/isEndOfDay separately. Should we cut them off or round them?

imbrish:
Replaced `preg_match` with `substr`/`strpos`. I think a simple regex makes the code easier to grasp than multiple conditions and outweighs the performance gains, but it's not a big deal.

src/Carbon/Carbon.php
imbrish: It'll be tricky to determine if input is of the `first monday of january` kind or rather `3 days 4 hours` without `try ... catch` block.

But as these methods are now externalized only generic `Invalid ISO 8601 specification`/`Invalid constructor parameters` could be then thrown for invalid input. Not very useful.

Even more difficult as `fromString` will not fail for `first monday of january`.

src/Carbon/CarbonPeriod.php
imbrish: Not a big fan of ternaries with long conditions here: `$format = 'Y-m-d'.($this->startDate && !$this->startDate->isStartOfDay() || $this->endDate && !$this->endDate->isStartOfDay() ? ' H:i:s' : '')` doesn't look very clean. I was thinking to make into a variable first or a separate method, but... why avoid `if .. else` so much?

src/Carbon/CarbonPeriod.php
imbrish: It for sure confused me :) If you want to change it I'd advocate keeping the behavior and renaming `recurrences` to `limit` or something like that.

src/Carbon/CarbonPeriod.php
kylekatarnls: I just never use `else` in my own project. Check https://github.com/phug-php, it's a very complex project, a lot more code than in Carbon, and it contains no single `else` statement. I did not invented this rule it's a PHPMD wide-used rule: https://phpmd.org/rules/cleancode.html#elseexpression

The meaning of this is: a `else` statement create a block when you can add treatment, it opens the door too complexity. If it's really needed, it deserves a dedicated method. If it cearly match the ternary pattern (like here), I mean: an assignation to the same variable in `if` and `else`, then ternary has the advantage to just be an operator like `+`, `-`, etc. it's a simple operation, not block, statement and stuff that more likely will multiply paths.

But if the ternary is long, there is no particular reason to write it on one line:
```php
$format = 'Y-m-d'.(
  $this->startDate && !$this->startDate->isStartOfDay() || $this->endDate && !$this->endDate->isStartOfDay()
    ? ' H:i:s'
    : ''
);
```

src/Carbon/CarbonPeriod.php
kylekatarnls: What I say is: start monday 3 recurrences of 1 day will give `3 times every 1 day from monday` when the CarbonPeriod iterate 4 times. That's why the `"times"` word may not be the good one or else the way to count.

src/Carbon/Carbon.php
kylekatarnls: OK to drop, if needed users would just have to use `Carbon::parse()` explicitly.

src/Carbon/Carbon.php
kylekatarnls: False, I configure my MySQL DateTime to (6) precision, that's means microseconds are stored and returned and it's a best practice. And microseconds zeroed is good for your isStartOfDay but will be always wrong for your isEndOfDay so it's more likely an argument in favor of a floor comparison. The fact is it will not be consistent in PHP 7.0: event `10:10:10`->endOfDay will give: `10:10:10.999999` so you will never get isEndOfDay truthy in PHP 7.0 if the date hadn't be created originally with `.999999` microseconds.

You may also consider this rule is not relevant: `$this->startDate && !$this->startDate->isStartOfDay() || $this->endDate && !$this->endDate->isStartOfDay()` because you can fall on start/end of day by accident. It may be better to always display hour part, then have other `toString` method, like `toDateRangeString`, then think about new `is*` methods in a dedicated issue.

src/Carbon/CarbonPeriod.php
imbrish: That reflects the actual behavior - 3 recurrences in CarbonPeriod will return 3 results. What confused me was DatePeriod which would give me 3 results for 3 recurrences with EXCLUDE_START but 4 results without EXCLUDE_START.

Thus I think about renaming the feature. Particularly that recurrences can be mixed with end date and other filters (for example 01-10 → 01-11 despite 5 recurrences gives 2 results). So "limit" would be more accurate.

Could be made also into separate features to keep the standard behavior of recurrences, but provide limit functionality.

src/Carbon/Carbon.php
imbrish: MySQL `DATE` not `DATETIME`. Latter obviously contains time.

I started to think about `2018-05-16 00:00:00.000000` as equivalent to `2018-05-16` because of how MySQL treats date-only datetime - as precisely midnight on that day. If you insert `2018-05-16` it will be expanded to `2018-05-16 00:00:00.000000` if you select where `x < 2018-05-16` you'll get only days prior to given. So it doesn't matter if start and end are both midnight by accident - cutting of `0`s is just an equivalent notation, but more compact and easier for humans.

I get that MySQL is not a full picture, I just didn't think about `toString` representation as needing to be so precise. Just like Carbon's `toString` ignores microseconds. It will never be precise, because we cannot know what the filters do. I was thinking about it more as a debug method "uh oh what does this one contain? let's echo it.". Probably should've made it clear before adding translations.

I don't get your PHP 7 example. How come endOfDay will give: 10:10:10.999999? Or do you mean if I do `Carbon::parse('... 23:59:59')` it won't be endOfDay? Well same will be true for `Carbon::parse('... 23:59')`...

src/Carbon/CarbonInterval.php
```diff
@@ -375,11 +375,11 @@ public static function make($var)
         if (is_string($var) && !empty($var)) {
             $var = trim($var);
 
-            if (preg_match('/^P/', $var)) {
+            if (substr($var, 0, 1) === 'P') {
```
Glavic: Even shorter code would be `$var[0] === 'P'` :)

kylekatarnls: No thanks :) Reserve `[]` for arrays allow to identify array keys faster.

tests/Carbon/IsTest.php
```diff
@@ -708,28 +708,47 @@ public function testIsSaturday()
...
         $this->assertTrue(Carbon::now()->endOfDay()->isEndOfDay());
 
         $this->assertFalse(Carbon::parse('15:30:45')->isEndOfDay());
         $this->assertFalse(Carbon::now()->startOfDay()->isEndOfDay());
     }
 
+    public function testIsEndOfDayWithMicroseconds()
+    {
+        $this->assertTrue(Carbon::parse('23:59:59.999999')->isEndOfDay(true));
+        $this->assertTrue(Carbon::now()->endOfDay()->isEndOfDay(true));
```
kylekatarnls: I guess it will fail on PHP 7.0, but we can just skip the test with a reference to the bug.

kylekatarnls:
microseconds support is as bad as this: https://3v4l.org/MtAlY

kylekatarnls:
I take this action. I will squash and rebase all that history to be able to require it with composer and start documenting 1.28.

PR #1271, opened 2018-04-30
Carbon interval cascade

kylekatarnls:
Allow custom cascade factors to handle duration strings input and output such as work logging

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -939,7 +939,7 @@ class CarbonInterval extends DateInterval
             $result = ($result + $this->$source) / $factor;
         }
 
-        if (!$cumulativeFactor) {
+        if (isset($target) && !$cumulativeFactor) {
             $result += $this->$target;
         }
 
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -165,7 +165,7 @@ class CarbonInterval extends DateInterval
             $spec .= $months > 0 ? $months.static::PERIOD_MONTHS : '';
 
             $specDays = 0;
-            $specDays += $weeks > 0 ? $weeks * Carbon::DAYS_PER_WEEK : 0;
+            $specDays += $weeks > 0 ? $weeks * static::getDaysPerWeek() : 0;
             $specDays += $days > 0 ? $days : 0;
 
             $spec .= $specDays > 0 ? $specDays.static::PERIOD_DAYS : '';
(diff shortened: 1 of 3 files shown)


imbrish:
Could you explain here shortly how these custom cascading factors work?

kylekatarnls:
It allow to change factors:

```php
CarbonInterval::setCascadeFactors(array(
    'seconds' => array('minutes', Carbon::SECONDS_PER_MINUTE),
    'minutes' => array('hours', Carbon::MINUTES_PER_HOUR),
    'hours' => array('dayz', 8),
    'dayz' => array('weeks', 5),
));
CarbonInterval::getFactor('hours', 'dayz'); // 8
CarbonInterval::getFactor('dayz', 'months'); // null
CarbonInterval::getFactor('dayz', 'weeks'); // 5
```

So with this settings you can do your work log:
```php
$work = CarbonInterval::fromString('3d')
    ->add(CarbonInterval::fromString('1d 5h'))
    ->add(CarbonInterval::fromString('7h'))
    ->cascade()
    ->forHumans(true);

echo $work; // 1w 4h
```

imbrish:
What about cascading units which are not defined in the factors array?

Additionally defining factors as "week is 5 days" would be more intuitive, ie:
```php
CarbonInterval::setCascadeFactors(array(
    'dayz' => array(8, 'hours'),
    'weeks' => array(5, 'dayz'),
));
```

kylekatarnls:
They will simply not cascade:
```php
CarbonInterval::setCascadeFactors(array(
     'minutes' => array('hours', Carbon::MINUTES_PER_HOUR),
     'dayz' => array('weeks', 5),
));
echo CarbonInterval::fromString('3d 50m')
     ->add(CarbonInterval::fromString('1d 5h 30m'))
     ->add(CarbonInterval::fromString('7h 45m'))
     ->add(CarbonInterval::fromString('1w 15m'))
     ->cascade()
     ->forHumans(true);
```
Output: `1w 4d 14h 20m` different groups just don't mix like in the original DateInterval.

I will try to revert the array format if it does not cost too much for performances.

imbrish:
I think `total()` should throw Exception when source unit cannot be converted to target unit because of missing definition in cascades array.

imbrish:
Allow it to be omitted only if value of the unit is 0. It makes no sense that an interval of 3h 30m will give you total of 30m because 3h is omitted in the cascades.

kylekatarnls:
You could want to cascade day-month-year and second-minutes-hours, then get totalDays and totalSeconds from this interval.

But a 0-code or `-1` to explicitly split the cascade could be fine.

imbrish:
Yeah that would make sense, missing or falsey value forces a breakpoint.

PR #1273, opened 2018-05-01
Add Diff translations for Dutch

megaxlr:
```
There is still one string that I could  not translate,  since that does
not exist in Dutch.

If anyone else knows the word for "before yesterday" in English, please
leave a comment on the PR.

"Voorgisteren" would be the closest I can imagine, but its not correct.

Closing briannesbitt/Carbon#1272

Signed-off-by: MegaXLR <admin@megaxlr.net>
```

### DIFF
--- a/src/Carbon/Lang/nl.php
+++ b/src/Carbon/Lang/nl.php
@@ -28,4 +28,9 @@ return array(
     'from_now' => 'over :time',
     'after' => ':time later',
     'before' => ':time eerder',
+    'diff_now' => 'nu',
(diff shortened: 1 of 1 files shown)


kylekatarnls:
All words are not mandatory, if "2 days before" is better than any other word or word group, then it should be kept as is and so translation will fallback to default translation for this language.

robbertstevens:
'Eergisteren' is the word for before yesterday

kylekatarnls:
Thanks for confirmation.

PR #1277, opened 2018-05-03
Add diff translations for Serbian

Glavic:


src/Carbon/Lang/sr_Cyrl.php
```diff
@@ -35,4 +35,9 @@
     'week_from_now' => '{1} :count недељу|{2,3,4} :count недеље|[5,Inf[ :count недеља',
     'week_ago' => '{1} :count недељу|{2,3,4} :count недеље|[5,Inf[ :count недеља',
 
+    'diff_now' => 'управо сада',
+    'diff_yesterday' => 'јуче',
+    'diff_tomorrow' => 'сутра',
+    'diff_before_yesterday' => 'прекјуче',
+    'diff_after_tomorrow' => 'прексутра',
```
kylekatarnls: @Glavic Google translate says `прексутра` is `forex` and would suggest `прекосутра` instead. Is it a mistake?

PR #1284, opened 2018-05-13
Occitan translation

Quenty31:
Locale file language for Occitan language

PR #1285, opened 2018-05-13
OcTest for Occitan language

Quenty31:


kylekatarnls:
Please add this commit to #1284. Thanks.

PR #1287, opened 2018-05-13
Add occitan language

kylekatarnls:


Quenty31:
Thanks for your help!
Do we know why it fails?

kylekatarnls:
Yes, see the output:
```
1) Tests\Localization\OcTest::testDiffForHumansLocalizedInOccitan
--- Expected
+++ Actual
@@ @@
-fa 2 segondas
+fa 2 segonda
/home/travis/build/briannesbitt/Carbon/tests/Localization/OcTest.php:29
/home/travis/build/briannesbitt/Carbon/tests/AbstractTestCase.php:161
/home/travis/build/briannesbitt/Carbon/tests/AbstractTestCase.php:167
/home/travis/build/briannesbitt/Carbon/tests/Localization/OcTest.php:77
--
```
https://travis-ci.org/briannesbitt/Carbon/jobs/378387048

From my guess `Carbon::now()->subSeconds(2)->diffForHumans()` returns `fa 2 segonda` (would means that symfony/translation has not 2 has plural count in Occitan).

Are you sure 2 is plural in Occitan? Can you check the plural form rule for this language and change tests to trigger the plural (if it exists but is different).

Else, if a short form (for abbreviations) exists in Occitan, please provide it. (See **en** file: https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/en.php)

Quenty31:
Yes, like French, Catalan, Spanish, 2 is plural in Occitan.
0 and 1 are singular.
[...]
'min' => ':count min|:count min',
'second' => ':count segonda|:count segondas',
's' => 's' => ':counts|:counts',
[...]
I changed the min and s lines, it should be okay like this.
Many thanks @kylekatarnls !

kylekatarnls:
First French and Spannish are different: 0 is singular in French but not in Spannish.

So we must to find the official Occitan plural rule (known if it's fr group, es group or something else) then add it to: https://github.com/symfony/translation/blob/master/PluralizationRules.php

Because it's not supported for the moment in symfony/translation.

kylekatarnls:
ca is like es: 0 is plural.

@Quenty31 Can we extrapolate than oc is more like es and ca, rather than french and so propose to symfony/translation too add oc in the same group than ca and es?

PR #1289, opened 2018-05-14
Throw exceptions for unknown CarbonInterval setters

imbrish:
Currently magic setters of CarbonInterval will swallow any invalid input without complaining:

```php
CarbonInterval::foobar(123); // returns null
CarbonInterval::create()->foobar(123);
CarbonInterval::create()->foobar = 123;
```

However getters will throw an exception:

```php
CarbonInterval::create()->foobar; // InvalidArgumentException
```

It's inconsistent with Carbon, which will throw exceptions in all these cases. It's also difficult to debug when you mistype a setter somewhere in your code and then try to find out what's wrong.

Obviously it's a breaking change thus dunno to which version it should go.


kylekatarnls:
You can either wait for the 2.x development to start or remove the exception for the moment.

src/Carbon/CarbonInterval.php
```diff
@@ -181,38 +181,9 @@ public static function create($years = 1, $months = null, $weeks = null, $days =
...
-            case 'minute':
-                return new static(null, null, null, null, null, $arg);
-
-            case 'seconds':
-            case 'second':
-                return new static(null, null, null, null, null, null, $arg);
-        }
+        return call_user_func_array(
+            array(new static(0), $name), $args
+        );
```
kylekatarnls: @imbrish This is shorter code but it means construct + call instead of just construct. It rely on the fluent setter when previously not needed. This means this could easily break some CarbonInterval sub-class.

imbrish:
I'm fine to wait with this until 2.0.

kylekatarnls:
Rebased: #1325

PR #1296, opened 2018-05-16
Fix parsing relative formats with mock and microseconds before PHP 7.1

imbrish:
Fixed a little inconsistency when parsing relative formats with a mock instance prior to PHP 7.1:

```php
Carbon::now(); // 2018-05-16 18:11:12.123456
Carbon::parse('12:25'); // 2018-05-16 12:25:00.000000

Carbon::setTestNow('2018-05-16 18:11:12.123456');
Carbon::parse('12:25'); // PHP >= 7.1: 2018-05-16 12:25:00.000000
                        // PHP  < 7.1: 2018-05-16 12:25:00.123456
Carbon::parse('17:31:24.654321'); // PHP >= 7.1: 2018-05-16 17:31:24.654321
                                  // PHP  < 7.1: 2018-05-16 17:31:24.123456
```

All absolute and relative formats will now work correctly with microseconds.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1530,7 +1530,7 @@ class Carbon extends DateTime implements JsonSerializable
         $date2 = new DateTime('2001-12-25T17:59:13Z');
         $date2->modify($time);
 
-        if ($date1 == $date2) {
+        if ($date1 == $date2 || $date1->format('Y-m-d') === $date2->format('Y-m-d')) {
             return false;
         }
 
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -471,6 +471,8 @@ class Carbon extends DateTime implements JsonSerializable
         // instance then override as required
         $isNow = empty($time) || $time === 'now';
         if (static::hasTestNow() && ($isNow || static::hasRelativeKeywords($time))) {
+            list($relativeKeywords, $clearMicroseconds) = static::hasRelativeKeywords($time) ?: array(false, false);
+
             $testInstance = clone static::getTestNow();
 
             //shift the time according to the given time zone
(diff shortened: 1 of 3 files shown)


src/Carbon/Carbon.php
```diff
@@ -1500,22 +1512,29 @@ public static function hasTestNow()
     /**
      * Determine if a time string will produce a relative date.
      *
+     * Return false when time is absolute or otherwise return array with two elements.
+     * First element is always true, second will be true if microseconds should be cleared.
+     *
      * @param string $time
      *
-     * @return bool true if time match a relative date, false if absolute or invalid time string
+     * @return false|array
```
kylekatarnls: BC means 2.x, but 2.x will be for PHP >= 7.1.8 only on which this fix will no longer be relevant

tests/Carbon/TestingAidsTest.php
```diff
@@ -111,18 +111,39 @@ public function testParseRelativeWithTestValueSet()
...
-    public function testHasRelativeKeywords()
+    public function testParseRelativeWithTestValueSetAndMicroseconds()
     {
-        $this->assertFalse(Carbon::hasRelativeKeywords('sunday 2015-02-23'));
-        $this->assertTrue(Carbon::hasRelativeKeywords('today +2014 days'));
-        $this->assertTrue(Carbon::hasRelativeKeywords('next sunday -3600 seconds'));
-        $this->assertTrue(Carbon::hasRelativeKeywords('last day of this month'));
-        $this->assertFalse(Carbon::hasRelativeKeywords('last day of last month of 2015'));
-        $this->assertTrue(Carbon::hasRelativeKeywords('first sunday of next month'));
-        $this->assertFalse(Carbon::hasRelativeKeywords('first sunday of January 2017'));
```
kylekatarnls: No test removing until 2.x

imbrish: These tests were moved to RelativeDateStringTest. Only `last day of last month of 2015` was removed because it's invalid (`new DateTime('last day of last month of 2015')` throws exception).

src/Carbon/Carbon.php
imbrish: Could be moved to a different method.

imbrish:
```php
$storage = Carbon::today()->toDateTimeString();
$this->assertEquals(Carbon::today(), new Carbon($storage));
```

This will fail only when:
1. PHP version is lower than 7.1
2. The `static::$testNow` is defined
3. The microseconds fallback is enabled
4. Dates are compared with microseconds

Indeed not much space for this bug to become apparent.

PR #1310, opened 2018-05-18
Update carbon period

imbrish:
Changes:

1. Removed caching of iteration results.
     - Every `toArray`, `count`, `first`, `last` call will cause a full iteration.
     - Initialization is done when current is null (only before first iteration).
     - Validation result is still cached per iteration.

2. Removed start date filter.
    - Start date is used only for initialization during rewind. Changing it during the iteration will have no effect whatsoever.
    - DateInterval can now be inverted during iteration.

3. Start date is now required.
    - Exception will be thrown when null is passed to setter, similarly to DateInterval.
    - By default start date is set to now.

### DIFF
--- a/src/Carbon/CarbonPeriod.php
+++ b/src/Carbon/CarbonPeriod.php
@@ -207,7 +207,7 @@ class CarbonPeriod implements Iterator, Countable
     /**
      * The cached validation result for current date.
      *
-     * @var bool|null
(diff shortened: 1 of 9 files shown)


kylekatarnls:
Let's restart from this point, now I will write the documentation and write expected code of each example as a unit test.

PR #1317, opened 2018-05-24
CarbonPeriod final touches

imbrish:
I looked through the code to check if everything looks good and ended up changing a few things here ad there.

@kylekatarnls Did you think about renaming "recurrences" to "limit"? I think it describes what the filter does better and avoids confusion with how DatePeriod recurrences work. This would give us freedom to implement native recurrences filter later and avoid conflict with DatePeriod if we decide to extend it after it's fixed.

### DIFF
--- a/src/Carbon/CarbonPeriod.php
+++ b/src/Carbon/CarbonPeriod.php
@@ -149,7 +149,7 @@ class CarbonPeriod implements Iterator, Countable
      *
      * @var bool
      */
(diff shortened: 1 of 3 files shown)


imbrish:
http://period.thephpleague.com/ has some interesting methods regarding comparing and modifying periods, but I guess it's a separate functionality and can be added later.

imbrish:
Note: add test for memory consumption during iteration

PR #1332, opened 2018-05-29
Fix history for version 1.28

imbrish:
- removed CarbonPeriod entry
- summarized "Added getCascadeFactors, setCascadeFactors, ..."
- added workaround for timezone bug, supprt for macros and invert method
- reordered

https://github.com/briannesbitt/Carbon/compare/version-1.27...version-1.28


kylekatarnls:
Thanks.

PR #1348, opened 2018-06-06
Allow to call modify/parse string as a fluent method

kylekatarnls:


imbrish:
It seems cool to just be able to call `Carbon::firstWednesdayOfLastMonth('Europe/Paris')` and get rid of all the `yesterday()` and `tomorrow()` methods, but... won't it be too confusing? Just imagine overtaking a project having no prior experience with Carbon and seeing all these weird methods. Is it really that much of an improvement over `Carbon::create('first wednesday of last month', 'Europe/Paris')`?

I had a related doubt when adding "aliases" to the CarbonPeriod. By providing many ways to do same thing we introduce confusion, which method is the best? This will inevitably lead people to using one variant here, and another there, making code inconsistent and probably more difficult to maintain. Maybe instead we should make the decision which way is the right one and stick with it?

I'm not firmly against I just wanted to point out a few possible issues.

Glavic:
I kind'a agree with @imbrish. People currently use `new DateTime('first ...')` and `Carbon::create('first ...')` would seem normal to them.

kylekatarnls:
I see there also the opportunity to chain things:
```php
$date = getSomeCarbonInstanceFromOtherPlace();
$date->firstDayOfNextMonth()->addWeekdays(5)->nextSunday();
```
In a cleaner and shorten way than:
```php
$date->modify('first day of next month')->addWeekdays(5)->modify('next sunday');
```

Glavic:
When you chain it like that it does look nicer :) Still, do

kylekatarnls:
"that" refer to short syntax or long syntax?

imbrish:
The fluent chain looks nice indeed and I'm sure Laravel folks will like it. But how often do you need something like `$date->firstDayOfNextMonth()->addWeekdays(5)->nextSunday();` in a real application? Can you think about some practical use cases?

36864:
Holiday calculations would be a use case. For example, some holidays happen "On the first Monday after the first Sunday of this month". This new fluent syntax would let you express that as `Carbon::firstSundayOfMarch()->nextMonday()`.

kylekatarnls:
Thanks for your opinions. So I close this for now as unmerged, will move it in a dedicated branch. Maybe we can do it without the try-catch and with some parsing instead which would be possible to generate a PHPDoc from and so enable IDE autocompletion.

PR #1349, opened 2018-06-07
day of week string

pirupius:
Currently days of the week are defined in constants.
```php
// These getters specifically return integers, ie intval()
var_dump(Carbon::SUNDAY);                          // int(0)
var_dump(Carbon::MONDAY);                          // int(1)
var_dump(Carbon::TUESDAY);                         // int(2)
var_dump(Carbon::WEDNESDAY);                       // int(3)
var_dump(Carbon::THURSDAY);                        // int(4)
var_dump(Carbon::FRIDAY);                          // int(5)
var_dump(Carbon::SATURDAY);                        // int(6)
```

So if say someone uses `Carbon::now()->dayOfWeek`, it returns an index of the array days. 

With this function, it uses the array index to return a day as a string

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -4642,4 +4642,47 @@ class Carbon extends DateTime implements JsonSerializable
 
         return call_user_func_array($macro, $parameters);
     }
+
+    /**
+     * Use Carbon::now() to determine day of week as a string.
+     *
+     *
+     * @return string $day
+     */
+    public function dayOfWeek()
+    {
+        switch ($this->dayOfWeek) {
+            case 0:
+                return self::$days[0];
+                break;
+
+            case 1:
+                return self::$days[1];
+                break;
+
+            case 2:
+                return self::$days[2];
+                break;
+
+            case 3:
+                return self::$days[3];
+                break;
+
+            case 4:
+                return self::$days[4];
+                break;
+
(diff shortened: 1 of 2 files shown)


kylekatarnls:
What difference with:
```php

$date = Carbon::now();

echo $date->formatLocalized('%A'); // week day translated date in the current locale
echo "\n";
echo $date->format('l'); // week day in English
```

pirupius:
@kylekatarnls the options you suggested are good work arounds but you must admit that helper looks way cleaner. So instead of doing 

    Carbon::now()->formatLocalized('%A')

you sould just do

    Carbon::now()->dayOfWeek()

pirupius:
@kylekatarnls it was just a suggestion that like you said would be a big switch statement on the backend but will be easier to the developers consuming the api with a helper way easier to remember. Thank you for your feedback though. i look forward to Carbon 2 :+1:

pirupius:
@kylekatarnls thank you for the feedback. will make the requested changes and push them so you can take a second pass at it.

kylekatarnls:
This is how I imagine such a feature: https://github.com/briannesbitt/Carbon/compare/master...kylekatarnls:string-getters

We can still discuss about naming, getter vs. method and how to document it.

kylekatarnls:
Fixed via #1354

PR #1352, opened 2018-06-11
Carbon timezone + microseconds

kylekatarnls:


imbrish:
1. Why both en and local variants? Is it due to `$c->englishMonth == 'may'` instead of `$c->month == 5`?
2. I'd tune properties names slightly, maybe sth like `monthName`, `weekdayName`, etc?
3. That create signature `create($year, $month = 1, $day = 1, $hour, $minute, $second, $tz);` doesn't look right, I guess optional parameters cannot be followed by obligatory?
4. Rename `isBirthday()` to `isAnniversary()`?
5. Milliseconds look useful 👍 Shouldn't "milli" be a valid unit just as "micro"?
8. What do you think about removing `$weeks` from constructor, for the same reason derivative `$milliseconds` didn't make it into the argument list?
7. Couldn't that `$getDefault` closure be simplified into a simple `$defaults` array? It'll spare some function call overhead and imo will be cleaner.
6. CarbonTimeZone 👍It doesn't look like it's used in the code though, guess `safeCreateDateTimeZone()` should be refactored into that new class?

kylekatarnls:
1. Why both en and local variants? Because there are different results.
2. `monthName` is ambiguous, you can't know if it take the current locale into account.
3. That create signature: this is just automatically generated, I will check why it changed.
4. Seems not related to this PR.
6. Indeed, `$weeks` in constructor seems not legit to me too.
7. This change is about not calling the format method just in time and not if not needed.
8. Yes.

kylekatarnls:
3. is OK now, falsy default value are now auto-generated in the interface

imbrish:
1 & 2: There are different results for any other locale too. I don't think dedicated properties for English are justified and it would be fine to keep just the localized versions. It's no more ambiguous than format method of jenssegers/date. Similarly I'd remove `formatLocalized()` in favor of `format()` (so same format specification can be used), optionally with locale parameter.

4: Still it makes sense to change it :)

5: Do you mean dynamic property or constructor parameter? I referred barely to `fromString` method.

kylekatarnls:
1 & 2. Not sure the jessegers/date pattern must be taken as is in Carbon. Do you mean current `format()` would no longer exist in favor of `formatLocalized()` renamed?
4. You can do it in another PR and so we could think and discuss it.
5. fromString now allow ms or µs with no problem.

imbrish:
I mean implementing format in jenssegers/date way so that it applies current/given locale to standard `date()` format like `D, M j, Y`.

The fromString allows `430000 micro` but not  `430 milli`.

kylekatarnls:
I find this naming annoying, `micro` means nothing without a unit. We should never use neither `micro` nor `milli` alone. It should be `ms`, `millisecond` or `milliseconds`.

But removing all `'micro'` references seems a bit violent.

kylekatarnls:
OK, finally allowed `milli`

PR #1362, opened 2018-06-19
Update pt_BR Language

felipeleite1:
Carbon::JUST_NOW, Carbon::ONE_DAY_WORDS and Carbon::TWO_DAY_WORDS now works in PT-BR Language

### DIFF
--- a/src/Carbon/Lang/pt_BR.php
+++ b/src/Carbon/Lang/pt_BR.php
@@ -28,7 +28,7 @@ return array(
     'from_now' => 'em :time',
     'after' => 'após :time',
     'before' => ':time atrás',
-    'diff_now' => 'hoje',
+    'diff_now' => 'agora',
     'diff_yesterday' => 'ontem',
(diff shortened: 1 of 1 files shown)


src/Carbon/Lang/pt_BR.php
```diff
@@ -28,4 +28,13 @@
     'from_now' => 'em :time',
     'after' => 'após :time',
     'before' => ':time atrás',
+    'diff_now' => 'hoje',
```
kylekatarnls: `diff_now` only apply for a date less than 1 second ago. But there is no "today" word. In english for example if the date is 2 hours old, you get "2 hours ago". So are you sure there is no better fit for "just now"?

felipeleite1: Thanks for your reply, in this case the better word is "agora"

kylekatarnls:
Thanks!

Issue #1364, opened 2018-06-21
Make it possible to force diffForHumans() to return "before/after" or "ago/from now" as I wish, independently of the $other argument being present or not


rulatir:
I have valid use cases where I want diffForHumans() to say "ago" even though I am giving a reference date, and "before" even though I am not giving a reference date.

Please make it possible to specify this with an extra option, as opposed to binding this to whether the reference date was given or not.

imbrish:
I don't think that will justify adding fifth parameter to `diffForHumans`.

You can instead call `$date->diffForHumans($other/null, true)` to get the difference without either "ago" or "before" and add wanted suffix yourself.

It could be possible to rename second parameter to `$suffix` and let it accept special values like `Carbon::ABSOLUTE` and `Carbon::RELATIVE` in addition to default true for automatic and false for none, but as it is BC break I guess only in 2.0.

kylekatarnls:
@imbrish Warning, ago, from now, before, after, are not suffixes in every languages, in French, ago and from_now gives prefixes. So I recommend to always use the translator:

```php
$date->diffForHumans(null, true);
// Get absolute syntax even if $other is null
$transId = $date->isFuture() ? 'after' : 'before';
echo Carbon::translator()->trans($transId, array('time => $date));

$date->diffForHumans($other, true);
// Get relative syntax even if $other is not null
$transId = $date->isFuture() ? 'from_now' : 'ago';
echo Carbon::translator()->trans($transId, array('time => $date));
```

Else I agree to change the `$absolute` boolean parameter with `$mode` integer parameter using constants in the version 2.

kylekatarnls:
Will be implemented in version 2 with the following: https://github.com/briannesbitt/Carbon/pull/1372

kylekatarnls:
Merged, you can test it by installing the verions `dev-version-2.0`

PR #1372, opened 2018-06-26
Implement #1364 diffForHumans can be forced to relative/absolute mode

kylekatarnls:
The `$absolute` boolean parameter for `diffForHumans` become a `$syntax` integer flag:

```php
CarbonInterface::DIFF_ABSOLUTE // no difference modifiers: match the old $absolute = true
CarbonInterface::DIFF_RELATIVE_AUTO // add before/after if a date is given as second argument, add ago/from now if no other date is given: match the old $absolute = false
CarbonInterface::DIFF_RELATIVE_TO_NOW // add ago/from now
CarbonInterface::DIFF_RELATIVE_TO_OTHER // add before/after
```

### DIFF
--- a/src/Carbon/Traits/Difference.php
+++ b/src/Carbon/Traits/Difference.php
@@ -311,23 +311,25 @@ trait Difference
      * 5 months after
      *
      * @param Carbon|null $other
-     * @param bool        $absolute removes time difference modifiers ago, after, etc
-     * @param bool        $short    displays short format of time units
-     * @param int         $parts    displays number of parts in the interval
+     * @param int         $syntax difference modifiers (ago, after, etc) rules
+     * @param bool        $short  displays short format of time units
+     * @param int         $parts  displays number of parts in the interval
      *
      * @return string
      */
-    public function diffForHumans($other = null, $absolute = false, $short = false, $parts = 1)
+    public function diffForHumans($other = null, $syntax = CarbonInterface::DIFF_RELATIVE_AUTO, $short = false, $parts = 1)
     {
         /* @var CarbonInterface $this */
-        $isNow = $other === null;
         $interval = [];
+        $syntax = (int) $syntax;
+        $absolute = ($syntax === CarbonInterface::DIFF_ABSOLUTE);
+        $isNow = $syntax === CarbonInterface::DIFF_RELATIVE_TO_NOW || $syntax === CarbonInterface::DIFF_RELATIVE_AUTO && $other === null;
 
         $parts = min(6, max(1, (int) $parts));
         $count = 1;
         $unit = $short ? 's' : 'second';
 
-        if ($isNow) {
+        if ($other === null) {
             $other = $this->nowWithSameTz();
         } elseif (!$other instanceof DateTimeInterface) {
             $other = static::parse($other);
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -434,10 +434,21 @@ use JsonSerializable;
  */
 interface CarbonInterface extends DateTimeInterface, JsonSerializable
 {
+    /**
+     * Diff wording options.
+     */
     public const NO_ZERO_DIFF = 01;
     public const JUST_NOW = 02;
     public const ONE_DAY_WORDS = 04;
     public const TWO_DAY_WORDS = 010;
+
+    /**
+     * Diff syntax options.
+     */
+    public const DIFF_ABSOLUTE = 1; // backward compatibility with true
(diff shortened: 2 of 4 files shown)


imbrish:
Hey! Wouldn't it be enough to access the constants by `static::...`? If my reasoning is correct they should be inherited from the interface and accessible in the traits.

I like the backwards compatibility. Unfortunately the constants are lengthy, particularly with FQCN, like when used in templates. Maybe non-int values fe `true/false/'now'/'other'` would give more flexibility?

Also current implementation changes results for non-numeric strings or floats less than 1, which become 0 when casted to int.

src/Carbon/Traits/Difference.php
```diff
@@ -311,23 +311,25 @@ public function secondsUntilEndOfDay()
...
      *
      * @return string
      */
-    public function diffForHumans($other = null, $absolute = false, $short = false, $parts = 1)
+    public function diffForHumans($other = null, $syntax = CarbonInterface::DIFF_RELATIVE_AUTO, $short = false, $parts = 1)
     {
         /* @var CarbonInterface $this */
-        $isNow = $other === null;
         $interval = [];
+        $syntax = (int) $syntax;
```
imbrish: This will change result for non-numeric strings or floats less than 1 which become 0 when casted, but previously would evaluate to true.

src/Carbon/Traits/Difference.php
```diff
@@ -311,23 +311,25 @@ public function secondsUntilEndOfDay()
...
+        $absolute = ($syntax === CarbonInterface::DIFF_ABSOLUTE);
+        $isNow = $syntax === CarbonInterface::DIFF_RELATIVE_TO_NOW || $syntax === CarbonInterface::DIFF_RELATIVE_AUTO && $other === null;
```
imbrish: I would suggest renaming this variable to sth like `$relativeToNow`.

kylekatarnls:
About access the constants by `static::...`. Indeed it's what we do for `static::HOURS_PER_DAY` so we will need to choose 1 way and stick to it.

Using `static` means that implementing `CarbonInterface` is mandatory to use any of our traits. Even if it's not exactly what we use them for, as traits are public, we must act as if they could be used separately.

So interfaces also should be split into the different needs. So we would have `CarbonDiffInterface` is mandatory for `Difference` trait, then `CarbonInterface` would extends different other interfaces matching the traits.

src/Carbon/Traits/Difference.php
kylekatarnls: Yes but previously, `2` and `3` was evaluated to `true`, now they are special modes. It's not meant to be a perfect backward compatibility support, `boolean` well formed as in the doc and unit tests will continue to work the same and I think it's enough backward compatibility for a change that comes in a major version upgrade.

kylekatarnls:
> Maybe non-int values fe true/false/'now'/'other' would give more flexibility?

I would prefer to make those too-many-arguments methods multiple methods some auto-generated methods like:
```(short|long)(Absolute|Relative|RelativeToOther|RelativeToNow)DiffForHumans```
I known it's a bit long but it would probably the cleaner syntax for a call in templates.

src/Carbon/Traits/Difference.php
imbrish: Compatibility could be stronger with just these two special cases. On the other hand I expect most people typed true so it's probably fine.

kylekatarnls:
> It all depends whether traits are for code separation or for standalone use.

You cannot forbid the user to use the trait alone. So we must make our traits working as they can be used. As long as private traits does not exist in PHP, we should consider a trait as an entity with its own public API.

> $date->diffForHumans('other', 'short')

But there is more arguments: date and parts.

> calling ago/from now "absolute", before/after "relative"

But it would be wrong. Before "absolute" meant: no modifiers at all, and I find this more accurate for a diff, all other are relative with different wordings.

imbrish:
I thought about `function diffForHumans(...$args)` and parsing arguments one by one but I realize it's not the most obvious way to do it.

At first calling a diff between dates absolute didn't seem right, but thinking about it in terms of absolute **value** as opposed to a negative makes sense. Let's leave it at that.

The fact that we cannot prohibit usage of traits doesn't mean that we have to make them completely standalone or resign from them altogether. We guarantee no BC breaks in public methods and tested behavior thus we stick to it. We don't give any guarantees about internal code/traits so no restrictions.

Users can extend Carbon classes or use Reflection - does it mean that now we cannot change protected and private methods? Making things private for the sake of protecting against complains doesn't make sense, just as complaining that an update broke something you were not supposed to do. I think developers and users should have more faith in each other and our lives will be simpler :)

PR #1373, opened 2018-06-26
[i18n]Occitan update for #1367

Quenty31:
Line 40 has to be plural, otherwise it's "còp"


Quenty31:
Changed! Thanks for the tip.
Now making a PR for Catalan, I only have to drop the ` on the O and it's done

kylekatarnls:
Thanks!

PR #1382, opened 2018-07-04
Wrong day of a week exception

max-kovpak:
If set start of a week more than 6 or less than 0 and then call startOfWeek we will get an infinite while loop... So, I think it is a good idea to throw Exception when a user sets start/end of a week.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1451,6 +1451,10 @@ class Carbon extends DateTime implements JsonSerializable
      */
     public static function setWeekStartsAt($day)
     {
+        if ($day > static::SATURDAY || $day < static::SUNDAY) {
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Thanks. I agree. Could you add one unit test for each to cover those exceptions.

max-kovpak:
@kylekatarnls done

mikemand:
Is that test actually able to run? It doesn't have the `.php` suffix. (If PHPUnit doesn't need it, then I'm going to make all of my tests without it. 😁)

kylekatarnls:
And what if you submitted a PR to fix it?

mikemand:
@kylekatarnls I was asking if it really works. If it **is** a bug, sure - I'll PR it.

kylekatarnls:
Same for me I was just proposing you to fix it. But now it's done: #1383 Thanks y'all guys.

PR #1387, opened 2018-07-12
addDaysNoOverflow added

Maxim-Mazurok:
Inspired by [question](https://stackoverflow.com/q/51295911/4536543) on SO, I've created function `addDaysNoOverflow()`, that acts pretty much like `addMonthsNoOverflow()`.
Hope that it's ready for merge, thanks.

### DIFF
--- a/tests/Carbon/AddMonthsTest.php
+++ b/tests/Carbon/AddMonthsTest.php
@@ -39,7 +39,7 @@ class AddMonthsTest extends AbstractTestCase
         );
     }
 
-    public function providerTestAddDayNoOverflow()
+    public function providerTestAddDaysNoOverflow()
     {
         return array(
             array(-2, 2016, 1, 29),
@@ -47,6 +47,17 @@ class AddMonthsTest extends AbstractTestCase
             array(0, 2016, 1, 31),
             array(1, 2016, 1, 31),
             array(2, 2016, 1, 31),
+        );
+    }
+
+    public function providerTestSetDaysNoOverflow()
+    {
+        return array(
+            array(0, 2016, 1, 31),
+            array(1, 2016, 1, 1),
+            array(2, 2016, 1, 2),
+            array(32, 2016, 1, 31),
+            array(100, 2016, 1, 31),
         );
     }
 
@@ -64,7 +75,7 @@ class AddMonthsTest extends AbstractTestCase
     }
 
     /**
-     * @dataProvider \Tests\Carbon\AddMonthsTest::providerTestAddDayNoOverflow
+     * @dataProvider \Tests\Carbon\AddMonthsTest::providerTestAddDaysNoOverflow
      *
      * @param int $days
      * @param int $y
@@ -74,6 +85,19 @@ class AddMonthsTest extends AbstractTestCase
     public function testAddDaysNoOverflow($days, $y, $m, $d)
     {
         $this->assertCarbon($this->carbon->addDaysNoOverflow($days), $y, $m, $d);
+    }
+
+    /**
+     * @dataProvider \Tests\Carbon\AddMonthsTest::providerTestSetDaysNoOverflow
+     *
+     * @param int $days
+     * @param int $y
+     * @param int $m
+     * @param int $d
+     */
+    public function testSetDaysNoOverflow($days, $y, $m, $d)
+    {
+        $this->assertCarbon($this->carbon->setDaysNoOverflow($days), $y, $m, $d);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


test82987:
please, rename it to setDayNoOverflow and replace this $this->modify((int)$value . ' day'); with $this->day((int)$value) or you'll have problem with months.

Maxim-Mazurok:
Thanks, @test82987, I've renamed function and added tests. Actually, my version passes tests, while your is failing. It may be because `modify` function will increase month and that's why we need last day of **previous** month. While your version just changes day.

Maxim-Mazurok:
@test82987 I've added function setDaysNoOverflow with tests, check it out.

tests/Carbon/AddMonthsTest.php
```diff
@@ -39,6 +39,28 @@ public function providerTestAddMonthNoOverflow()
...
     }
 
+    public function providerTestAddDaysNoOverflow()
+    {
+        return array(
+            array(-2, 2016, 1, 29),
+            array(-1, 2016, 1, 30),
+            array(0, 2016, 1, 31),
+            array(1, 2016, 1, 31),
+            array(2, 2016, 1, 31),
```
kylekatarnls: You won't get expected results with big gap `-40`, `40`, `-100`, `100`, etc.

tests/Carbon/AddMonthsTest.php
```diff
@@ -39,6 +39,28 @@ public function providerTestAddMonthNoOverflow()
...
+    }
+
+    public function providerTestSetDaysNoOverflow()
+    {
+        return array(
+            array(0, 2016, 1, 31),
+            array(1, 2016, 1, 1),
+            array(2, 2016, 1, 2),
+            array(32, 2016, 1, 31),
+            array(100, 2016, 1, 31),
```
kylekatarnls: You won't get expected results with negative values `-1`, `-40`, `-100`, etc.

test82987:
let's admit addDaysNoOverflow is not needed, when setDaysNoOverflow - does not require `-sub` version. Half problems done. Next - negatives should not work at all if you set negative date with no overflow there should be an exception. Next- big numbers - my suggestion throw an same exception if the number is more than 31. Next - zero - should give the same exception in my opinion.

kylekatarnls:
`$date->setDaysNoOverflow($day)` is:
```php
$date->day(min($day, $date->daysInMonth));
```

Should be pretty equivalent to your code. And to allow negative numbers:
```php
$date->day(max(0, min($day, $date->daysInMonth)));
```

imbrish:
Why not just (with `sub` variant):
```php
public function addDaysNoOverflow($value)
{
    return $this->setDaysNoOverflow($this->day + $value);
}
```
And, as @kylekatarnls proposed:
```php
public function setDaysNoOverflow($value)
{
    return $this->day(max(0, min($value, $this->daysInMonth)));
}
```
I think this method could be useful, despite it doesn't solve the actual problem from Stack Overflow.

On a side note, I find the "NoOverflow" suffix confusing, ie it's not clear what won't overflow just by looking at `addMonthsNoOverflow` and `addDaysNoOverflow`. Maybe something like `incrementMonth`, `addDaysWithinMonth` could work better?

kylekatarnls:
Yes, add/sub should call a unified setter to avoid duplicate code. And there it seems we could have all those methods as 1 line of code. So maybe not so necessary.

Then if we have such possibility for day not overflow month, we could have it for any couple of units:
- set/add/sub hour without overflowing day
- set/add/sub week/month/quarter without overflowing year
- set/add/sub milli/microseconds without overflowing second/minute/hour
- etc.

As you can see, it will quickly add a plethora of methods.

imbrish:
You're right that this belongs to the business logic not a date library. Like Taylor often says in Laravel repo, "create a macro" 😄

kylekatarnls:
Maybe we should rather have a single method for that (only compatible with Carbon 2):
```php
public function setNoOverflow($value, $valueUnit, $overflowUnit)
{
    $original = $this->copy();

    return $original->set($valueUnit, $value)
        ->min($original->endOf($overflowUnit))
        ->max($original->startOf($overflowUnit));
}
```
Then addNoOverflow/subNoOverflow calling it. Then here you can set/add/sub any unit not overflowing any unit having endOf/startOf compatible with.

imbrish:
I think this may even be a better than having a bunch of magic `__call` methods. Which applies also in case of other functionalities. As far as I remember this is how moment.js goes about that.

kylekatarnls:
Not really, moment.js is a bit bad about this, it just disabled overflow for month since 2.1.0.

So before this version you can't prevent overflow, after this version, you can't allow overflow.

And a priori, there is nothing for other units.

kylekatarnls:
Fixed via method `addUnitNoOverflow`, `subUnitNoOverflow` and  `setUnitNoOverflow` https://github.com/briannesbitt/Carbon/releases/tag/2.0.0-beta.1

PR #1396, opened 2018-07-20
Support closure based $toStringFormat

jhoff:
Currently you can only use a string to set a default __toString() format.

This allows a closure to be used, which will execute whenever the instance is converted to a string.

This allows conditional formatting based on the current value.

For example:

```php
Carbon::setToStringFormat(function ($dateTime) {
    return $dateTime->isCurrentYear() ? 'M jS' : 'M jS, Y';
});

echo Carbon::now(); // July 19th

echo Carbon::now()->subYears(2); // July 19th, 2016
```

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -1847,7 +1847,7 @@ class Carbon extends DateTime implements JsonSerializable
     /**
      * Set the default format used when type juggling a Carbon instance to a string
      *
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Nice :) Can you just replace the doctype `mixed` with `string|Closure`?

jhoff:
Updated.

PR #1398, opened 2018-07-23
fix exception message

fyzzy1943:


kylekatarnls:
Indeed this message is invalid when called via `instance`, but it's right when called via `resolveCarbon` (this method is called by `setDateFrom`, `setTimeFrom`, `setDateTimeFrom`, `min`, `max`, `average`, `diff*` methods that are pletroa) and `null` but not `string` is allowed via `isSame*`.

So, simply change the message will make the protected method itself right according to what it does. But the most important is the public API, and in the public API only `instance` will be correct when many more methods will get an incorrect message.

A better way would maybe to have an argument to customize the error exception and so:
- `resolveCarbon` would call it with the current message as second argument.
- `isSame*` would call it with a null or date message.
- and other methods would call it without custom message and it would fallback to the new message you propose.

fyzzy1943:
Thank you, I update code, the second argument will add other expected parameters to the exception message.

PR #1400, opened 2018-07-23
Add start/end of second, make the microsecond handling more consistent

szeber:
The endOf* methods were inconsistent when it comes to microseconds. The endOfDay method sets the microseconds to 999999, while the endOfHour and endOfMinute doesn't.

Also there was no startOfSecond/endOfSecond method, which is very useful when it comes to microsecond handling.

Added microsecond conditions to the tests of the startOf/endOf methods, and made the assertCarbon handle the microsecond comparison based on PHP version instead of having to do that in the test method itself.


kylekatarnls:
This change has yet been done in Carbon version 2, so I perfectly agree to merge this, let just fix unit tests (broken because symfony/translation fix a plural bug and our tests was aligned on the bad plural).

PR #1403, opened 2018-07-27
Patch catalan translation

xaviqv:
'from_now' current translation is wrong. It is a literal spanish translation from "dentro de x días / semanas" that can be understood in Catalan but is grammatically wrong.

### DIFF
--- a/src/Carbon/Lang/ca.php
+++ b/src/Carbon/Lang/ca.php
@@ -25,7 +25,7 @@ return array(
     'second' => ':count segon|:count segons',
     's' => ':count segon|:count segons',
     'ago' => 'fa :time',
-    'from_now' => 'dins de :time',
+    'from_now' => 'd\'aquí :time',
     'after' => ':time després',
     'before' => ':time abans',
     'diff_now' => 'ara mateix',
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Checked. Thanks!

Quenty31:
D’aquí una setmana OR Dins una setmana, sounds the same to my occitan ears.
Some of my dictionaries say Dins + epoch like «summer», _dins l'estiu_, and D’aquí + period.
And other accept both.

Quenty31:
You may apply the same modifications to the Occitan, so we stay closer :)

kylekatarnls:
@xaviqv Do you know Occitan and if so do you agree it would be relevant to align those languages ? https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/oc.php

xaviqv:
@Quenty31 I like you find it useful even for Occitan (which I may call Occità) ;) Also I saw your PR. Anyway, @Kylekatarnis, i'm sorry I couldn't give any kind of help with that language which is totally unknown to me, although it's closer to Catalan than others.

PR #1406, opened 2018-07-27
Full Translation Added

saaiful:
Full Translation Added


kylekatarnls:
Hi, thanks. Can you confirm the following forms are used in Bengali:
গত পরশু this would appear for something that happened 2 days ago
পরশু this would appear for something that happened in 2 days

If those words are not commonly used this way, it's better to not add them and let fallback on 2 দিন পূর্বে and এখন থেকে 2 দিন.

Something else, you have keys that appear twice (after/before/second/year, etc.) Could you please remove the duplicate? Thanks.

saaiful:
@kylekatarnls  We use "পরশু" as "after tomorrow" and "গত পরশু" as Before tomorrow.

kylekatarnls:
Before tomorrow is not right. It should match 2 days ago (the day before yesterday) and 2 days from now (the day after tomorrow).

kylekatarnls:
Hi @saaiful could your provide some source for the translation of 2 days before or explain what is "before tomorrow" (from an english and latin languages point of view, before tomorrow = today).

kylekatarnls:
Hi @saaiful, as the 2-days words are not mandatory, I will merge your PR without them. Please let me know if you have more information.

PR #1434, opened 2018-09-11
Refactor the comparison functions

richardhughes:
The short hand versions doesn't really help with the readability of the code and since both functions (short hand and fully named functions) exist we should be promoting clean readable code. This will aid in the deprecation of the short hand functions at a later date. 

- Move the short hand functions to be aliases of the full named functions


kylekatarnls:
Thanks! I'm in favor of this change.

PR #1438, opened 2018-09-19
Fix abbreviations

Fellzo:
Fix wrong abbreviations (old was equals to full)


kylekatarnls:
It could come to next versions 1.34 and 2.2. Is there a particular reason to put it in 1.29?

PR #1440, opened 2018-09-20
Add interface_exists check - fixed in 2.x by #1430

antonioribeiro:
Applying the same fix which was done in #1430


kylekatarnls:
No problem to put this in 1.34

PR #1450, opened 2018-10-03
Remove the JsonSerializable interface and prevent composer validate warning

ntzm:
Before:

> $ composer validate
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
Defining autoload.psr-4 with an empty namespace prefix is a bad idea for performance

After:

> $ composer validate
./composer.json is valid

PHP requirement is above 5.5 so it's safe to drop


kylekatarnls:
@ntzm Regarding to the possibility of having PHP without the JSON extension (see #1461) it is theoretically still possible to get `JsonSerializable` missing even with the last PHP version. So we are considering reverting your pull-request and not having it in the version 2.4. Please feel free to comment #1461 and give your opinion.

PR #1451, opened 2018-10-04
Update the documents to use the longer format functions as a default

richardhughes:
In #1434 we changed the logic to be in the longer named functions. This is a change to the docs so we can start to point people to use the longer function names.

Not sure if this is the way to update the docs - so any feedback is appreciated :)

Thanks!


richardhughes:
I think there is an issue with the generated docs regarding the German Locale - I'll take a look :)

kylekatarnls:
The localization of the version 1 required OS languages packages. It's not a problem I have some editions to do to the doc now Carbon 2 is officially supported in Laravel. I will take your branch an regenerate it.

kylekatarnls:
Thanks.

PR #1453, opened 2018-10-07
Update without-composer instructions

digilist:
As Carbon no longer is a single file library, the current information is outdated and does not work anymore.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,11 +79,11 @@ printf("Now: %s", Carbon::now());
 
 ### Without Composer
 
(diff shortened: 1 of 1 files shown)


kylekatarnls:
Thanks! Completly forgot the README.

kylekatarnls:
I changed a bit because it's not the recommended way. If you download sources without running composer, you won't have dependencies needed for translation or future dependencies we could add. And autoload could be really hard to code properly to get all classes loaded.

So use instead the bundled ZIP we provide for each release, it contains the source, the vendor (dependencies) and an autoloader file yet automatically compiled by composer, the whole thing yet optimized. So it should be really easier this way and ensure, you get it working as with composer.

digilist:
Thanks, yeah that's better :+1:

PR #1455, opened 2018-10-08
Set correctly phpDocs

omarkdev:
Looking some classes I noticed the class has some errors phpDocs annotations, this PR has the intention to solve this errors.

### DIFF
--- a/src/Carbon/Traits/Difference.php
+++ b/src/Carbon/Traits/Difference.php
@@ -392,11 +392,17 @@ trait Difference
      * Get the difference in a human readable format in the current locale from current instance to an other
      * instance given (or now if null given).
      *
-     * @param null $other
-     * @param null $syntax
-     * @param bool $short
-     * @param int  $parts
-     * @param null $options
+     * @param Carbon|null $other
+     * @param int         $syntax  difference modifiers (ago, after, etc) rules
+     *                             Possible values:
+     *                             - CarbonInterface::DIFF_ABSOLUTE
+     *                             - CarbonInterface::DIFF_RELATIVE_AUTO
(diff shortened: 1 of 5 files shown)


src/Carbon/Traits/Difference.php
```diff
@@ -391,6 +391,14 @@ public function diffForHumans($other = null, $syntax = null, $short = false, $pa
      *
      * Get the difference in a human readable format in the current locale from current instance to an other
      * instance given (or now if null given).
+     *
+     * @param null $other
+     * @param null $syntax
+     * @param bool $short
+     * @param int  $parts
+     * @param null $options
```
kylekatarnls: There are no null-types, it would mean you can only pass null as those parameters, please copy types and descriptions from `diffForHumans`. Thanks. https://github.com/briannesbitt/Carbon/pull/1455/files#diff-39a84c4c7e44c741019124ce379a98b5R352

src/Carbon/Traits/Difference.php
```diff
@@ -456,6 +464,14 @@ public function to($other = null, $syntax = null, $short = false, $parts = 1, $o
      *
      * Get the difference in a human readable format in the current locale from an other
      * instance given (or now if null given) to current instance.
+     *
+     * @param null $other
+     * @param null $syntax
+     * @param bool $short
+     * @param int  $parts
+     * @param null $options
```
kylekatarnls: There are no null-types, it would mean you can only pass null as those parameters, please copy types and descriptions from `to`. Thanks. https://github.com/briannesbitt/Carbon/pull/1455/files#diff-39a84c4c7e44c741019124ce379a98b5R439

PR #1459, opened 2018-10-11
Run locale update only if services are registered

CupOfTea696:
Only run the locale update if the 'events' and 'translator' services are registered in the container.


kylekatarnls:
Thanks!

PR #1461, opened 2018-10-11
require JSON extension

browner12:
since we use features from the JSON extension, we should tell composer to require it.

http://php.net/manual/en/book.json.php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
     "minimum-stability": "dev",
     "require": {
         "php": "^7.1.8",
+        "ext-json": "*",
         "symfony/translation": "^4.0"
     },
     "require-dev": {


kylekatarnls:
Hum, I realize maybe we have made a mistake by merging #1450.

Until now we were not supposed to require JSON only provide a JSON-compatible class. We do not call json_encode or json_decode inside the library. And before #1450, if the interface was missing we used an embedded polyfill.

So Carbon up to 2.3 is supposed to work even without the JSON extension enabled. If not it's not on purpose. And it probably no longer work on dev-master but it's not on purpose, as 2.4 is not yet released we should consider reverting #1450 rather than requiring the JSON extension.

I have to think about it and all feedbacks are welcome.

Yet thanks notice it!

browner12:
I would tend to agree. I would revert and add back in the polyfill interface, and then these changes should go to the next major release.

does this package follow the same structure as Laravel, where `master` is the next major release?  That's why I PR'ed this to `master`.

While the change probably was technically a BC break, I don't think you'll run into any real world issues because according to the docs:

> As of PHP 5.2.0, the JSON extension is bundled and compiled into PHP by default.

PR #1469, opened 2018-10-16
Corrected translations

kristianbangsoe:
'from_now' => 'om :time',
to
'from_now' => ':time siden',

### DIFF
--- a/src/Carbon/Lang/da.php
+++ b/src/Carbon/Lang/da.php
@@ -25,7 +25,7 @@ return [
     'second' => 'få sekunder|:count sekunder',
     's' => ':count sekund|:count sekunder',
     'ago' => ':time siden',
-    'from_now' => 'om :time',
+    'from_now' => ':time siden',
     'after' => ':time efter',
     'before' => ':time før',
     'formats' => [


kylekatarnls:
Hi, it would mean both `ago` and `from_now` would have the same translation. Are you sure there is no way  to distinguish past and future?

kylekatarnls:
In Google translation:
https://translate.google.com/#da/en/om%202%20dage
`om` is right while `siden` is the contrary:
https://translate.google.com/#da/en/2%20dage%20siden

`'from_now'` key refer to **time after now** (future)
`'ago'` key refer to **time before now** (past)

So it also match translations in moment.js:
https://github.com/moment/moment/blob/2.22.2/locale/da.js#L34

This seems wrong to me, if you're sure about this please provide sources and explanations.

kylekatarnls:
If om is in, the keys are right and they match your examples, it's most probably a bad usage and not a bug. What code are you using and output are you expecting from it?

kylekatarnls:
See the following:
```php
$sonner = Carbon::now()->subMinutes(10)->locale('da');
$later = Carbon::now()->addMinutes(10)->locale('da');

echo $later->diffForHumans();
echo "\n";
echo $sonner->diffForHumans();
```

It outputs:
```
om 10 minutter
10 minutter siden
```

It matches your example and it matches the Google translation.

PR #1483, opened 2018-10-25
Add trans docblock

kevdotbadger:
Just adds the missing docblock for the trans() method.

### DIFF
--- a/src/Carbon/Translator.php
+++ b/src/Carbon/Translator.php
@@ -110,8 +110,8 @@ class Translator extends Translation\Translator
     /**
      * Returns the translation.
      *
-     * @param int $id
-     * @param array $directory
+     * @param int    $id
+     * @param array  $directory
      * @param string $domain
      * @param string $locale
      *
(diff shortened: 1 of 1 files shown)


src/Carbon/Translator.php
```diff
@@ -107,6 +107,16 @@ public function removeDirectory(string $directory)
         }));
     }
 
+    /**
+     * Returns the translation.
+     *
+     * @param int    $id
+     * @param array  $directory
```
kylekatarnls: There is no directory, this is the interpolations parameter, we named if `parameters` because it's the same name as in symfony/translation used by the parent method.

kevdotbadger:
Whoops. My bad, changed it now.

src/Carbon/Translator.php
kevdotbadger: Updated it, sorry about that.

src/Carbon/Translator.php
```diff
@@ -107,6 +107,16 @@ public function removeDirectory(string $directory)
         }));
     }
 
+    /**
+     * Returns the translation.
+     *
+     * @param int    $id
```
kylekatarnls: The `$id` is casted to `string` and would nearly never be a number. So the type should just be `string`

PR #1484, opened 2018-10-25
fixes #1482

### LINKED ISSUE #1482: static::make() stripping microseconds from CarbonInterval object
Hello,

When providing a CarbonInterval object to the `->add()` method, microseconds are dropped from the resulting computation.

```php
$x = new CarbonInterval(0, 0, 0, 0, 10, 10, 100000);
$y = new CarbonInterval(0, 0, 0, 0, 20, 20, 200000);
$x->add($y);
echo $x->minutes.':'.$x->seconds.':'.$x->microseconds;
```
Carbon version: **^2.00**
PHP version: **7.2.11**

I expected to get:
```
30 :30:300000
```

But I actually get:
```
30:30:0
```

Thanks!

From what I can see, the issue stems from two issues.
1) the `static::make()` call in the `add()` method. Since the first if check compares to a `\DateInterval` object, which CarbonInterval extends, then it goes into `static::instance()`, which instantiates a _new_ CarbonInterval object from the original, just without the microseconds.
2) Within the `add()` method itself, there is no `$this->microseconds += $interval->f * $sign`. (I was previously able to get around this by overloading, calling the parent, then just adding the microseconds myself with v1, but with v2, due to the above issue, the microseconds are stripped from the incoming CarbonInterval object.

Zxurian:
prevent calling static::make in add() if already CarbonInterval instance

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -1112,7 +1112,7 @@ class CarbonInterval extends DateInterval
             $value = 1;
         }
 
-        $interval = static::make($unit);
+        $interval = $unit instanceof self ? $unit : static::make($unit);
 
         if (!$interval) {
             throw new InvalidArgumentException('This type of data cannot be added/subtracted.');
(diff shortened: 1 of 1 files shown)


kylekatarnls:
Microseconds should be fixed globally (right in instance method). I will provide a fix for v2.5.

Issue #1511, opened 2018-11-14
Add more PHP versions to Travis runs


kylekatarnls:
Some patchs of PHP have major impact on Carbon behavior (such as change from PHP 7.0.21 to 7.0.22, or 7.1.7 to 7.1.8) meaning a test can pass in the last version of PHP 7.0, 7.1, etc. but not in some particular ranges of older patch versions, that's why we should enable a wider range of PHP versions in our TravisCI configuration.

kylekatarnls:
I parsed PHP Changelog to detect every change in the Date API, now I try to run tests for all of those versions in Travis to see if there are some failing tests not detected and see wich version Travis is able to run:
https://travis-ci.org/kylekatarnls/Carbon/builds/455013800

kylekatarnls:
A lot of version missed for PHP < 5.6 and 2 version missing for PHP 5.6, Above it, no problem. This range is enough in my opinion as PHP < 5.6 is now very low.

Tests revealed a failure we have to investigate for PHP 7.1.3, 7.1.4 and 7.1.5.

kylekatarnls:
OK for Carbon 2 with no conflict.

kylekatarnls:
PHP versions added for Carbon 1, incompatibilities discovered to be fixed in #1516

Issue #1503, opened 2018-11-12
diffInSeconds broken after PHP 7.2.12 release


driesvints:
Hello,

I encountered an issue with the following code:

```php
$now = Carbon::now();

// some milliseconds pass.

echo Carbon::now()->diffInSeconds($now, false);
```
Carbon version: **1.34.1**
PHP version: **7.2.12**

I expected to get:

```
0
```

But I actually get:

```
A much higher number (differs depending on how much milliseconds have passed).
```

I believe this is because https://bugs.php.net/bug.php?id=77007 was merged for 7.2.12 causing differences between these versions. I'm not sure how this should be solved.

driesvints:
At the moment builds on laravel/framework are failing because of the above issue: https://travis-ci.org/laravel/framework/jobs/453254179

I've traced it down to this method where the `diffInSeconds` call is made: https://github.com/laravel/framework/blob/5.7/src/Illuminate/Cache/Repository.php#L563-L572

I've already tried replacing the `Carbon::createFromTimestamp($duration->getTimestamp())` call by just passing the `$duration` instance (which should be fine in theory) but without luck.

kylekatarnls:
I tried the following:

```php
$now = Carbon::now();

usleep(500000);

echo Carbon::now()->diffInSeconds($now, false);
```

And get 0 or -1, so this seems OK.

Could you add some sample of output with diffInSeconds and the exact difference for this output using:
```php
echo Carbon::now()->diffInSeconds($now, false);
echo "\n\n";
echo Carbon::now()->format('U.u') - $now->format('U.u');
```

kylekatarnls:
Note: we also noted PHP 7.2 started to fail yesterday on our Laravel+Carbon unit tests:
https://travis-ci.org/kylekatarnls/carbon-laravel/builds/453545395

Probably related.

kylekatarnls:
https://bugs.php.net/bug.php?id=77007 is a very big problem for us and the way we handle difference. It clearly can be the root cause.

I'm a bit exhausted of the breaking patches of the PHP DateTime API. I will try to create a work-around.

driesvints:
@kylekatarnls feeling you man. Thanks for looking into this and helping out. Really appreciating it 👍

driesvints:
Any chance we can solve this in a way so people on <7.2.12 aren't left in the dark? If I can help out in any way please let me know.

kylekatarnls:
[...]

PR #1513, opened 2018-11-14
 Test all version that changed PHP date behavior

kylekatarnls:
#1511 Add more PHP versions to Travis runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,12 +50,23 @@ branches:
 
 matrix:
   include:
+    - php: 7.1.8
+    - php: 7.1.9
+    - php: 7.1.11
+    - php: 7.1.15
+    - php: 7.1.17
+    - php: 7.1.20
+    - php: 7.1.24
     - php: 7.1
     - php: 7.1
       env: setup=lowest
     - php: 7.1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -50,12 +50,23 @@ branches:
...
     - php: 7.1
     - php: 7.1
       env: setup=lowest
     - php: 7.1
       env: setup=stable
 
+    - php: 7.2.0
+    - php: 7.2.3
+    - php: 7.2.5
+    - php: 7.2.8
```
driesvints: @kylekatarnls isn't it best to also add 7.2.12 which caused https://github.com/briannesbitt/Carbon/issues/1503 ?

kylekatarnls: 7.2.12 is currently 7.2 (last patch by default), so 7.2.12 does not need to be added until the next date change impacting 7.2.x. If we yet add it, we will run the tests twice on the same version. Then even if PHP released 7.2.13 tomorrow, if the date/time API did not change on this patch, we will not add this specific version to .travis.yml. The goal is to run at least once the tests on one PHP version between each DateTime API change.

PR #1525, opened 2018-11-19
Estonian and Portuguese languages update

jphms:
Just a small update to use Estonian locale in Carbon 2.


kylekatarnls:
Thanks a lot! 😎

kylekatarnls:
Could you update https://github.com/jphms/Carbon/blob/master/tests/Localization/EtTest.php with the actual translations?

src/Carbon/Lang/pt.php
```diff
@@ -28,6 +28,15 @@
     'from_now' => 'em :time',
     'after' => ':time depois',
     'before' => ':time antes',
+    'diff_now' => 'agora',
+    'diff_yesterday' => 'ontem',
+    'diff_tomorrow' => 'amanhã',
+    'diff_before_yesterday' => 'anteontem',
+    'diff_after_tomorrow' => 'depois de amanhã',
```
kylekatarnls: Are you sure `anteontem` and mostly `depois de amanhã` are commonly used? If not, we should rather not translate them and so fallback to calendar translations.

jphms:
Yes, I'll do that. 🤟

src/Carbon/Lang/pt.php
kylekatarnls: Thanks.

jphms:
> Could you update https://github.com/jphms/Carbon/blob/master/tests/Localization/EtTest.php with the actual translations?

Done. 👍

PR #1535, opened 2018-11-23
Fix character escaping in hasFormat()

rdarcy1:
Fixes the regex validation of the format string, which previously didn't correctly deal with escaped characters.


kylekatarnls:
Thanks! Well done.

PR #1536, opened 2018-11-23
fix IDE inspection error

vadik2014:
There was an inspection error in IDE when using chaining on `Carbon::now()`, saying "Method X not found in class". And that huge diff about millis was actually generated by running `composer phpdoc`.

### DIFF
--- a/phpdoc.php
+++ b/phpdoc.php
@@ -16,6 +16,7 @@ $tags = [
     ['call', 'diffForHumans'],
 ];
 $nativeMethods = [
+    'now' => 'static static',
     'format' => 'string',
     'modify' => 'static',
     'getOffset' => 'int',
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -85,6 +85,8 @@ use DateTime;
  * @property-read string         $tzAbbrName                                                                         alias of $timezoneAbbreviatedName
  * @property-read string         $locale                                                                             locale of the current instance
  *
+ * @method        static static  now($tz = null)                                                                     call \DateTime::now if mutable or \DateTimeImmutable::now else.
+ *                                                                                                                   http://php.net/manual/en/datetime.now.php
  * @method        string         format($format)                                                                     call \DateTime::format if mutable or \DateTimeImmutable::format else.
  *                                                                                                                   http://php.net/manual/en/datetime.format.php
  * @method        static         modify($modify)                                                                     call \DateTime::modify if mutable or \DateTimeImmutable::modify else.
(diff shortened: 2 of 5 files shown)


kylekatarnls:
This PHPDoc block should be for magic methods and DateTime native methods, each Carbon static method has its own block:

`now` return doc is:
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Traits/Creator.php#L136

Maybe we should apply a global replacement of `return static` with `return static|CarbonInterface`.

vadik2014:
Another way to fix it is add a docblock to interface method itself:
```
    /**
     * @param string|null $tz
     * @return CarbonInterface
     */
    public static function now($tz = null);
```
But it breaks the overall structure, and I couldn't find another way to place this description inside `@method`s block, other than modifying that array.

kylekatarnls:
Hi, I create a temporary branch. Can you try to install `"nesbot/carbon": "dev-php-doc-return"` and see if your IDE properly find methods?

vadik2014:
Hi Kyle!

Sorry, didn't check this thread for a while. I tried installing your dev branch but unfortunately it doesn't help. Only putting PHPDoc in `CarbonInterface` helps.

vadik2014:
PHPStorm v2018.3. It worked fine on previous versions, apparently they changed PHPDoc inspection algorithms.

vadik2014:
Also if we either add
```
* @method static static now($tz = null)
```
to `Carbon.php` or move `now()` from `Creator` trait to `Carbon` class (including its PHPDoc), it also fixes the problem (same applies to `CarbonImmutable`).

vadik2014:
I'll try to do that. Meanwhile why not add extra comment line to classes? It can take JetBrains years to fix it and during this period the library will be hardly usable. And I expect PhpStorm users to update IDE version sooner or later, and it is quite widespread.

kylekatarnls:
I will wait for feedback from JetBrains staff:
https://youtrack.jetbrains.com/issue/WI-44377

vadik2014:
Thanks! I will also try to generalize this issue and post a ticket, too.

kylekatarnls:
A possible handle from our side is to add a PHPDoc block for each method in the `CarbonInterface`:
```php

    /**
     * @param string|DateTimeZone|null $tz
     *
     * @return CarbonInterface
     */
    public static function now($tz = null);
```

But we should not handle only `now`. It should be done via our `phpdoc.php` script, so we can have it for any method and apply it automatically on change.

kylekatarnls:
Tried to do a simple automated copy of the PHPDoc blocks from traits to the interface on `php-doc-return` and it works fine for me. Can you update on `"nesbot/carbon": "dev-php-doc-return"` and see if all is fine for you.

kylekatarnls:
Sorry, pushed only to my fork, now pushed to upstream.

vadik2014:
Thanks! Now it's much better, but gives "multiple implementations". Seems that only PHPDocs for static methods should be copied to `CarbonInterface`.

kylekatarnls:
Yep, merge it to master is not a problem as it's just but I will check the "Multiple implementations" reason before releasing the new tag.

kylekatarnls:
Merged on `dev-master`

PR #1538, opened 2018-11-24
Add CarbonInterface to PHPDoc return static statements

kylekatarnls:


vadik2014:
Hi Kyle! Could you please reply to JetBrains in https://youtrack.jetbrains.com/issue/WI-44377 ?

kylekatarnls:
I thought post I created would be automatically watched, but I wasn't notified. You can answer to this post too, it's just the code of the screenshot with Carbon 2.5, nothing secret here.

PR #1551, opened 2018-12-16
Updated the Norwegian translation

DanielRTRD:


kylekatarnls:
Thanks, I detected an existing problem thanks to this PR with ordinal `:number.` so I will first merge this in a branch, but I will soon merge it in master and release this for 2.8 when the ordinal bug is fixed.

PR #1554, opened 2018-12-18
Added separator parameter to diffForHumans

oriceon:
In some cases we need another aproach for presented time in diffForHumans. So i added a little change with a separator condition.

### DIFF
--- a/src/Carbon/Carbon.php
+++ b/src/Carbon/Carbon.php
@@ -4123,10 +4123,10 @@ class Carbon extends DateTime implements JsonSerializable
      * 5 months after
      *
      * @param Carbon|null $other
-     * @param bool        $absolute removes time difference modifiers ago, after, etc
-     * @param bool        $short    displays short format of time units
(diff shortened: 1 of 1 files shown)


kylekatarnls:
Hi, thanks, this sounds useful. But this will conflict with the existing `$options` parameter in the version 2. This means an harder migration from version 1 to 2 if you use this parameter. And int and string can cast one to the other so we can't really detect if the user wanted to pass int-options or string-separator. Maybe using an array-options could have multiple parameters in one such as a `'separator' => ', '` entry.

But first, we don't add features on the version 1, in v1 we fix bugs then we merge/copy most important features from v2.

So to implement this feature we would need at least 1 PR for v2 (master branch), then an optional PR for v1 (version-1.next) but both should have the same input and output and so the same unit tests (we require new unit tests for each new feature).

Do you plan to migrate to Carbon 2 soon? Maybe implementing it only in v2 could be easier.

oriceon:
OK, Greate. I`ll port this on V2 :)

PR #1580, opened 2019-01-29
Support for symfony/translation ^3.4

brusch:
Used interfaces `\Symfony\Component\Translation\TranslatorInterface` and `\Symfony\Component\Translation\TranslatorBagInterface` are compatible from `3.4` to `4.0` so why should we lock out users of still supported Symfony 3.4 ☺


kylekatarnls:
`2.11.0` released.

PR #1581, opened 2019-01-31
Add NL period language definitions

Propaganistas:
The missing Dutch language lines.

A note about the `period_interval` definition though. In Dutch we have an inflective keyword for singular intervals, depending on the gender of the word it refers to (notice there's no trailing `-e`). Unfortunately there's currently no clear way to get this right. An edge case, but I can imagine other languages might need this as well. Perhaps a similar approach as `ordinal` might come to the rescue?

|        | **one**  | **multiple** |
|--------|------|----------|
| **second** | elke | elke     |
| **minute** | elke | elke     |
| **hour**   | ⚠️ elk   | elke     |
| **day**    | elke | elke     |
| **week**   | elke | elke     |
| **month**  | elke | elke     |
| **year**   | ⚠️ elk  | elke     |

### DIFF
--- a/src/Carbon/Lang/nl.php
+++ b/src/Carbon/Lang/nl.php
@@ -33,6 +33,10 @@ return [
     'diff_tomorrow' => 'morgen',
     'diff_after_tomorrow' => 'overmorgen',
     'diff_before_yesterday' => 'eergisteren',
+    'period_recurrences' => ':count keer',
+    'period_interval' => 'elke :interval',
+    'period_start_date' => 'van :date',
+    'period_end_date' => 'tot :date',
     'formats' => [
         'LT' => 'HH:mm',
         'LTS' => 'HH:mm:ss',


kylekatarnls:
Yes `period_interval` can be a closure (any translation can). But what happen for:

`every 1 hour and 30 minutes`
`every 1 day and 1 hour`

What is the rule when it comes to intervals with multiple units?

Propaganistas:
Hmm, I supposed singular interval keywords would omit the `1`: `every year` instead of `every 1 year`

For example in Dutch `every 1 year` (`elke 1 jaar`) isn't even valid. It really should be `elk jaar`.

In case of combined granularity the keyword only applies to the first unit:

`every 1 hour and 30 minutes` becomes `elk uur en 30 minuten`
`every 1 week and 1 hour` becomes `elke week en 1 uur`
`every 1 year and 1 month` becomes `elk jaar en 1 maand`

kylekatarnls:
OK, in English and French "every year" is valid but "every year and 6 months" is not. But if "elk jaar en 1 maand", we can simply trim the `"1 "` inside the closure. I will write the closure and tests. You could check if tests are then all right in Dutch.

Propaganistas:
The expected period outputs in tests seem quite ok.
Note that there are a few leading `2`s missing in the year references. Will lead to failed tests.

kylekatarnls:
Hi @Propaganistas please check the following results: https://github.com/briannesbitt/Carbon/pull/1581/files#diff-2fe03638017f94603391efe35656d36dR226

Thanks.

Propaganistas:
Hi @kylekatarnls looks ok to me. Some constructs are rather unnatural but not necessarily wrong, but I would think other languages suffer from the code-driven construct issues as well.

kylekatarnls:
I think I will add the list option in the next version turning `1 hour 30 minutes 12 seconds` into `1hour, 30 minutes and 12 seconds`. So the "en" will also appears in Dutch.

Issue #1552, opened 2018-12-17
No-code tool for translation


kylekatarnls:
Add/fix translations is not an easy task for non-coders people. And it's not easy to understand if not familiar the plural syntax, the days/months order, and so on.

An online tool could provide fields to complete each missing translation, then showing some examples (the unit tests) the translator would have to validate.

A dictionary could check the spelling seems right in the given language then a translator tool could give a score to the given translation to detect basic mistakes and inversions.

The translator would provide optional GitHub account or custom name/e-mail.

The tool would create a pull-request with generated translations and unit tests using the provided author informations.

PR #1582, opened 2019-01-31
Extend Belgian Dutch (nl_BE) lang from Dutch (NL)

Propaganistas:
Belgian Dutch differs very little from regular Dutch, mostly only by pronunciation and some French influences in number and date notations.

Let's keep it tightly coupled by recursively merging only the actual differences.

### DIFF
--- a/src/Carbon/Lang/nl_BE.php
+++ b/src/Carbon/Lang/nl_BE.php
@@ -9,56 +9,8 @@
  * file that was distributed with this source code.
  */
 
-return [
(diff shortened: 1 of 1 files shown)


kylekatarnls:
This principle could be applied on some other languages groups. It could be less easy to contribute to translation and also achieve the no-code translation tool (https://github.com/briannesbitt/Carbon/issues/1552). I will merge it and do some tests and try to unify the handling of languages inheritance.

PR #1583, opened 2019-01-31
Use list join in interval dump through period dump

kylekatarnls:


kylekatarnls:
@Propaganistas I added "and" / "en" in the dumps, is it still right in Dutch too?

Propaganistas:
It's perfect. Thanks!

PR #1584, opened 2019-02-04
Adding short formats into a sk.php

jadamec:
Unfortunately Slovak language doesn't have a short format for each and every time part, but if there's an option for it why not to give an opportunity of some slangs. Here they are.


kylekatarnls:
Thanks!

PR #1585, opened 2019-02-05
Sk short units

kylekatarnls:
Add digits for short units in SK and unit tests


kylekatarnls:
@jadamec Can you check unit tests with the new short units: https://github.com/briannesbitt/Carbon/pull/1585/files#diff-5794713023423a1a93d9b187f5de50d3

Thanks.

jadamec:
@kylekatarnls everything is good except these two. Their shorter version sounds quite weird. 😃 But let's give an option to Slovakian users... Please change this two and then it's good. Thanks.

`'y' => ':count rok'` should be `'y' => ':count r'`
`'d' => ':count deň|:count dni|:count dní'` should be `'d' => ':count d'`

kylekatarnls:
We never say it should be 3 letters in every languages. For example in English, it's 1 letter (except for year) and there is no space between number and short unit.

The goal is to provide a short syntax for intervals common enough to be understood in the given language.

Are you sure it's best to have 1 letter for day and year only?

jadamec:
Great, thank you! :)

PR #1650, opened 2019-03-07
Fix addRealMicrosecond() with low float precision

cmorbitzer:
This PR fixes the error `DateTime::modify(): Failed to parse time string (@1.3961412E+9.000000) at position 9 (2): Unexpected character` when calling `addRealMicrosecond()` when the INI `precision` is set too low.

It fixes the issue by casting the result of `floor()` to an integer.


kylekatarnls:
Nice catch. Thanks.

PR #1654, opened 2019-03-09
Custom format function

kylekatarnls:
New features:
- ago() alias
- timespan() method
- createFromIsoFormat() (like createFromFormat but with format of ->isoFormat() method)
- createFromLocaleFormat() (like createFromFormat but with string in any language)
- createFromLocaleIsoFormat (like createFromFormat but with format of ->isoFormat() method and string in any language)
- locale() can now take fallback locales as rest parameters
- parseFromLocale() translate a string from a given language to English then parse it to create a date
- rawFormat() always call the native PHP DateTime format() method
- format() can now be customized at object level via settings() or globally by sub-class
- rawParse() always call the native PHP DateTime parse() method
- parse() can now be customized by sub-class
- rawCreateFromFormat() always call the native PHP DateTime createFromFormat() method
- createFromFormat () can now be customized by sub-class


kylekatarnls:
Hi @jenssegers, here is how I plan to implement cool features of https://github.com/jenssegers/date using the internal Carbon 2 translation data.

I wanted to keep the same behavior by default for format(), parse() and createFromFormat() for backward-compatibility but they now can be customized and here is how simple it can be to get an equivalent Jessengers\Date class:

https://github.com/briannesbitt/Carbon/pull/1654/files#diff-548b00f950f56564bedbfe355d3bfbcf

Your feedback is welcome.

PR #1667, opened 2019-03-20
Fix createFromFormat() to match native implementation when using test now

cmorbitzer:
The [documentation](http://php.net/manual/en/datetime.createfromformat.php) for `DateTime::createFromFormat()` states that "portions of the generated time which are not specified in format will be set to the current system time." However, this is has never been true for the microsecond portion, which is set to zero if not specified in the format.

There is an open ticket about this (https://bugs.php.net/bug.php?id=74332) but it doesn't appear that it will be fixed anytime soon. It seems reasonable to proceed as if the current behavior is the expected behavior for now.

Currently, the result of `Carbon::createFromFormat()` matches `DateTime::createFromFormat()` except when using test now, in which case `Carbon::createFromFormat()` returns a microsecond portion from the test now. This PR sets the microsecond to zero when it is not specified in the format while using test now so that the two will be aligned.

The tests are written so that they will fail if `DateTime::createFromFormat()` does begin setting the microsecond portion, at which point backwards compatibility should be addressed.

### DIFF
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -443,7 +443,7 @@ trait Creator
     }
 
     /**
-     * @param string                          $format Datetime format
+     * @param string                          $format     Datetime format
      * @param string                          $time
      * @param \DateTimeZone|string|false|null $originalTz
      *
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -442,6 +442,13 @@ trait Creator
         return static::today($tz)->setTimeFromTimeString($time);
     }
 
+    /**
+     * @param string                          $format Datetime format
+     * @param string                          $time
+     * @param \DateTimeZone|string|false|null $originalTz
+     *
+     * @return \DateTimeInterface|false
+     */
     private static function createFromFormatAndTimezone($format, $time, $originalTz)
     {
         // Work-around for https://bugs.php.net/bug.php?id=75577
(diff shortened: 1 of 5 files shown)


kylekatarnls:
Thanks, very good test.

kylekatarnls:
The test does not pass because you modify microseconds of `$this->now` but `$d_testnow` won't have them.

I suggest:
```php
Carbon::setTestNow();
$nativeDate = Carbon::createFromFormat('Y-m-d H:i:s', '1975-05-21 22:32:11');
Carbon::setTestNow(Carbon::now());
$mockedDate = Carbon::createFromFormat('Y-m-d H:i:s', '1975-05-21 22:32:11');

$this->assertSame($mockedDate->micro === 0, $nativeDate->micro === 0);
```

I think it more simply explain the behavior we expect now and for the future if the bug is fixed.

kylekatarnls:
I mistaken, it's an other test that's now broken.
https://github.com/cmorbitzer/Carbon/edit/fix-createFromFormat-microseconds/tests/Carbon/TestingAidsTest.php?pr=%2Fbriannesbitt%2FCarbon%2Fpull%2F1667#L196
https://github.com/cmorbitzer/Carbon/edit/fix-createFromFormat-microseconds/tests/CarbonImmutable/TestingAidsTest.php?pr=%2Fbriannesbitt%2FCarbon%2Fpull%2F1667#L205

The `createFromFormat` with microseconds was in fact tested. I have to consider wisely if we can really change this.

It seems relevant because testing aids shouldn't be used in real code, so it should not break anything in production for applications.

And testing aids should match the reality as close as possible. Currently, this microsecond can be a unexpected output that unit tests of users does not reveal. So it seem better to break unit tests that use it and make users aware of this bug.

cmorbitzer:
Thanks for the quick review!

Issue #1681, opened 2019-03-31
Your wish-list to Carbon


kylekatarnls:
Hello,

I have some ideas for Carbon improvements in next months but before writing the road-map I would like to get your opinion to be sure there isn't something else important to do, then to prioritize the whole list.

  - Add a composer.json extra.carbon setting to configure default settings of Carbon (locale, timezone, macros) or pointing to files for config and macros/mixins.
  - A tool (probably a CLI command `./vendor/bin/carbon`) to:
    - Look for macros/mixins then enable auto-completions for them (adding PHPDoc) so your IDE will understand where is the definition, the parameters and returned types. (This could also be called automatically on composer install.)
    - Help migration from Carbon 1 to 2 (even if you have some Laravel <= 5.7 restriction).
  - Improve documentation with a global search bar, including live examples and methods from referenced macros/mixins. Adding a troubleshooting for known PHP pitfalls (DST, month overflow, etc.)
  - More generic methods `is`, `next`, `previous` getting strings such as "sunday", "1 July" and automatically isXyz(), nextXyz() and previousXyz() if it matches the given string.
  - Improve and create more mixins libraries such as https://github.com/kylekatarnls/business-day and https://github.com/kylekatarnls/business-time
  - Improve performances (loading of translations files).
  - Improve translations (compare multiple sources to detect probable mistakes).
  - Port some methods from Carbon 2 to Carbon 1.

All of that will probably not land on Carbon in 2019, so please feel free to give me your top-3-priorities and/or if there is something else you would like to have in Carbon.

This survey is open to everyone but if you have a [Tidelift subscription](https://tidelift.com/subscription/pkg/packagist-nesbot-carbon?utm_source=packagist-nesbot-carbon&utm_medium=referral&utm_campaign=roadmap), are a [OpenCollective Sponsor](https://opencollective.com/Carbon#sponsor)/[Backer](https://opencollective.com/Carbon#backer) or a contributor to an open-source project using Carbon we would appreciate to know.

If for whatever reason you may not disclose some informations publicly, you still can let your e-mail address here to get in touch:
https://carbon.selfbuild.fr/

Thanks in advance for your participation.

GrahamCampbell:
It would be great to get a 2.27.0 release soon, or at least have the branch aliases correctly setup on master.

kylekatarnls:
2.27 is dev-master and is coming Wednesday (see https://github.com/briannesbitt/Carbon/milestone/45)

[...]

PR #1690, opened 2019-04-09
Fix floating point modulo in floatDiffInDays() when $hoursDiff is a float

pflueg:
Hi there!

I experienced an issue in the Carbon\Traits\Difference::floatDiffInDays() function when the used function floatDiffInHours() is returning a decimal part.

Carbon is using the modulo operand directly. Unfortunately PHP is stripping any decimal parts beforehand and is using the numbers only as real integers:

```
php > $hoursDiff = 48.13;
php > var_dump($hoursDiff % 24);
int(0)
```

For floating-point modulo the library needs to use the php function [fmod()](https://www.php.net/manual/en/function.fmod.php).

```
php > $hoursDiff = 48.13;
php > var_dump(fmod($hoursDiff, 24));
float(0.13)
```

In this PR i changed it, added a case to the unittests and adapted all expected values in the other assertions, that are now using the decimal part as well

Cheers o/

### DIFF
--- a/src/Carbon/Traits/Difference.php
+++ b/src/Carbon/Traits/Difference.php
@@ -454,7 +454,7 @@ trait Difference
     {
         $hoursDiff = $this->floatDiffInHours($date, $absolute);
 
-        return ($hoursDiff < 0 ? -1 : 1) * $this->diffInDays($date) + ($hoursDiff % static::HOURS_PER_DAY) / static::HOURS_PER_DAY;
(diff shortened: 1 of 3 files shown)


kylekatarnls:
Thanks for your help, it seems correct. I will compare precision after/before to be sure and check where else we use modulo. It could be in the next minor release.

pflueg:
@kylekatarnls : Do you already know when the next version is going to be released with this fix in it?

kylekatarnls:
Before the end of the week. Now that Carbon is stable enough, I plan to release 1 minor version per month on Carbon 2.x.
https://github.com/briannesbitt/Carbon/issues/1681

kylekatarnls:
2.17.0 released.

PR #1699, opened 2019-04-19
Fix #1698

### LINKED ISSUE #1698: isCurrentMonth() can not work well.
Hello,

```php
var_dump(Carbon::parse('2019-04-19')->isCurrentMonth(true));
```
Carbon version: 1.36
PHP version: 5.5

I expected to get:
```
true
```

But I actually get:
```
a PHP exception: Expected null, DateTime or DateTimeInterface, boolean given
```

Thanks!

JimChenWYU:
Fix the bug #1698 and Add some tests.


kylekatarnls:
Thanks! This will be released in 1.37.0.

Issue #1716, opened 2019-05-08
[Translation tool] ka_GE


akalongman:
|English|ka_GE|comments|
|--|--|--|
|2 hours ago|2 საათშიში|This should be "2 საათის უკან"|
|2 hours from now|2 საათის წინ|This should be "2 საათში"|
|2 hours before|2 საათი უკან|This should be "2 საათით ადრე"|
|2 hours after|2 საათი შემდეგ|This should be "2 საათის შემდეგ"|

----
Please make these changes as soon as possible

kylekatarnls:
I acknowledge this and will try to add it in the May release but I have to check it deeper, Google Translate and moment.js both give different results for those cases and ka locale is a special case:
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/ka.php

Only some words ends change according to past/future but not all (so the closures in ka.php) we have to handle this in a way that does not make things worst (for other units mostly) and we also should handle composite intervals: `2 hours and 30 minutes ago`.

It would help me if you could give me more samples with composite intervals, with 1 hour, with days.

Or if you know the exact rule and can implement it in PHP, feel free to propose a pull-request on **Lang/ka.php**.

akalongman:
@kylekatarnls first of all I'll send a PR with updated tests, which result is correct for each case (obviously they will fail for now).
Second step should be updating the ka rules and in this I need your help, because translation mechanism was changed when I updated ka rules in the library last time.

kylekatarnls:
Thanks a lot. If you can do the tests and provide some official references where we can se the same sentences, I will implement the function.

kylekatarnls:
Hi,

I have a difficulty, why do we need the `შემდეგ` suffix after `3 წლის` but not `2 საათის` what is the rule?

kylekatarnls:
I compared with i18n in closure-library to get the rules. I close this issue to continue the discussion on the PR, please review the test changes. #1719

PR #1719, opened 2019-05-10
Improve Georgian localization

akalongman:
Improve Georgian (ka & ka_GE) localization

- [x] Add valid values on Georgian for each test case
- [x] Update library according to tests

Related issue: #1716


kylekatarnls:
I checked in Google Translation (translate Gregorian to English) and all seem OK. I will implement the changes.

kylekatarnls:
As per moment.js, ka would use 12-hours cloak with AM/PM:
https://github.com/moment/moment/blob/develop/src/locale/ka.js#L21

@akalongman Is it a mistake as you used 24-hours in unit tests?

kylekatarnls:
glibc does not use neither AM/PM

akalongman:
Yes, moment.js uses 12  hours clock, but we have not AM/PM suffixes (we use "6 o'clock of evening" etc.) and for simplify decided to use 24 hours clock. 
Btw Russia also uses 12 hours clock but in the library used 24 hours clock, so I think its ok.
Would be great if we had an option for specifying which format to use

kylekatarnls:
You can still use any format inside isoFormat(), LT, LL, etc. are just shortcuts to most common formats of each language. We should not add options, it would be unmaintainable for the hundred of languages we support.

tests/Localization/KaGeTest.php
```diff
@@ -97,130 +97,130 @@ class KaGeTest extends LocalizationTestCase
         // Carbon::parse('2018-04-10 00:00:00')->isoFormat('DDDo')
         'მე-100',
         // Carbon::parse('2018-02-10 00:00:00', 'Europe/Paris')->isoFormat('h:mm a z')
-        '12:00 დილის cet',
+        'ღამის 12:00 cet',
```
kylekatarnls: This one will not be possible as the format is explicitly `h:mm a z`, order can be changed in macro-formats (LTS, LL, etc.) but not when the order is given explicitly by `isoFormat`

kylekatarnls: About ღამის, დილის, can you give me the hours range for each?

akalongman:
@kylekatarnls so, what fo you think, leave 24 hours clock or not? For us - Georgians its ok, because we use 24 hours format in everyday life.

tests/Localization/KaGeTest.php
akalongman: I seen that example in Ru test. Is it mistake as well?

akalongman: ღამის 22:00 - 04:00
დილის 04:00 - 11:00
შუადღის/დღის 11:00 - 16:00
საღამოს 16:00 - 22:00

I'll try to find some official docs about these time ranges

kylekatarnls:
24-hours is OK for LT, LTS, LLL and LLLL formats but, we still need to implement AM/PM equivalent for users that will explicitly call `->isoFormat('h:mm a z')` is such case, we give the user what he ask whatever the language.

kylekatarnls:
I added English matching sentences in unit tests please review them. I will wait a few days so you can check it, then I'll release it in 2.19.1.

akalongman:
I'll take a look in next few days

kylekatarnls:
I still have a problem about weekdays (broken tests), when to use `კვირა` and when to use `კვირას`?

akalongman:
@kylekatarnls 
კვირა is a Sunday (Week as well depending on context), 
კვირას is a On Sunday, when you speak about some action on Sunday, e.g. "I am going abroad on Sunday"
Do you need more examples?

kylekatarnls:
I think I get it. I added the following rule: asking for a day alone კვირა, a day in sentence/date : კვირას.

kylekatarnls:
I merge this one for 2.19.1. It's as good as I can trying to reach your proposition keeping the whole thing working in an automated way. I replaced RegExp with explicit replacements in closures in ka.php so maybe you can check it more easily and propose a PR if you see a mistake.

PR #1757, opened 2019-06-19
 Change toIso8601String to use Iso8601

jycr753:
- The formats are not the same, and Carbon should return format that is expected

### DIFF
--- a/src/Carbon/Traits/Converter.php
+++ b/src/Carbon/Traits/Converter.php
@@ -253,7 +253,7 @@ trait Converter
      */
     public function toIso8601String()
     {
-        return $this->toAtomString();
+        return $this->rawFormat(DateTime::ISO8601);
     }
 
     /**


kylekatarnls:
The ISO 8601 norm says:
> timezone [...] is therefore represented as "09:30Z" or "0930Z"
So both `date('O')` (like in `DateTimeInterface::ISO8601`) and `date('P')` (like in `DateTimeInterface::ATOM`) are valid regarding the norm.

`DateTimeInterface::ISO8601` is the format PHP choose to implement the norm among different possibilities. If you want this specific implementation, use `->format(DateTimeInterface::ISO8601)`. If you want an ISO 8601 format `->toIso8601String()` is compliant so we should not change it (would be a breaking change for something that is not wrong).

jycr753:
Thanks for the replied, that makes sense :)

PR #1767, opened 2019-07-03
Small grammatical changes to macros documentation

andrewnclark:
I use Carbon too often to give nothing in return so I made some small grammatical changes to the macros documentation. I’ll continue to take a look at other areas as I find a spare hour here and there. 

I like to document when my I've wrote enough code for one day!


kylekatarnls:
Thanks a lot!

Issue #1776, opened 2019-07-06
[Laravel] Service Provider performance


barryvdh:
Hello,

So I'm probably nitpicking here about 0.5ms, but anyways, Re: https://github.com/briannesbitt/Carbon/pull/1363

In light of checking the Laravel boot performance, I was checking which service providers are making the biggest impact on the boot time for a 'hello world' request. The Carbon ServiceProvider stood out because it took more time then other providers. (And because it is required by Laravel itself, but auto-registered, it's easy to miss because it's not there with the rest).
I'm not sure what `setLocale` actually does, but it seems to create a Translator instance and load the message from disk.  In regards with performance on an optimized config (see https://twitter.com/barryvdh/status/1147547152366874628) I get +/- 3ms to reach the Controller. If I put Carbon in the `dont-discover` list, it drops about 0.5ms. Which obviously isn't a lot an absolute terms, but it's still about 16% of the entire boot of Laravel.

Removing the ServiceProvider now would obviously break lot's of projects, but does the `setLocale` need to be this heavy? Am I doing something wrong (seeing the result on opcached production, but also on Valet+)? 

Possible solutions:
 - Defer the Translator/Message until it's actually used (it's just for formatLocalized / diffInHumans, right?) Perhaps set the Locale as a variable on the factory and use that when creating the translator when it's required.
- Cache the messages using Laravel (eg. in bootstrap/cache/carbon.php or storage/framework/carbon)

Issue Tempate:
I encountered an issue with the following code:
```php
    public function updateLocale()
    {
        $translator = $this->app['translator'];

        if ($translator instanceof Translator || $translator instanceof IlluminateTranslator) {
            Carbon::setLocale($translator->getLocale());
        }
    }
```

Carbon version: 2.20.0
PHP version: 7.2.8 / 7.3.6

I expected to get: No performance penalty for booting Laravel
But I actually get: 0.5ms 'delay' when using the ServiceProvider

kylekatarnls:
> it's just for formatLocalized / diffInHumans, right?

Nope, it's a lot more methods (all week methods for example will use first_day_of_week and day_of_first_week_of_year of the locale), getters for month and week days names (short and long), isoFormat, parseFromLocale, translateTimeString, it's used by CarbonInterval and CarbonPeriod. So intercepting all this function to lazy load the messages is not really an option.

> Cache the messages using Laravel

Not an option neither as we don't know in advance the locale (it can be any language with any region among hundreds).

[...]

PR #1777, opened 2019-07-06
[Laravel] Defer Translator + Simplify Provider

barryvdh:
As suggested in #https://github.com/briannesbitt/Carbon/issues/1776 , maybe we can wait to load the actual Translator until it's used?

I'm also not sure why we need the Laravel Translator or the Event to get the current Locale. The getLocale() is available from the App container already.

### DIFF
--- a/src/Carbon/Laravel/ServiceProvider.php
+++ b/src/Carbon/Laravel/ServiceProvider.php
@@ -13,7 +13,9 @@ class ServiceProvider extends \Illuminate\Support\ServiceProvider
 {
     public function boot()
     {
-        if (!$this->app->bound('events') || !$this->app->bound('translator')) {
+        $this->updateLocale();
+
+        if (!$this->app->bound('events')) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


barryvdh:
Also I should probably note that I have not extensible tested it, but 'it works on my device' ;)

kylekatarnls:
The locale is not a constante, it can change using `App::setLocale('fr');` so we have to synchronize it on the event for locale change.

barryvdh:
Yes I understand the sync, but if you don't have the event dispatcher (for some reason, but mostly you have it anyways?) You can still get the current locale. Currently you are bailing out before setting it. And the translator doesn't have to be loaded just to get the Locale.

barryvdh:
You are right about the benchmark/micro optimization. I think it would be required to test further to see what real life would change. It's a bit hard to test with opcache and all.  It seems to switch a bit all the time, but it *seems* like it's a bit faster for non-localized requests and the same for localized requests. But need to find a good benchmark still.

```
setDefaultLocale:
string(17) "setDefaultLocale: 0.47ms"
string(38) "Now: Sat Jul 06 2019 22:49:54 GMT+0200"
string(13) "Boot: 1.483ms"
string(18) "Diff: 1 dag eerder"
string(23) "Localized boot: 2.018ms"

Current
string(18) "setLocale: 1.057ms"
string(38) "Now: za jul. 06 2019 22:50:15 GMT+0200"
string(13) "Boot: 1.796ms"
string(18) "Diff: 1 dag eerder"
string(23) "Localized boot: 2.076ms"
```

barryvdh:
If you want I can create a PR for just the Laravel changes (eg. use the app locale directly instead of the translator)

kylekatarnls:
I extracted it: #1779

Tests pass and it still work in my Laravel app.

Next step will be to make getLocale and setLocale to not trigger the lazy-load of the translator.

So I close this, but feel free to comment the other PR and propose the optimization you put in `setDefaultLocale` but directly in `setLocale` so everyone will benefit of it, not only Laravel users.

barryvdh:
Thanks. The reason I created a new method was that I don't know the return value yet.

PR #1787, opened 2019-07-13
Correct day names in  Pashto

spinzar:
correct days and months name in  pashto

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -47,5 +47,8 @@
         <testsuite name="jenssegers">
             <directory>tests/Jenssegers</directory>
         </testsuite>
+        <testsuite name="cli">
+            <directory>tests/Cli</directory>
+        </testsuite>
     </testsuites>
 </phpunit>


kylekatarnls:
Hi, I don't see matching for those words. Can you provide a source/reference for those translations?

spinzar:
Hi,thanks for asking this question, Original Pashto language calendar have its own Zodiac  Signs and Names in Pashto ,  same like chines Zodiac  years.
below i am sharing with you a  five links  to conform it.

https://en.wikipedia.org/wiki/Solar_Hijri_calendar
http://www.whatiscalled.com/zodiac-signs-in-pashto/
https://thepashto.com/numbers.php
https://www.revolvy.com/folder/Pashto-names-for-the-months-of-the-Solar-Hijri-.../243571
https://polyglotclub.com/wiki/Language/Central-pashto/Vocabulary/Days-of-the-Week
https://www.wikizero.com/en/Solar_Hijri_calendar
https://www.safiahaleem.com/?p=3335
https://en.wikipedia.org/wiki/La%E1%B9%9Bam

kylekatarnls:
Hi @spinzar sadly, changing month names is not enough to support the Hijri calendar as dates are different.

As you can see in other languages, such as [Arabic](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/ar.php#L19) we expect month names to always return Gregorian months no matter the language and region. Moreover, note than calendar rather depends on the region than the language. I mean you can speak a given languages in multiple countries using different calendars.

In our case, if we just change month names to Hijri ones, we will have wrong dates when running:

```php
Carbon::now()->locale('ps')->isoFormat('d mmmm');
```

This displays the day number and the month (according to Gregorian calendar); if we calculate month/day numbers with Gregorian calendar then display Hijri names, we will have inconsistent dates such as `31 مرغومی` (when we are January the 31st) while we would expect `25 تله` (and that would of course change every year).

Non-Gregorian calendars support is not planned. There are many, It would need a support in all of our languages, that would be a big amount of additional maintenance and right now we don't have enough ressources for this.

spinzar:
i understand your logic about months name and it is clear.
but what about correction of weekdays name  you didn't answered it. 

'weekdays' => ['اتوار', 'ګل', 'نهه', 'شورو', 'زيارت', 'جمعه', 'خالي'],
    'weekdays_short' => ['ا', 'ګ', 'ن', 'ش', 'ز', 'ج', 'خ'],
    'weekdays_min' => ['ا', 'ګ', 'ن', 'ش', 'ز', 'ج', 'خ'],

kylekatarnls:
Hi @spinzar, days should start with `Sunday` in the array so the index will match `date('w')`, then the start of week is changed on need accordingly to `first_day_of_week`, your list starts with Saturday, so days will be shifted in output.

spinzar:
Oh, they also have a mistake for Sunday ?
i saw it. 
correct ward is ''اتوار ''	Atwar
not  ''اتور ''

kylekatarnls:
OK, I rewrote it in one commit with unit tests #1799

spinzar:
https://thepashto.com/numbers.php
have correct names of days in Pashto.


and polyglotclub.com have 3 uncorrect names for Sunday , Monday, and Wednesday.
https://polyglotclub.com/wiki/Language/Central-pashto/Vocabulary/Days-of-the-Week

kylekatarnls:
thepastho.com site name feels trustable :D

Issue #1796, opened 2019-07-15
Add comparison methods to CarbonInterval


kylekatarnls:
Using `totalMicroseconds` methods to compare 2 intervals could be added:
- greaterThan / gt
- greaterThanOrEqualTo / gte
- lessThan / lt
- lessThanOrEqualTo / lte

kylekatarnls:
Ready for 2.22.0 #1801

Thank you to @burobo

PR #1801, opened 2019-07-16
Add comparison methods to CarbonInterval.

burobo:
#1796


src/Carbon/CarbonInterval.php
```diff
@@ -1706,4 +1706,116 @@ public function equalTo($interval): bool
...
+     *
+     * @return bool
+     */
+    public function greaterThan($interval): bool
+    {
+        if (!($interval instanceof self)) {
+            $interval = self::make($interval);
+        }
+
+        return $interval !== null && $this->totalMicroseconds > $interval->totalMicroseconds;
```
kylekatarnls: @burobo I would say if `$interval` is `null`, it's like 0 seconds, so `$this` is always "greater". For example `var_dump((new DateTime()) > null);` is `true` in PHP and `var_dump((new DateTime()) < null);`. It is also more consistent if `greaterThan` is `true` when `lessThan` is `false`.

tests/CarbonInterval/ComparisonTest.php
```diff
@@ -37,4 +37,100 @@ public function testEqualToFalse()
         $this->assertFalse($oneDay->eq(CarbonInterval::hours(24)->microsecond(1)));
         $this->assertFalse($oneDay->eq(CarbonInterval::hours(23)->minutes(59)->seconds(59)->microseconds(999999)));
     }
+
```
kylekatarnls: @burobo Add assertions for the case `self::make()` returned `null` (so `$interval = null`), for example:
```php
$this->assertFalse($oneDay->gt('xxx'));
```

src/Carbon/CarbonInterval.php
burobo: I get it. I'll fix it soon. Thank you!

PR #1869, opened 2019-08-26
Fix #1849 Slovenian unit declensions

### LINKED ISSUE #1849: [Translation tool] sl_SI
|English|sl_SI|comments|
|--|--|--|
|1 hour|1 uro|This should be "1 ura". If you want to say "1 hour till arrival" then you would use "1 ura", but if you want to say "This took me 1 hour to complete" then "1 uro" is correct. But I think mostly the first example applies so "1 ura" would be better...|
|an hour|1 uro|This should be "1 ura". If you want to say "1 hour till arrival" then you would use "1 ura", but if you want to say "This took me 1 hour to complete" then "1 uro" is correct. But I think mostly the first example applies so "1 ura" would be better...|
|1 minute|1 minuto|This is the same as with the hours. If you say "Only 1 minute left" then you use "1 minuta", but if you say "We're going in 1 minute" then you use "1 minuto".|
|a minute|1 minuto|This is the same as with the hours. If you say "Only 1 minute left" then you use "1 minuta", but if you say "We're going in 1 minute" then you use "1 minuto".|
|1 second|1 sekundo|The same as hours and minutes|
|a few seconds|1 sekundo|The same as hours and minutes|
|2 hours before|pred 2 uri|This should be "2 uri pred" if you want to use it in a sentence like "2 hours before the accident", but if you just say "2 hours before" then you use "pred 2 urama" (but it would be even better to use a word instead of number "pred dvema urama") Because grammar...
|2 hours after|čez 2 uri|This should be "2 uri po" if you want to use it in a sentence like "2 hours after the accident", but if you just say "2 hours after" then you use "po 2 urah" (but even prettier would be to use a word "po dveh urah") Because grammar...

morpheus7CS:
Submitting fixes in response to Issue [#1849](https://github.com/briannesbitt/Carbon/issues/1849).

Kind regards,
g


kylekatarnls:
Thanks a lot!

kylekatarnls:
Testable with [2.24.0-beta.3](https://github.com/briannesbitt/Carbon/releases/tag/2.24.0-beta.3)

It will be released next week in 2.24.0.

PR #1874, opened 2019-08-29
Enhance guide for static macros

spawnia:
Clarify the difference between instance macros and static macros and add an example.

### DIFF
--- a/docs/index.src.html
+++ b/docs/index.src.html
@@ -3102,11 +3102,12 @@ Carbon::serializeUsing(null);
 
     <p>
     <pre class="live-editor"><code class="php">{{::lint(
-Carbon::macro('diffFromYear', function ($year, $absolute = false, $short = false, $parts = 1) {
-    return $this->diffForHumans(Carbon::create($year, 1, 1, 0, 0, 0), $absolute, $short, $parts);
+Carbon::macro('diffFromYear', static function ($year, $absolute = false, $short = false, $parts = 1) {
+    return self::this()->diffForHumans(Carbon::create($year, 1, 1, 0, 0, 0), $absolute, $short, $parts);
 });
 )}}
 
+// Can be called on Carbon instances: self::this() = current instance ($this)
 {{::exec(echo Carbon::parse('2020-01-12 12:00:00')->diffFromYear(2019);/*pad(78)*/)}} // {{eval}}
 echo "\n";
 {{::exec(echo Carbon::parse('2020-01-12 12:00:00')->diffFromYear(2019, true);/*pad(78)*/)}} // {{eval}}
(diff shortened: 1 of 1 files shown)


docs/index.src.html
```diff
@@ -3092,12 +3092,13 @@ <h1 id="api-json">JSON</h1>
...
-            href="https://laravel.com/docs/5.6/responses#response-macros">response</a> or <a
-            href="https://laravel.com/docs/5.6/collections#extending-collections">collections</a>. The Carbon <code>macro()</code>
-        works the exact same way as the Laravel MacroableTrait one, it takes a method name as first argument and a
-        closure as second. This will make the closure action available on all Carbon instances (and also as a Carbon static
-        method) as a method with the given name.</p>
+    <p>You may be familiar with the macro concept if you are used to working with Laravel and objects such as
+        <a href="https://laravel.com/docs/responses#response-macros">response</a>
+        or <a href="https://laravel.com/docs/collections#extending-collections">collections</a>.
+        Carbon macros works just like the Laravel <code>Macroable</code> Trait.
+
```
kylekatarnls: Browser will remove new lines. Use `</p><p>` to create new paragraphs.

docs/index.src.html
```diff
@@ -3092,12 +3092,13 @@ <h1 id="api-json">JSON</h1>
 
     <h1 id="api-macro">Macro</h1>
 
-    <p>You may be familiar with the macro concept if you are used to working with Laravel and objects such as <a
-            href="https://laravel.com/docs/5.6/responses#response-macros">response</a> or <a
-            href="https://laravel.com/docs/5.6/collections#extending-collections">collections</a>. The Carbon <code>macro()</code>
-        works the exact same way as the Laravel MacroableTrait one, it takes a method name as first argument and a
-        closure as second. This will make the closure action available on all Carbon instances (and also as a Carbon static
-        method) as a method with the given name.</p>
```
kylekatarnls: We should keep "as a method with the given name".

About static method, it can be interesting to create macros that are both static and dynamic, like:
```php
Carbon::macro('doubleYear', function () {
  if (isset($this)) {
    return $this->year * 2; // If called dynamically
  }

  return Carbon::now()->year * 2; // If called statically
});

Carbon::parse('1990')->doubleYear(); // 3980
Carbon::doubleYear(); // 4038
```

So I'm not sure we should un-document it. For me the `static function` is just a way to make it more explicit (and help static analyzers) but every macros (with the `static` keyword or not) will always be available both dynamically and statically until Laravel would decide to change it making the `static` keyword mandatory and using 2 different variables for `__call` available macros and `__callStatic` available macros. For now, Laravel and Carbon both use a single common variable.

kylekatarnls:
I will pause this change until #1876 is resolved. As some PHP 8 future changes will require to adjust the way macros are declared, we will have to propose some recommandations for PHP 8 compatibility. Once I get a working solution, I will update this part of the doc.

kylekatarnls:
Hi @spawnia what do you think about the new `self::this()` feature and advertising users to use it instead of `$this` so every macro/mixin closures can be marked as `static` and called on both instance and statically (and properly autocompleted by the IDE if using the CLI)?

spawnia:
I like the idea of making the macros work consistently and improving IDE support.

However, i think some Macro`s don't make sense dynamically, for example something that is similar to `createFromFormat`. Will this new style be able to accomodate that?

kylekatarnls:
@spawnia about "don't make sense dynamically", the possibility to call dynamically static method is not a macro specificity, it's the native behavior of PHP: https://3v4l.org/qZLVn

And IDEs will suggest static methods on instances.

So the previous state is:
- You have dynamic macros/methods that can be used statically or not (you have no way to know as there is no convention). You have to check the code to see if it handle the possibility that `$this` may not be set.
- You have static macros/methods: the explicit `static` keyword ensure you that this can be called on both static class or instances.

The next state is:
- Following the convention, you can't have dynamic macros/methods, and you can't use `$this`
- But your static macros can get the context in a safe way that ensure you get a date even if you call it statically.

If we would like to disallow macros to be called dynamically or statically, we would need to see how it could be consistent with other macros in Laravel. Then we would need dedicated options/methods to add those restricted macros. The `static` keyword would not be enough.

Then we would have to be sure it's actually still needed. Since we provide a default date for static macros, what would be the case when we would forbid to call them statically? And for pure static macros, what would be the reason to disallow to call it on instances?

Such constraints could also be easily handled on the runtime from inside the macro:
https://3v4l.org/82206

I would provide an easier way to strictly enforce those constraints and document it if we found concrete cases where it actually make sense.

spawnia:
You are right, PHP does not allow to constrain methods to be called only statically. That is not limited to Macros, so we can not really do anything about it.

Let's go ahead with what is there now, it is a definite improvement and step in the right direction.

PR #1883, opened 2019-09-09
Fix isBefore documentation

bouchardm:
There was an error in the documentation of the function isBefore.

2018-07-25 12:45:16 is before 2018-07-25 12:45:18.


kylekatarnls:
Thanks!

PR #1903, opened 2019-10-07
Use CarbonInterface instead of $this in PHPDoc

krzysztofrewak:
Should fix problem with hints in PHPStorm.

carboninterface


kylekatarnls:
Hi,

Thanks for your help.

From line https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php#L19 `<autodoc generated by `composer phpdoc`>` to `</autodoc>`

The PHPDoc is generated using https://github.com/briannesbitt/Carbon/blob/master/phpdoc.php

So if you modify it manually, it will be erased on next run (at release for example).

So this change needs to be written as rules in phpdoc.php but first I need to check if this change would not break other IDE, other syntaxes or other static analyzers.

kylekatarnls:
I'm OK for the change as `$this` currently in use is not good in `CarbonImmutable`. `CarbonInterface` will not be good in the `Date` trait but this is a minor drawback compared to the advantage.

I close this in favor of #1904 that will fix all the classes (so you can chain more than 2 methods and will also be correct with `CarbonImmutable`) and use the script so PHPDoc is aligned and changes will be kept synchronized.

Namoshek:
I notices this issue as well and it has actually been fixed in PhpStorm 2019.3 EAP apprently. Possibly as a side-effect of [this bugfix](https://youtrack.jetbrains.com/issue/WI-28091). Appreciate the change though, the false-positives drove me crazy. 😄

kylekatarnls:
I have PHPStorm 2019.3.2 and it has the bug, it's more likely the fix provided in Carbon 2.25.2 that fixed it if you updated your composer dependencies today.

Namoshek:
Nope, I've not had the issue since early last week when I updated to the 19.3 EAP. Weird.

PR #1909, opened 2019-10-14
Fix Return Type Annotation

Bellardia:
There seems to be a bug in the return type annotation for `validateCurrentDate`.

This bug crashes `Psalm`.


kylekatarnls:
Hi, I will release a patch for this, but note that this is fixed in Carbon 2 (since May 2018). Please consider upgrading to benefit it, many improvement and long support.

PR #1915, opened 2019-10-21
Fix typehint for static create methods (Fix #1914)

### LINKED ISSUE #1914: Method not found in static
Hello,

```php
Carbon::createFromFormat('d.m.Y', $date)->setTime(0, 0, 0);
```

Carbon version: 2.25.2

PHP version: 7.3.7

While the code executes just fine, I am getting the following error in PHPStorm:
image

It looks like the PHPDoc blocks are incorrect as they state the following:

```php
/**
 * @method        static static createFromFormat(string $format, string $time, string|\DateTimeZone $timezone = null)
```

Which I guess indicates that the createFromFormat() returns a static instance (if that's even a thing). Changing it to return `CarbonInterface` instead seems to resolve this issue.

Thanks!

(I guess I could have made a PR, but I wanted to allow some discussion beforehand).

spaceemotion:
I tried to use `self` but somehow it didn't replace the hints properly.


kylekatarnls:
I guess we would have needed to replace the whole `'static self'` as they are grouped, but it will be just fine this way.

Thanks!

Issue #1918, opened 2019-10-24
New createFromFormat for CarbonInterval


xfudox:
This would be a proposal, not an issue, but i wasn't sure on where to put it.

The proposal is to implement a `createFromFormat($format, $time)` constructor for `CarbonInterval` as `Carbon` object already has, this to better handle MySQL (and others DBMS) `time` data type or other string representations of time intervals.

It is tedious to use a long

```php
Carboninterval::create(0, 0, 0, 0, 1, 30, 0, 0)
```

or even a

```php
Carboninterval::hours(1)->minutes(30)
```

since i have to split the original string into several variables, and eventyally calculate days, weeks and so on.

It would be more readable and user friendly something like 

```php
Carboninterval::createFromFormat('H:i', '1:30')
```

kylekatarnls:
Hi, this sounds very interesting and pretty simple.

I can already give you a macro that would do exactly that:

```php
CarbonInterval::macro('createFromFormat', static function (string $format, string $interval) {
  return Carbon::createFromFormat('!'.ltrim($format, '!'), $interval, 'UTC')->diffAsCarbonInterval(
    Carbon::parse('1970-01-01 00:00:00', 'UTC')
  );
});

echo Carboninterval::createFromFormat('H:i', '1:30');
```

If you want to propose it as a core method of CarbonInterval, I would release it in next minor release.

kylekatarnls:
Will be released in 2.27.0

PR #1921, opened 2019-10-29
Add createFromFormat static method to CarbonInterval

xfudox:
Added `createFromFormat` method to `CarbonInterval`'s core for #1918 .

Useful to handle time interval strings like MySQL `time` field (e.g.: '10:20:00'):
E.g.:
```php
CarbonInterval::createFromFormat('H:i:s', '10:20:00')
```

is equivalent to:
```php
list($hours, $minutes, $seconds) = explode(':', '10:20:00');
new CarbonInterval(0, 0, 0, 0, $hours, $minutes, $seconds);
```
or
```php
CarbonInterval::hours($hours)->minutes($minutes)->seconds($seconds);
```

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -484,13 +484,13 @@ class CarbonInterval extends DateInterval
 
         if (preg_match('/s([,.])([uv])$/', $format, $match)) {
             $interval = explode($match[1], $interval);
-            $i = count($interval) - 1;
-            $interval[$i] = str_pad($interval[$i], $match[2] === 'v' ? 3 : 6, '0');
+            $index = count($interval) - 1;
+            $interval[$index] = str_pad($interval[$index], $match[2] === 'v' ? 3 : 6, '0');
             $interval = implode($match[1], $interval);
         }
 
-        for ($i = 0; $i < $length; $i++) {
-            $expected = mb_substr($format, $i, 1);
+        for ($index = 0; $index < $length; $index++) {
+            $expected = mb_substr($format, $index, 1);
             $nextCharacter = mb_substr($interval, 0, 1);
             $unit = static::$formats[$expected] ?? null;
 
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -11,6 +11,7 @@
 namespace Carbon;
 
 use BadMethodCallException;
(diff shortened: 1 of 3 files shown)


kylekatarnls:
Hi @xfudox, thanks for your PR, would you mind please adding unit tests for your feature.

See as an example:
`make` test:
https://github.com/briannesbitt/Carbon/blob/master/tests/CarbonInterval/ConstructTest.php#L283
or `__construct` test:
https://github.com/briannesbitt/Carbon/blob/master/tests/CarbonInterval/ConstructTest.php#L21

Thanks

src/Carbon/CarbonInterval.php
```diff
@@ -436,6 +436,21 @@ public static function create($years = 1, $months = null, $weeks = null, $days =
         return new static($years, $months, $weeks, $days, $hours, $minutes, $seconds, $microseconds);
     }
 
+    /**
+     * Parse a string into a new CarbonInterval object according to the specified format.
+     *
+     * @param string $format
+     * @param string $interval
+     *
+     * @return void
```
kylekatarnls: `@return static`

This method is not void.

And I suggest adding parameters description and an example (will appear in the API reference of the documentation):
```php
    /**
     * Parse a string into a new CarbonInterval object according to the specified format.
     *
     * @example
     * ```
     * echo Carboninterval::createFromFormat('H:i', '1:30');
     * ```
     *
     * @param string $format  Format of the $interval input string
     * @param string $interval Input string to convert into an interval
     *
     * @return static
     */

kylekatarnls:
Hi @xfudox I had no time to fix the PHPDoc and add unit tests, so I postponed this for version 2.27. Thanks for your understanding.

xfudox:
Yeah, sorry for I'm late, I've very little free time to work on this recently, hope to fix this asap.

kylekatarnls:
Hi @xfudox thanks for unit tests and doc. The unit tests are not passing in Travis. Can you have a look?

xfudox:
Yeah @kylekatarnls , I'll take a look asap.

kylekatarnls:
Obviously, my solution will only work with hours, not for day/month/year, so we actually need to parse the format in a custom way. I take the assignee to propose something else.

kylekatarnls:
Hi @xfudox can you check the new implementation?

Issue #559, opened 2015-12-17
diffInDays and diffForHumans not returning expected result


inctor:
Hey,

I encountered that if the "diffInDays()" returns a value between 7 and 14, in this case 13 days, the "diffForHumans()" returns 1 week, and not 2 weeks, whereas it's closer to 2 weeks, than 1 week?

I recall it being able to handle this properly, can't 100% confirm it though.

Is it possible to achieve a similar result as the one in the last comment, the "extended" diffForHumans(), and still keeping it as dynamic, as it is now?

```
[3] > $now = Carbon\Carbon::now()->addDays(13); $now->diffForHumans();
// '1 week from now' // Actual is 1 week and 6 days
[4] > $now = Carbon\Carbon::now()->addDays(14); $now->diffForHumans();
// '2 weeks from now' // Actual is 2 weeks
[5] > $now = Carbon\Carbon::now()->addDays(15); $now->diffForHumans();
// '2 weeks from now' // Actual is 2 weeks and 1 day
```

wmather:
This is a big thing for me too. I see this type of thing all over the internet (Github, Facebook) where their timestamps round down or up too early or late.

In my circumstance, I want to show the connecting time between the arrival and departure of two flights. `diffForHumans()` always shows the highest unit, so 1h 55m shows as 1 hour, but this is misleading and inaccurate. It would be a very useful feature to extend this method to show the *exact* difference.

wmather:
I've found that the solution would be to use CarbonInterval. What I really want to do is this, but creating a CarbonInterval instance from a DateInterval instance that came from Carbon is specifically denied: https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/CarbonInterval.php#L230

    $diffInterval = $date->diff($date2); //Instance of DateInterval

    echo Carbon\CarbonInterval::instance($date)->forHumans;

SteveEdson:
This is a problem for me as well. It would be great if there was a new method, or a parameter within an existing method to increase the granularity of the diff to include minutes, or round to the nearest hour, day, week etc.

rimace:
So what would be the definition for this? I'd say:
If the next lower unit is bigger than zero, also print that value. So for the first example the method would print "1 week and 6 days from now". Am I right?
[...]

PR #1930, opened 2019-11-13
Add option for rounding in forHumans

tylercollier:
Hi. Thanks for your suggestion to submit a PR [here](https://github.com/briannesbitt/Carbon/issues/559#issuecomment-552608654). This is a rough draft and I'm open to feedback. Thanks!

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -1371,10 +1371,28 @@ class CarbonInterval extends DateInterval
             return $this->translate($unitData['unit'], $interpolations, $count, $translator, $altNumbers);
         };
 
-        foreach ($diffIntervalArray as $diffIntervalData) {
+        foreach ($diffIntervalArray as $index => $diffIntervalData) {
             if ($diffIntervalData['value'] > 0) {
                 $unit = $short ? $diffIntervalData['unitShort'] : $diffIntervalData['unit'];
                 $count = $diffIntervalData['value'];
+                if ($options & CarbonInterface::ROUND && count($interval) === $parts - 1 && $index < count($diffIntervalArray) - 1) {
+                    $roundValue = 0;
+                    $nextIntervalData = $diffIntervalArray[$index + 1];
+                    if ($nextIntervalData['unit'] === 'month') {
+                        $roundValue = $nextIntervalData['value'] / 12;
+                    } elseif ($nextIntervalData['unit'] === 'week') {
+                        $roundValue = $nextIntervalData['value'] / (365 / 12 / 7);
+                    } elseif ($nextIntervalData['unit'] === 'day') {
+                        $roundValue = $nextIntervalData['value'] / 7;
+                    } elseif ($nextIntervalData['unit'] === 'hour') {
+                        $roundValue = $nextIntervalData['value'] / 24;
+                    } elseif ($nextIntervalData['unit'] === 'minute') {
+                        $roundValue = $nextIntervalData['value'] / 60;
+                    } elseif ($nextIntervalData['unit'] === 'second') {
+                        $roundValue = $nextIntervalData['value'] / 60;
+                    }
+                    $diffIntervalData['value'] += round($roundValue);
+                }
                 $interval[] = $transChoice($short, $diffIntervalData);
             } elseif ($options & CarbonInterface::SEQUENTIAL_PARTS_ONLY && count($interval) > 0) {
                 break;
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -1269,6 +1269,26 @@ class CarbonInterval extends DateInterval
         return [$syntax, $short, $parts, $options, $join, $aUnit, $altNumbers, $interpolations];
     }
 
+    protected static function getRoundFactor(string $unit): ?float
+    {
+        switch ($unit) {
+            case 'month':
+                return static::getFactor('months', 'years') ?: Carbon::MONTHS_PER_YEAR;
+            case 'week':
+                return static::getFactor('weeks', 'months') ?: Carbon::WEEKS_PER_MONTH;
+            case 'day':
+                return static::getDaysPerWeek();
+            case 'hour':
+                return static::getHoursPerDay();
+            case 'minute':
+                return static::getMinutesPerHour();
+            case 'second':
+                return static::getSecondsPerMinute();
+        }
+
+        return null;
+    }
+
     /**
      * Get the current interval in a human readable format in the current locale.
      *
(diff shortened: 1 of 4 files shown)


src/Carbon/CarbonInterval.php
```diff
@@ -1371,10 +1371,28 @@ public function forHumans($syntax = null, $short = false, $parts = -1, $options
             return $this->translate($unitData['unit'], $interpolations, $count, $translator, $altNumbers);
         };
 
-        foreach ($diffIntervalArray as $diffIntervalData) {
+        foreach ($diffIntervalArray as $index => $diffIntervalData) {
             if ($diffIntervalData['value'] > 0) {
                 $unit = $short ? $diffIntervalData['unitShort'] : $diffIntervalData['unit'];
                 $count = $diffIntervalData['value'];
+                if ($options & CarbonInterface::ROUND && count($interval) === $parts - 1 && $index < count($diffIntervalArray) - 1) {
```
kylekatarnls: What is `$index < count($diffIntervalArray) - 1`?

src/Carbon/CarbonInterval.php
```diff
@@ -1371,10 +1371,28 @@ public function forHumans($syntax = null, $short = false, $parts = -1, $options
...
+                        $roundValue = $nextIntervalData['value'] / (365 / 12 / 7);
+                    } elseif ($nextIntervalData['unit'] === 'day') {
+                        $roundValue = $nextIntervalData['value'] / 7;
+                    } elseif ($nextIntervalData['unit'] === 'hour') {
+                        $roundValue = $nextIntervalData['value'] / 24;
+                    } elseif ($nextIntervalData['unit'] === 'minute') {
+                        $roundValue = $nextIntervalData['value'] / 60;
+                    } elseif ($nextIntervalData['unit'] === 'second') {
+                        $roundValue = $nextIntervalData['value'] / 60;
+                    }
```
kylekatarnls: I guess we could have an array and only 1 if:
```php
$dividers = [
    'month' => 12,
    'week' => 365 / 12 / 7,
    'day' => 7,
    'hour' => 24,
    'minute' => 60,
    'second' => 60,
];

$divider = $dividers[$nextIntervalData['unit']] ?? null;

if ($divider) {
    $roundValue = $nextIntervalData['value'] / $divider;
}
```

tests/CarbonInterval/ForHumansTest.php
```diff
@@ -245,4 +245,74 @@ public function testOptionsAsArray()
...
+        $interval = CarbonInterval::days(2)->hours(23);
+        $this->assertEquals('3 days', $interval->forHumans(['parts' => 1, 'options' => CarbonInterface::ROUND]));
+    }
+
+    public function testRoundDown()
+    {
+        CarbonInterval::setLocale('en');
+        $interval = CarbonInterval::days(2)->hours(11);
+        $this->assertEquals('2 days', $interval->forHumans(['parts' => 1, 'options' => CarbonInterface::ROUND]));
+    }
```
kylekatarnls: Can we have a round down and up for each unit, so we check the dividers are correct and don't change later by accident.

Plus, it will fulfill the coverage: https://codecov.io/gh/briannesbitt/Carbon/compare/e01ecc0b71168febb52ae1fdc1cfcc95428e604e...898f3582ab324874f48bbf5a8f60a97e4e2c755b/diff

Mostly I'm interesting in seeing what rounding happens when I have a month+week interval or a year+week interval.

src/Carbon/CarbonInterval.php
kylekatarnls: Moreover, we should use `static::getCascadeFactors()` so it takes the customization into account. and don't use magic numbers.
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/CarbonInterval.php#L236

This part is a bit tricky as cascade factors may skip a unit (for example day-to-month skipping week) so this enhancement may be for an other pull request.

tests/CarbonInterval/ForHumansTest.php
```diff
@@ -245,4 +245,74 @@ public function testOptionsAsArray()
...
+        CarbonInterval::setLocale('en');
+        $interval = CarbonInterval::days(2)->minutes(59);
+        $this->assertEquals('2 days', $interval->forHumans(['parts' => 1, 'options' => CarbonInterface::ROUND]));
+    }
+
+    public function testRoundMultipleParts()
+    {
+        CarbonInterval::setLocale('en');
+        $interval = CarbonInterval::days(2)->minutes(45)->seconds(59);
+        $this->assertEquals('2 days 46 minutes', $interval->forHumans(['parts' => 2, 'options' => CarbonInterface::ROUND]));
```
kylekatarnls: Glad to see this option is working with multipart!

kylekatarnls:
I think, we can have a more global approach. As it already exists for `Carbon`, `CarbonInterval` could have floor/ceil/round methods for each unit and special methods to floor/ceil/round the smallest not-zero unit.

Then options ROUND/FLOOR/CEIL could simply call those methods, so they can be used with `forHumans` but also for other purposes.

I take the assignee to make a proposal.

kylekatarnls:
I modified it so it takes cascade custom factors into account.

tylercollier:
Thanks for jumping on this. I'm impressed with what you've done so quickly. I'm glad you helped, I'm not sure how quickly I would've figured out the cascade factors. It's too bad that using those makes your succinct code suggestion you gave not work, but oh well.

You asked "What is `$index < count($diffIntervalArray) - 1?`" and you probably figured out it's required for dealing with seconds, at least the way I coded it where it grabs the next interval, and you don't want to go past the last one. I was going to add a test that shows it's needed (although, sadly, a code coverage test won't show that it's necessary because it's on the same line as the rest of the `if` statement), but your test `testRoundMultiplePartsGap` has seconds in it, so that suffices.

You said "Mostly I'm interesting in seeing what rounding happens when I have a month+week interval or a year+week interval." Can you write tests for those?

I thought I should point out some behavior just in case you didn't see it because it might not be wanted. Check out this test:

```php
public function testRoundOverflowNonSequentialRoundUp()
{
    CarbonInterval::setLocale('en');
    $interval = CarbonInterval::years(2)->months(35);
    $this->assertEquals('5 years', $interval->forHumans(['parts' => 1, 'options' => CarbonInterface::ROUND]));
}
```

This sort of makes sense, but contrasts with previous (non rounding) behavior dramatically, where if you had 2 years, and only 1 part, since it would always round down, then you'd end up with `2 years`, instead of `5 years`. Like I said, I'm not sure what's really wanted here. But with the old non-rounding behavior, if you have `CarbonInterval::years(2)->months(35)`, then `forHumans(['parts' => 2])` will report `2 years 35 months`. One might say that's so absurd that it's not worth dealing with.

kylekatarnls:
You can set 30/7 as you can customize your cascade factors but in many cases, it's easier to work with integer values, else months-to-days conversions would turn integer months count into floating days count. But I see your point. In rounding we would like 14 days to round to 0 months. But 30 is also an approximation because `DateInterval` does not have a reference date, so mostly when intervals are 15 days, we would also get surprising results:
```php
$round = ['options' => CarbonInterface::ROUND];
Carbon::parse('2019-01-10')->diffForHumans('2019-01-25', $round); // 15 days should round to 0 month in a 31-days month
Carbon::parse('2019-02-10')->diffForHumans('2019-02-25', $round); // 15 days should round to 1 month in a 28 days month
```

I'm also OK with the `5 years` result, but in this case we should have:
```php
$interval = CarbonInterval::days(2)->minutes(780);
$this->assertEquals('3 days', $interval->forHumans(['parts' => 1, 'options' => CarbonInterface::ROUND]));
```

kylekatarnls:
Hi @tylercollier here is a proposal using existing rounding methods and allowing options FLOOR and CEIL too. It also make `CarbonInterval::days(2)->minutes(780)` round to 3 days (with parts = 1) and `CarbonInterval::days(2)->minutes(59)->seconds(58)` round to 2 days 1 hour (with parts = 2).

Please challenge this implementation adding tests with cases I may have forgotten and feel free to revert changes if needed.

Thanks,

tylercollier:
I couldn't come up with anything additional, nicely done! One question is why do you have floor as a rounding type when the default is floor?

Looks like you added some public methods, are those all going to be part of the docs?

kylekatarnls:
Yes, toArray, getNonZeroValues and getValuesSequence sounds to me like they could be useful in the public API.

About FLOOR, it's not the default behavior. I added a unit test to show it:

```php
CarbonInterval::days(2)->minutes(59)->seconds(62)
  ->forHumans(['parts' => 2, 'options' => CarbonInterface::FLOOR]));
// 2 days 1 hour

CarbonInterval::days(2)->minutes(59)->seconds(62)
  ->forHumans(['parts' => 2]));
// 2 days 59 minutes
```

This can only happen with custom intervals. When using diffForHumans(), ago(), etc. you can't have units overflowing the maximum. So indeed when using `diffForHumans` or methods directly on `Carbon` instance, the `FLOOR` option will never make any difference, but if you manipulate intervals (add/sub/multiply/etc.) it can help.

PR #1935, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/


kylekatarnls:
Hi, thanks for your proposal. I'm not sure the readability decrease worth the 2.26% (on calls only) it gets.This way to write function calls seems not very widely used. I have to check if it's the right move.

I would prefer to keep the source clean and recommend users to use more global solutions to optimize this performances issue (https://github.com/Roave/FunctionFQNReplacer or https://github.com/dg/composer-backslasher)

Then,
1. Do you know if we still have a gap with pre-compiled files (OP cache, APC cache).
2. The article tests PHP 7.1 which is no longer supported now. What is the impact on modern versions from 7.2 to 7.4?
3. What are the results when adding `use explode;` at the top of the files?

I will do my own benchmarks, 1 benchmark result on a blog is not sufficient to make a decision and change 1518 lines of codes.

draco2003:
Hey kylekatarnls no problem at all. 
Another option (that wasn't in the tool that i was using) is to use the syntax like this:
use function is_array;

Which would accomplish the same performance gains without the extra \s everywhere. 
Let me know if that would be a preferred syntax. 

Also going to close this PR and create a new one. The settings on the fixer added the \ to way more functions than just the list of functions that it should have.

kylekatarnls:
I will see check the state of art (how other projects handle it) and get in touch soon.

Issue #1946, opened 2019-11-30
Carbon Get startOfMonth has BUG [BUG]


olympikesoft:
When using createFromFormat with month integer cannot get the first date of month.
To reproduce:

```php
if ($request->query('month') != null) {
    $dt = Carbon::createFromFormat('m', intval($request->query('month')));
    $start_week = $dt->startOfMonth();
    $end_week =  $dt->endOfMonth();
}
```
Making dd to start_week and end_week gets error.
`dd($start_week, $end_week);`

Result:
```
Carbon @1577836799 {#1004
  date: 2019-12-31 23:59:59.999999 UTC (+00:00)
}
Carbon @1577836799 {#1004
  date: 2019-12-31 23:59:59.999999 UTC (+00:00)
}
```

This is a BUG  right?

@smazurov @dereuromark @lanort @jrobeson  @dkmonaghan  @kylekatarnls

kylekatarnls:
Not at all, this is explained in the first chapter of the documentation:
https://carbon.nesbot.com/docs/

You need either to use `CarbonImmutable`:
```php
if ($request->query('month') != null) {
    $dt = CarbonImmutable::createFromFormat('m', intval($request->query('month')));
    $start_week = $dt->startOfMonth();
    $end_week =  $dt->endOfMonth();
}
```

Or to copy your instances:
```php
if ($request->query('month') != null) {
    $dt = Carbon::createFromFormat('m', intval($request->query('month')));
    $start_week = $dt->copy()->startOfMonth();
    $end_week =  $dt->copy()->endOfMonth();
}
```

PR #1947, opened 2019-11-30
added copy() function for boundaries.

abdurrahmanriyad:
Referring to this issue https://github.com/briannesbitt/Carbon/issues/1946
For boundary functions, all of them are replacing the properties of an existing object and returning the same object which causes an error like mentioned in the issue.


kylekatarnls:
That's the expected behavior for every modifier including startOf*/endOf*, use `CarbonImmutable` instead. See the first chapter of the documentation: https://carbon.nesbot.com/docs/

PR #1971, opened 2019-12-27
Speed up resetMessages

GrahamCampbell:
Before:

image

After:

image

### DIFF
--- a/src/Carbon/Translator.php
+++ b/src/Carbon/Translator.php
@@ -190,7 +190,8 @@ class Translator extends Translation\Translator
         }
 
         foreach ($this->getDirectories() as $directory) {
-            $data = @require sprintf('%s/%s.php', rtrim($directory, '\\/'), $locale);
+            $data = @include sprintf('%s/%s.php', rtrim($directory, '\\/'), $locale);
+
             if ($data !== false) {
                 $this->messages[$locale] = $data;
                 $this->addResource('array', $this->messages[$locale], $locale);
--- a/src/Carbon/Translator.php
+++ b/src/Carbon/Translator.php
@@ -190,9 +190,9 @@ class Translator extends Translation\Translator
         }
 
         foreach ($this->getDirectories() as $directory) {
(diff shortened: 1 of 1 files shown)


kylekatarnls:
Thanks, good idea. I will check why Travis fails.

kylekatarnls:
Hi @GrahamCampbell it fails because using `require` on non-existing file exit the processus. Would the performance remain the same using `include`?

GrahamCampbell:
Yeh, should be fine.

kylekatarnls:
This improvement + preloading feature (of OPCache for PHP 7.4) + requiring favorite languages files in the preloading script should improve significantly the Carbon loading in Laravel boot.

GrahamCampbell:
Though, I don't think preloading is capable of pre-processing conditional includes.

kylekatarnls:
I'm basically suggesting when developing an app (a Laravel app for instance), you can add a preloading script:
```php
// Include all from composer classmap optimized autoload

require __DIR__.'/vendor/nesbot/carbon/src/Carbon/Lang/en.php';
```
Assuming the app only use English for instance, so **en.php** is preloaded and the boot of the app should load even faster as en.php would already be in memory, so if I well understand the preloading feature `$data = @include sprintf('%s/%s.php', rtrim($directory, '\\/'), $locale);` should not actually access the file if `$locale` is `"en"`.

GrahamCampbell:
Well, the opcache system will already do this?

kylekatarnls:
Not an expert of this feature, but what I understood is: OPCache will cache the file the first time it's loaded (so slower the first time), then check if the file changed and use cache if not changed for other loadings (faster).

With the preloading script, you can also fasten the first loading.

Bonus: you can clear cache on deploy to disable also the change checking. AFAIK this can be done with or without the preloading script.

GrahamCampbell:
Yeh, that's right.

kylekatarnls:
Thank you very much for your help, it will be released in 2.29.0.

Issue #1967, opened 2019-12-19
Week methods only work with local translator, not default locale


hackel:
Hello,

I encountered an issue with the following code:
```php
echo Carbon::getLocale(); // en
echo Carbon::now()->locale(); // en
echo Carbon::now()->startOfWeek()->dayName; // Monday
Carbon::setLocale('en_US');
echo Carbon::now()->startOfWeek()->dayName; // Monday

echo Carbon::now()->locale('en')->startOfWeek()->dayName; // Sunday
```
Trying to figure out all of the different (apparently competing?) localization settings makes my head spin.  The issue seems to be that "firstWeekDay" will only ever be determined by a local translator, not the default translator:

Carbon\Traits\Localization\Date:928
```php
case $name === 'firstWeekDay':
    return $this->localTranslator ? ($this->getTranslationMessage('first_day_of_week') ?? 0) : static::getWeekStartsAt();
```

I'm not sure what the point of `static::$weekStartsAt` is...backward compatibility?  It conflicts with what the default translator is supposed to return.

As far as I can tell, the only way to solve this is to use the deprecated methods `setWeekStartsAt()` AND `setWeekEndsAt()` (which for some bizarre reason is not calculated automatically), or explicitly include the `locale()` call every single time I use these methods (even though it already is the default).  Am I missing something obvious here?

I would propose removing the default values for `static::$weekStartsAt/EndsAt`, and changing the getters to only use them if they are explicitly set:
```php
case $name === 'firstWeekDay':
    return static::getWeekStartsAt() ?? $this->getTranslationMessage('first_day_of_week', null, 0);
```

This is a simple change and I would be happy to submit a PR if acceptable, but it would technically be breaking, even though it's a bugfix in my opinion.

Carbon version: 2.28.0

PHP version: 7.4.0

I expected to get: (with default 'en' or 'en_US' locale)

```
echo Carbon::now()->startOfWeek()->dayName; // Sunday
```

But I actually get:

```
echo Carbon::now()->startOfWeek()->dayName; // Monday
```

Thanks!

kylekatarnls:
Hello,

This is a legacy. `startOfWeek` exists for a long time before `first_day_of_week` was supported in different locales.

At this time Monday as the standard start of week was the default value for `startOfWeek` and when the I18N was added, default locale `en` has been chosen as `diffForHumans` returned English text.

So changing this behavior would not be a bug fix but really a breaking changes.

[...]

PR #1980, opened 2020-01-08
CarbonPeriod and CarbonImmutable not behaving similarly

deleugpn:
This Pull Request is in the form of Help Wanted / Possible Bug report.

I recently noticed that when using `CarbonPeriod` and Carbon, they differ in configuration of when the week start (Monday vs Sunday) causing inconsistency. I'm looking for help on validating whether there's something I can do to fix this issue or if it's actually a bug that would need to be addressed in Carbon itself.

The use case:

Parsing the date **2019-12-01** and asking for the first day of the Week, by default, returns **2019-11-25**, which is the Monday of said Week. Using CarbonPeriod to make an array of weeks will give an array of Carbon instances that are differently configured and will consider **Sunday** as the first day of the week. When asking for the first day of the week from **2019-11-25**, it further reduces 1 day to 24th of November (Sunday).

- Is there a way to instruct CarbonPeriod to be consistent with default Carbon settings?
- Is this something I can do while looping through each row of Carbon Period?
- Is this something worth investigating inside Carbon itself?

Thanks!

### DIFF
--- a/src/Carbon/CarbonPeriod.php
+++ b/src/Carbon/CarbonPeriod.php
@@ -1446,10 +1446,11 @@ class CarbonPeriod implements Iterator, Countable, JsonSerializable
         $this->key = 0;
         $this->current = call_user_func([$this->dateClass, 'make'], $this->startDate);
         $settings = $this->getSettings();
-        $locale = $this->getTranslatorLocale();
-        if ($locale) {
-            $settings['locale'] = $locale;
+
+        if ($this->hasLocalTranslator()) {
+            $settings['locale'] = $this->getTranslatorLocale();
         }
+
         $this->current->settings($settings);
         $this->timezone = static::intervalHasTime($this->dateInterval) ? $this->current->getTimezone() : null;
 
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -99,7 +99,13 @@ trait Creator
         $instance = new static($date->format('Y-m-d H:i:s.u'), $date->getTimezone());
 
         if ($date instanceof CarbonInterface || $date instanceof Options) {
-            $instance->settings($date->getSettings());
+            $settings = $date->getSettings();
+
+            if (!$date->hasLocalTranslator()) {
+                unset($settings['locale']);
+            }
+
+            $instance->settings($settings);
         }
 
         return $instance;
--- a/src/Carbon/Traits/Localization.php
+++ b/src/Carbon/Traits/Localization.php
@@ -138,6 +138,16 @@ trait Localization
     }
 
     /**
+     * Return true if the current instance has its own translator.
+     *
+     * @return bool
+     */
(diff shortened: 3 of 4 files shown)


kylekatarnls:
First of all here is how to specify explicitly the start of week and get 25th november for both:
```php
$sunday = CarbonImmutable::parse('2019-12-01');
$period = CarbonPeriod::create($sunday->startOfWeek(CarbonImmutable::MONDAY), '1 week', $sunday->endOfWeek(CarbonImmutable::MONDAY))->toArray();
$formattedSunday = $sunday->startOfWeek(CarbonImmutable::MONDAY)->format('Y-m-d H:i:s');

echo $formattedSunday."\n";
echo $period[0]->toImmutable()->startOfWeek(CarbonImmutable::MONDAY)->format('Y-m-d H:i:s');
```

And the root cause of this behavior is the locale. That's why setting the period on a locale having Monday as start of week will also work as you expect:
```php
$sunday = CarbonImmutable::parse('2019-12-01');
$period = CarbonPeriod::create($sunday->startOfWeek(), '1 week', $sunday->endOfWeek())->locale('en_GB')->toArray();
$formattedSunday = $sunday->startOfWeek()->format('Y-m-d H:i:s');

echo $formattedSunday."\n";
echo $period[0]->toImmutable()->startOfWeek()->format('Y-m-d H:i:s');
```

In a next major version, the default start of week might become Sunday (to be aligned on the default locale en = en_US) (see #1967)

Right now, the date use Monday as start of week if it has no explicit locale.

The bug in the particular case of the CarbonPeriod is the period propagate its settings to the dates it creates including its locale:

```php
$sunday = CarbonImmutable::parse('2019-12-01');
$period = CarbonPeriod::create($sunday->startOfWeek(), '1 week', $sunday->endOfWeek(), CarbonPeriod::IMMUTABLE)->toArray();
$formattedSunday = $sunday->startOfWeek()->format('Y-m-d H:i:s');

var_dump($sunday); // no localTranslator
var_dump($period[0]); // has localTranslator
```

So the fix would be to handle the case where localTranslator is `null` so it can be propagated to iterated dates.

Side note: use `CarbonPeriod::IMMUTABLE` option to not have to call `->toImmutable()` on each date.

deleugpn:
I forgot to thank you for your reply. I decided to follow your suggestion on explicitly setting when the Week starts `$sunday->startOfWeek(CarbonImmutable::MONDAY)` as it will even make my code resilient to a major change when Carbon decides to start on Sunday by default.
It completely resolved my issue and the code is nice.
I also adopted the suggestion for `CarbonPeriod` with the `::Immutable` parameter. Would be nice to have a named constructor for that, though 😄

kylekatarnls:
I get your test to pass without breaking other tests (without having to change the default start-of-week day).

We'll have the correct locale setting propagation through `CarbonPeriod` in the next version 2.29

Issue #1992, opened 2020-01-17
Setting ->timezone stop working after iterating on date properties or after unserialization


sorinsarca:
Hello,

According to the [docs](https://carbon.nesbot.com/docs/#api-setters), the `timezone` property setter should work: 
 _The following setters are implemented via PHP's `__set()` method. Its good to take note here that none of the setters, **with the obvious exception of explicitly setting the timezone, will change the timezone of the instance**._

But this stops working when the Carbon instance is unserialized or after the object was exported, printed, etc.

Working example

```php
$date = new \Carbon\Carbon("2020-01-01", "America/Vancouver");

echo $date->getTimezone()->getName(), PHP_EOL; // America/Vancouver

$date->timezone = "UTC";

echo $date->getTimezone()->getName(), PHP_EOL; // UTC
```

`unserialize` bug

```php
$date = new \Carbon\Carbon("2020-01-01", "America/Vancouver");

$date = \unserialize(\serialize($date));

echo $date->getTimezone()->getName(), PHP_EOL; // America/Vancouver

$date->timezone = "UTC";

// It should display UTC but it shows America/Vancouver
echo $date->getTimezone()->getName(), PHP_EOL; 
```

`var_export`, `print_r` or similar bug

```php
$date = new \Carbon\Carbon("2020-01-01", "America/Vancouver");

echo $date->getTimezone()->getName(), PHP_EOL; // America/Vancouver

$contents = var_export($date, true);

$date->timezone = "UTC";

// It should display UTC but it shows America/Vancouver
echo $date->getTimezone()->getName(), PHP_EOL;
```

`foreach` bug

```php
$date = new \Carbon\Carbon("2020-01-01", "America/Vancouver");

echo $date->getTimezone()->getName(), PHP_EOL; // America/Vancouver

foreach ($date as $value) {}

$date->timezone = "UTC";

// It should display UTC but it shows America/Vancouver
echo $date->getTimezone()->getName(), PHP_EOL;
```

`get_object_vars` bug

```php
$date = new \Carbon\Carbon("2020-01-01", "America/Vancouver");

echo $date->getTimezone()->getName(), PHP_EOL; // America/Vancouver

$vars = get_object_vars($date);

$date->timezone = "UTC";

// It should display UTC but it shows America/Vancouver
echo $date->getTimezone()->getName(), PHP_EOL;
```

And probably a lot of other examples.

------------

Carbon version: 2.28

PHP version: 7.3.*

[...]

PR #1993, opened 2020-01-17
Fix #1988 Allow timezone_type, timezone, date to be modified before __construct

### LINKED ISSUE #1988: InvalidArgumentException : Unknown setter 'date’
Hello,

I have an issue with Carbon dates that are serialised through [GitHub - opis/closure: Serialize closures (anonymous functions)](https://github.com/opis/closure)

Laravel version: 6.11.0
Carbon version: 2.28.0
PHP version:  7.3.10
Opis/Closure version: 3.5.1

The problem seems to occur after upgrading to Laravel 6 and therefor upgraded from Carbon 1 to Carbon 2.

I wasn’t sure if this belongs to the Carbon or Opis issues. So I’ve posted this question on both. I will provide links later.

When a closure is serialized:
```php
return serialize(new SerializableClosure($closure, true));
```

I get this error: `InvalidArgumentException : Unknown setter 'date’`.

This boils down to `/nesbot/carbon/src/Carbon/Traits/Date.php:1166`.

I’ve googled this issue for some time now and I believe it has got a lot to do with the depreciation of `serializeUsing` on Carbon 2.
As found in issue [Provide an alternative to serializeUsing to Laravel users · Issue #1870 · briannesbitt/Carbon · GitHub](https://github.com/briannesbitt/Carbon/issues/1870). @driesvints

However after reading multiple threats I’m not able to find a solution. I really hope someone can help me steering me to the right direction 🙏.

Thanks!

kylekatarnls:
This allow DateTime-compatible using `(new ReflectionClass(Carbon::class))->newInstanceWithoutConstructor()`

### DIFF
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -1037,6 +1037,12 @@ trait Date
      */
     public function __set($name, $value)
     {
+        if (!$this->constructed && in_array($name, ['timezone_type', 'timezone', 'date'])) {
+            $this->$name = $value;
+
+            return;
+        }
+
         $this->set($name, $value);
     }
 
@@ -1063,12 +1069,6 @@ trait Date
             foreach ($name as $key => $value) {
                 $this->set($key, $value);
             }
-
-            return $this;
-        }
-
-        if (!$this->constructed && in_array($name, ['timezone_type', 'timezone', 'date'])) {
-            $this->$name = $value;
 
             return $this;
         }
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -1067,6 +1067,12 @@ trait Date
             return $this;
         }
 
+        if (!$this->constructed && in_array($name, ['timezone_type', 'timezone', 'date'])) {
+            $this->$name = $value;
+
+            return $this;
+        }
+
         switch ($name) {
             case 'milliseconds':
             case 'millisecond':
@@ -1152,6 +1158,11 @@ trait Date
             case 'timezone':
             case 'tz':
                 $this->setTimezone($value);
+
+                break;
+
+            case 'date':
+                $this->modify($value);
 
                 break;
 
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -1040,13 +1040,13 @@ trait Date
      */
     public function __set($name, $value)
     {
-        if (!$this->constructed && in_array($name, ['timezone_type', 'timezone', 'date'])) {
-            $this->$name = $value;
+        if ($this->constructedObjectId === spl_object_hash($this)) {
+            $this->set($name, $value);
 
             return;
         }
 
-        $this->set($name, $value);
+        $this->$name = $value;
     }
 
     /**
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -1161,11 +1161,6 @@ trait Date
 
                 break;
 
-            case 'date':
-                $this->modify($value);
-
-                break;
-
             default:
                 if (static::hasMacro($macro = 'set'.ucfirst($name))) {
                     $this->$macro($value);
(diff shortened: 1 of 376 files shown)


src/Carbon/Traits/Date.php
```diff
@@ -1155,6 +1161,11 @@ public function set($name, $value = null)
 
                 break;
 
+            case 'date':
```
sorinsarca: For safety reasons, I think this case should be postponed until #1992 is fixed, otherwise there will be the same bug for the `date` property.

kylekatarnls: Yep, indeed it's a first try that shouldn't be there.

stefro:
After commit 0a5ed5a I get `InvalidArgumentException : Unknown setter 'date'` again.

Shouldn't it be (`!$this` -> `$this`):
```php
        if ($this->constructed && in_array($name, ['timezone_type', 'timezone', 'date'])) {
            $this->$name = $value;
            return $this;
        }
```

If I change that, it works fine.

kylekatarnls:
Of course, thanks. Can't explain what the unit test pass. Let's recheck all that.

kylekatarnls:
Oups sorry, I read too fast, nope, it should not be `$this->constructed`, `!$this->constructed` is correct, but as unserialized instances do not trigger __construct, the solution may be to also set the `constructed` property in `__wakeup` as @msarca did in #1994.

stefro:
I've pulled the 8209f88 into my project and unfortunately it gives me the `InvalidArgumentException : Unknown setter 'date'` again.

kylekatarnls:
However this code added is the same process as in opis/closure: https://github.com/briannesbitt/Carbon/pull/1993/commits/638291e402940a002713d6fb5a3a5fe0c03c6ca3#diff-62330e87d38a327cd927526b11a0da6cR104
And this unit test is passing.

@msarca Do you see something that could differ in this unit test from opis/closure unserialization?

> It's the `$this->constructed` part that confuses me. I'll try to go over the code to understand.

It's about enabling the __set feature only once the datetime is initialized (__construct or __wakeup).

So before the construction (`!$this->constructed`), we set the property as a public one (`$this->$name = $value`)

stefro:
Maybe it my poor understanding of this complex piece of code. 
>but as unserialized instances do not trigger __construct, the solution may be to also set the constructed property in __wakeup as @msarca did in #1994.
I get the `$this->constructed = true;` in `Serialization.php` since you've put it in the `__wakeup()` method.

But the ` $this->constructed = true;` in `Creator.php (line 79) confuses me. When I comment it out, no exceptions are thrown. When I leave it in I still get `InvalidArgumentException : Unknown setter 'date'`.

kylekatarnls:
Yes this error is gone and the __set is broken, it's like you would simply remove the whole `__set` and so breaking many feature.

If you still have the error, it would mean opis/closure set the `date` property after the date is yet initialized.

kylekatarnls:
I just checked to be sure, and using `$this->constructed` instead of `!$this->constructed` breaks 37 other unit tests :( So nope, not a solution.

kylekatarnls:
Let's make some dreams of it, solution will raise up tomorrow.

msarca:
> @msarca Do you see something that could differ in this unit test from opis/closure unserialization?

Yep. We also iterate over the properties of the ancestors using a while loop `$while($reflection = $reflection->getParentClass())`

stefro:
>Let's make some dreams of it, solution will raise up tomorrow.

Great idea! I'll be back in the morning. Thanks for your effort today, all of you. Are any of you guys at Laracon EU this year? I feel like I need to buy you some beers :-)

msarca:
I've updated my PR and came up with a solution that works. Take a look guys and give me some feedback tomorrow.

stefro:
I can conform that 5ddb5b85 in the PR from @msarca works! Awesome!!!!! 🎉

kylekatarnls:
Awesome, we're almost done, we just have 1 unit test that fail on PHP 7.4:
https://travis-ci.org/briannesbitt/Carbon/jobs/638654028#L463

I will check

stefro:
I’m not at home right now. I will check as soon as I’m back to confirm.

kylekatarnls:
Just one more thing, the ID also need to be updated when cloned.

PR #1994, opened 2020-01-17
Fixed serialization issue

msarca:
A `Carbon` object can be obtained in 3 ways:
- using the `new` keyword 
- using the `newInstanceWithoutConstructor` method
- using the `unserialize` function

I've added a new property called `isDateTimeInitialised` that will tell us if a Carbon object has been initialised or not. The only time when the object is not initialised is when the `newInstanceWithoutConstructor` is used to obtain the object. 

When setting a property, checking if the object was initialised or not is done in the `__set` method. This is simply because we need to be able to serialize imutable objects too.

### DIFF
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -32,6 +32,8 @@ use InvalidArgumentException;
  */
 trait Creator
 {
+    use ObjectInitialisation;
+
     /**
      * The errors that can occur.
      *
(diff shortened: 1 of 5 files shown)


msarca:
The issue here was caused by the fact that when we were iterating over the source object properties, the `isDateTimeInitialised` was set to `true`. This property was also set to the target object :) making it look like an object built via `new` or `unserialize`. 

So in order to fix this  I've renamed `isDateTimeInitialised` to `constructedObjectId` and when the `__construct` or `__wakeup` methods are called this property is set to the value returned by `spl_object_hash($this)`.

stefro:
I can conform this last commit works like a charm! Great job! 🎉

Issue #2002, opened 2020-01-27
CarbonInterval does not honor setting "toStringFormat".


digitaladapt:
Hello,

I encountered an issue with the following code:
```php
echo CarbonInterval::createFromFormat('H:i:s', '01:02:03')->settings(['toStringFormat' => '%H:%I:%S']);
```

Carbon version: **2.29.1**

PHP version: **7.2.24**

I expected to get:

```
"01:02:03"
```

But I actually get:

```
"1 hour 2 minutes 3 seconds"
```

CarbonInterval never checks settings for formatting override, and instead unconditionally uses the function $this->forHumans(), rather than respecting the value set in $this->localToStringFormat.

Purposed patch attached (use existing Converter Trait):

[CarbonInterval-ToStringFormat.txt](https://github.com/briannesbitt/Carbon/files/4119371/CarbonInterval-ToStringFormat.txt)

Thanks!

kylekatarnls:
I'm OK for this new feature. Would you like to submit a pull-request so your authorship would be saved in [contributors page](https://github.com/briannesbitt/Carbon/graphs/contributors)?

kylekatarnls:
Will be fixed in 2.30.

PR #2004, opened 2020-01-28
#2002 Fix CarbonInterval not honoring "toStringFormat" setting, with tests.

digitaladapt:


digitaladapt:
Apparently using the Converter Trait was a bad idea.
Going to fix the code and try again.

PR #2016, opened 2020-02-17
createFromFormat may also return false (Carbon\CarbonImmutable)

HenkPoley:
Fixes vimeo/psalm reporting RedundantConditionGivenDocblockType on `if ($maybeCarbon instanceof Carbon)`  in code that uses Carbon.


kylekatarnls:
See https://github.com/briannesbitt/Carbon/pull/2017#issuecomment-586882423

kylekatarnls:
For now I prefer to close and keep the PHPDoc as is. I would re-open if there are other issues to be considered.

PR #2017, opened 2020-02-17
createFromFormat may also return false (Carbon\Carbon)

HenkPoley:
Fixes vimeo/psalm reporting  RedundantConditionGivenDocblockType on `if ($maybeCarbon instanceof Carbon)` in code that uses Carbon.


kylekatarnls:
For the record, if you're using the default mode (strict mode), then, it's really redundant, as it throws an exception:
```php
        if (static::isStrictModeEnabled()) {
            throw new InvalidArgumentException(implode(PHP_EOL, $lastErrors['errors']));
        }

        return false;
```

So `createFromFormat()` can only return `false` if the deprecated `Carbon::useStrictMode(false);` has been called.

I'm not sure we should have this edge-case in the PHPDoc.

And if we do, it should not break the PHPDoc columns alignment, so it should be done in:
https://github.com/briannesbitt/Carbon/blob/master/phpdoc.php#L30

And use the automated generation:
```shell
composer phpdoc
```

HenkPoley:
Ah, I see.. I don't use that mode.

Sadly PhpDoc cannot document exception classes on interfaces 😒

kylekatarnls:
To be sure, you can check `Carbon::isStrictModeEnabled()` and `CarbonImmutable::isStrictModeEnabled()`, if they're `true` you should never get `createFromFormat()` returning `false` and so don't need `if ($maybeCarbon instanceof Carbon)`

If you have a doubt, show me the code chunk where it's used.

kylekatarnls:
For now I prefer to close and keep the PHPDoc as is. I would re-open if there are other issues to be considered.

PR #2019, opened 2020-02-21
always use non-deprecated translation method

xabbuh:


kylekatarnls:
I approve this change, but I need to check the unit tests failure.

PR #2021, opened 2020-02-24
Skip test when an English locale is unavailable

rgson:
testSetLocale fails in environments lacking the desired English locales.

Other tests that rely on specific locales being available on the system are
skipped if those locales are unavailable.

This change will similarly skip this test if the desired locale is not found.


kylekatarnls:
This is correct regarding other tests where we do equivalent checking.

Could you align `testComaDecimalSeparatorLocale` on both `Carbon` and `CarbonImmutable` tests that should check "fr" locale is available too?

This way all the tests using `setlocale` would have the same checking.

rgson:
@kylekatarnls Done.

kylekatarnls:
Perfect, thanks a lot.

PR #2045, opened 2020-03-25
Fix doctrine orm carbon type

Flodess:
Hello, 

When I tried the new carbon type, I got a "Data missing".

This is due to postgres not using the 'Y-m-d H:i:s.u' format but 'Y-m-d H:i:s' when microseconds are set to zero.

The solution is based on this comment :
https://github.com/doctrine/dbal/issues/2873#issuecomment-535452749

### DIFF
--- a/src/Carbon/Doctrine/CarbonType.php
+++ b/src/Carbon/Doctrine/CarbonType.php
@@ -52,8 +52,7 @@ trait CarbonType
             return $class::instance($value);
         }
 
-        $date = $class::createFromFormat('Y-m-d H:i:s.u', $value)
-            ?: $class::instance(date_create($value));
+        $date = $class::parse($value);
 
         if (!$date) {
             throw ConversionException::conversionFailedFormat(


Flodess:
The solution is based on this comment :
https://github.com/doctrine/dbal/issues/2873#issuecomment-535452749

kylekatarnls:
I think here we could simply use parse instead. Does the following work fine for you with Postgres?

```php
$date = $class::parse('Y-m-d H:i:s.u', $value) ?: $class::instance(date_create($value));
```

This should also ease to change the precision keeping existing data.

For the record, I highly recommend to store microseconds precision. This is easy to trim data you don't need but impossible to get the data you didn't store.

src/Carbon/Doctrine/CarbonType.php
```diff
@@ -52,8 +53,9 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
             return $class::instance($value);
         }
 
-        $date = $class::createFromFormat('Y-m-d H:i:s.u', $value)
-            ?: $class::instance(date_create($value));
+        $date = $class::instance(
+            DateTime::createFromFormat('Y-m-d H:i:s.u', $value) ?: date_create($value)
+        );
```
kylekatarnls: Can you try if we can optimize this with:
```php
$date = $class::parse($value);
```

kylekatarnls:
Thanks, I will publish a patch for that today.

FlodessRdcRs:
Waouh, that was fast ! Thx :)

Flodess:
Waouh, that was fast! Thx :)

kylekatarnls:
2.32.1 released.

PR #2046, opened 2020-03-26
Update uz.php

Adizbek:
As we used carbon in our project, some translations look wierd so I decided to translate them properly, please merge them as soon as possible.


src/Carbon/Lang/uz.php
```diff
@@ -48,8 +49,8 @@
     's' => ':count с',
     'ago' => ':time олдин',
     'from_now' => 'Якин :time ичида',
-    'after' => ':time пас аз он',
-    'before' => ':time пеш аз он',
+    'after' => ':time дан кейин',
+    'before' => ':time олдин',
```
kylekatarnls: This translation is the same as "ago", is there any way to distinguish both?

before is relative: we will leave the office 2 hours before the night (both event are in the future but the first is before the second)
ago is absolute: we leaved the office 2 hours ago (= 2 hours before now).

kylekatarnls:
Thanks for your contribution, can you check the comment above. Ideally ago and before should not be the same if the language has a way to distinguish both.

PR #2050, opened 2020-03-30
Update doctrine ORM Carbon type

Flodess:
Hello,

I don't think the comment is required for both Datetime and DatetimeImmutable.
"true" on DatetimeImmutable only should be enough, so, no need to override requiresSQLCommentHint.


kylekatarnls:
I don't get why this setting should be different in mutable or immutable. I guess the both types should require it or not.

kylekatarnls:
Oh, I think I know why you propose this change now.

For a fresh project, you can do:
```php
doctrine:
    dbal:
        types:
            datetime_immutable: \Carbon\Doctrine\DateTimeImmutableType
            datetime: \Carbon\Doctrine\DateTimeType
```

And so rely datetime => Carbon based on type only.

But if you need to migrate incrementally your app, you may have:
```php
doctrine:
    dbal:
        types:
            carbon_immutable: \Carbon\Doctrine\DateTimeImmutableType
            carbon: \Carbon\Doctrine\DateTimeType
```

And then you need the comment to distinguish Carbon from DateTime typing.

This is how I understand the comment hint. But I may have misunderstood it. If you think there's a mistake, please show me an example of the issue it may causes.

Flodess:
Mmmm, sorry I should have give more context on this.

So, yeah I am in the first context. And having the comment generated in the migration for datetime made no sense to me in this case.

I quickly tested your second case and it doesn't generate automatically the rights comments.
`$this->addSql('COMMENT ON COLUMN test_carbon.carbon IS \'(DC2Type:datetime_immutable)\'');`
So, only one out of two was added, and it's the wrong type. :/
For this case, if I understand correctly, getName() need to be implemented and need to match the dbal type in the configuration. But if we do that, I don't think the datatype can still be used to override datetime and datetime_immutable. :/

I think the first case should be prioritized. On the long term, I don't see any reason to keep both on a project. If someone want to migrate incrementally its app, he is free to implement its own datatype to override getName().

I hope I'm clear enough in my explanation ><

kylekatarnls:
OK, it seems not to be a blocking point to let this `return true` for now. So I will take the time to find the better way to make the reverse detection work with both setting kinds.

kylekatarnls:
Hi @Flodess this has been merged and new types are added so you can either:
```yaml
doctrine:
    dbal:
        types:
            datetime_immutable: \Carbon\Doctrine\DateTimeImmutableType
            datetime: \Carbon\Doctrine\DateTimeType
```
or:

```yaml
doctrine:
    dbal:
        types:
            carbon_immutable: \Carbon\Doctrine\CarbonImmutableType
            carbon: \Carbon\Doctrine\CarbonType
```
(Note than "Carbon" classes must be changed too accordingly).

You can test it requiring `"nesbot/carbon": "2.x-dev"`

Please tell me if you have any issue with the new version.

Flodess:
Hi @kylekatarnls,

Looks good ! :+1: The migration diff file was generated as I expected.
Thx for the work you put into it ! :)

PR #2051, opened 2020-03-30
Fix ORM configuration paths

Flodess:
Hello,

Just a quick fix on the paths in the doc. :)


kylekatarnls:
Thanks, that was the first namespace I gave before realizing this actually not depends on Symfony.

PR #2055, opened 2020-04-05
Typo in example of phpdoc

canarslan12:
Nothing important but correction is correction :)

### DIFF
--- a/src/Carbon/Traits/Comparison.php
+++ b/src/Carbon/Traits/Comparison.php
@@ -354,9 +354,9 @@ trait Comparison
      *
      * @example
      * ```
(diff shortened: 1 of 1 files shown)


canarslan12:
Wait a minute, isn't it needs to be true?
https://github.com/briannesbitt/Carbon/blob/af3420c262376a5bfa6bd0eb4ac0945ea0fff773/src/Carbon/Traits/Comparison.php#L399

kylekatarnls:
Yes it does :'(

kylekatarnls:
Thanks a lot, that's exactly why between and the third argument is too ambiguous and why explicit methods betweenIncluded and betweenExcluded should be used instead :D

PR #2056, opened 2020-04-06
Add ext-mbstring to requirements

cino:
Since Carbon is using mb_ functions this should be in the composer.json to make sure the mbstring extension is installed. 

(And also, for my personal sanity when browsing the code in PHPStorm)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
     "require": {
         "php": "^7.1.8 || ^8.0",
         "ext-json": "*",
+        "ext-mbstring": "*",
         "symfony/translation": "^3.4 || ^4.0 || ^5.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


kylekatarnls:
I wonder if it causes issues to users using symfony/polyfill (or similar), thanks to the polyfill, they can run Carbon with no trouble with PHP built with mb extension. But adding this requirement will prevent them from installing next version Carbon unless they would use `--ignore-plateform-reqs` (which also ignore PHP level and other missing extensions, so not ideal).

kylekatarnls:
Maybe we could consider requiring `"symfony/polyfill-mbstring": "^1.0",` instead, so it makes the extension optional, use it if present, else use the polyfill. It would not break any existing application as it would allow any currently existing version of the polyfill-mbstring (and it contains all the functions we need).

Then for many people, it will not change anything (polyfill-mbstring is already installed via Symfony, Laravel and many other packages).

Why do you think about?

cino:
I agree. I didn't think about the polyfill when creating the PR. I totally agree that requiring the polyfill is a better (non breaking) solution. 👍

PR #2062, opened 2020-04-08
Fix #2059 timezone offset name for not full hour

### LINKED ISSUE #2059: Invalid offset name for timezone without full hour offset
Hello,

```php
\Carbon\CarbonTimeZone::create('Pacific/Marquesas')->toOffsetName()
```

Carbon version: 2.31.0

PHP version: 7.2.25

I expected to get:

```
-09:30
```

But I actually get:

```
-10:30
```

Timezone Pacific/Marquesas has offset -09:30, no summer time (please to check https://www.zeitverschiebung.net/en/timezone/pacific--marquesas).

The bug is in line [code](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/CarbonTimeZone.php#L157)
```
$hours = floor($minutes / 60);
```
It floors 9.5 to 10. It should get integer value of the number. Minutes are calculated below. This bug is probably for all timezones without full hour (like : +13:45)

Thanks!

omatrycy:
Fix #2059 problem with timezone offset
Use data provider for testing multiple cases

### DIFF
--- a/src/Carbon/CarbonTimeZone.php
+++ b/src/Carbon/CarbonTimeZone.php
@@ -154,7 +154,7 @@ class CarbonTimeZone extends DateTimeZone
     {
         $minutes = floor($this->getOffset($date ?: Carbon::now($this)) / 60);
 
-        $hours = floor($minutes / 60);
+        $hours = (int) ($minutes / 60);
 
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Thanks 👍 

This will be available in 2.33.

PR #2066, opened 2020-04-13
Accept ar_AE in localization test

rgson:
Avoids a test failure because of "Expected: 'ar' Actual: 'ar_AE'".
It corresponds to the approach to 'sr' and 'zh' in the same test.


kylekatarnls:
Fair. 👍

PR #2069, opened 2020-04-20
Fix static analyses of Carbon

deleugpn:
The current update of Larastan started emitting a new error (as reported here https://github.com/nunomaduro/larastan/issues/541). Making the method signature consistent with the Interface could fix the problem.


kylekatarnls:
Thanks!

deleugpn:
Thank you for your great work on Carbon!

PR #2086, opened 2020-05-12
Fix typo

kkevindev:


kylekatarnls:
Thanks, I will regenerate the documentation tomorrow with this fix too when I'll release the 2.35.0.

PR #2093, opened 2020-05-21
Update Sinhala translations

MalindaWMD:
Updated a couple of translations according to the usual usage in Sinhala.

Major changed are,

- `12:00 පෙරවරු, 12:00 පෙ.ව.` should be `පෙරවරු 12:00, පෙ.ව. 12:00`,

- In `තත්පර 1 කට පෙර`, space between `1` and `කට` should be removed. Also when using `කින්` and `න්`, leading space should be removed.

- `පෙර වරු`, `පස් වරු` should be `පෙරවරු`, `පස් වරු`. (Without space)

### DIFF
--- a/src/Carbon/Lang/si.php
+++ b/src/Carbon/Lang/si.php
@@ -20,7 +20,7 @@ return [
     'year' => '{1}වසර 1|වසර :count',
     'a_year' => '{1}වසරක්|වසර :count',
     'month' => '{1}මාස 1|මාස :count',
-    'a_month' => '{1}මාසය|මාස :count',
+    'a_month' => '{1}මාසයක්|මාස :count',
     'week' => '{1}සති 1|සති :count',
     'a_week' => '{1}සතියක්|සති :count',
     'day' => '{1}දින 1|දින :count',
@@ -30,22 +30,22 @@ return [
     'minute' => '{1}මිනිත්තු 1|මිනිත්තු :count',
     'a_minute' => '{1}මිනිත්තුවක්|මිනිත්තු :count',
     'second' => '{1}තත්පර 1|තත්පර :count',
-    'a_second' => '{1}තත්පර කිහිපයකට|තත්පර :count',
-    'ago' => ':time කට පෙර',
+    'a_second' => '{1}තත්පරයක්|තත්පර :count',
+    'ago' => ':timeකට පෙර',
     'from_now' => function ($time) {
         if (preg_match('/දින \d+/', $time)) {
             return $time.' න්';
         }
 
-        return $time.' කින්';
+        return $time.'කින්';
     },
-    'before' => ':time කට පෙර',
+    'before' => ':timeකට පෙර',
     'after' => function ($time) {
         if (preg_match('/දින \d+/', $time)) {
-            return $time.' න්';
+            return $time.'න්';
         }
 
-        return $time.' කින්';
+        return $time.'කින්';
     },
     'diff_now' => 'දැන්',
     'diff_yesterday' => 'ඊයේ',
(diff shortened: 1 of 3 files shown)


tests/Localization/SiLkTest.php
```diff
@@ -195,29 +195,29 @@ class SiLkTest extends LocalizationTestCase
...
         'වසර 2 මාස 3 දින 1 තත්පර 1',
         // Carbon::now()->addYears(3)->diffForHumans(null, null, false, 4)
-        'වසර 3 කින්',
+        'වසර 3කින්',
         // Carbon::now()->subMonths(5)->diffForHumans(null, null, true, 4)
-        'මාස 5 කට පෙර',
+        'මාස 5කට පෙර',
         // Carbon::now()->subYears(2)->subMonths(3)->subDay()->subSecond()->diffForHumans(null, null, true, 4)
-        'වසර 2 මාස 3 දින 1 තත්පර 1 කට පෙර',
+        'වසර 2 මාස 3 දින 1 තත්පර 1කට පෙර',
```
kylekatarnls, line 208: This looks weird. Can you explain the spacing rule in Sinhala?

kylekatarnls:
I don't know why Travis results did not show on this PR but here they are:
https://travis-ci.org/github/briannesbitt/Carbon/builds/689719927

To change 12:00 පෙරවරු, 12:00 පෙ.ව. into පෙරවරු 12:00, පෙ.ව. 12:00, you just have to use `LT` format instead of `h:mm a` for instance. And so tests with explicit format `h:mm A, h:mm a` should not be changed as with custom user order, the usual order for the current language (given by the `LT` format won't apply).

MalindaWMD:
It's actually like this,

In English it's *2 years 3 months 1day 1 second. *
*2 years *-> වසර 2
*3 months *-> මාස 3
*1 day -> *දින 1

We have to use a space after each word(වසර,  මාස ...), But when it comes to
ago, before and after, '*කට' *should be used without a leading space.
Otherwise it has a different meaning.

One more thing, I think i can make that translation more meaningful. I'll
update the PR.

MalindaWMD:
No need for any other modifications since we cannot make this translations 100% perfect in every language. We can proceed without any updates.

kylekatarnls:
OK for the spacing rule. But the unit tests need to pass to merge the PR:
https://travis-ci.org/github/briannesbitt/Carbon/builds/689719927

MalindaWMD:
Yes. I'll update it, but that's not the way it's used in Sinhala. 
It's okay since we have to stick with the rules. 👍

kylekatarnls:
If the user use the code:
```php
Carbon::parse('2018-02-10 00:00:00')->isoFormat('h:mm A')
```
We can't deliver `A h:mm`, that's not a Carbon library choice, that's up to the user to decide to use `Carbon::parse('2018-02-10 00:00:00')->isoFormat('LT')` instead.

The same apply for my own language/region: here in France, we don't use meridiem at all. No one never says AM/PM or any equivalent, we count hours from 0 to 23 and it's just clear enough. :)

But if some user `->isoFormat('h:mm A')` it will become 5:30 PM in French which will just sounds "American" and not easy to read for French people too. If the developer writes ` ->isoFormat('LT')` it will output `පෙ.ව. 5:30` n Sinhala and it will just displays `17:30` in French.

But still unit tests should check that meridiem properly display rather it's commonly used in the current language or not, while the standard formats and orders is driven by the `formats` array in translations files.

src/Carbon/Lang/si.php
```diff
@@ -18,28 +18,28 @@
  */
 return [
     'year' => '{1}වසර 1|වසර :count',
-    'a_year' => '{1}වසරක්|වසර :count',
+    'a_year' => '{1}වසර 1|වසර :count',
```
kylekatarnls, line 21: a_year in the singular form is supposed to be a translation for `"a year"` with no number.

MalindaWMD:
It's more meaningful with the update. BTW not sure why this was passed in
local when i was testing it.

MalindaWMD:
Sorry mate for the delay. Was at work.

In Sinhala translations, those terms should be updated. 
I just took a deep look and tested with different scenarios with the Sinhala translations. What happens was either it just failed unit tests( While trying to use more common translations) or it just returned the wrong result for some formats. 

So what I understand is we cannot localize results 100% while following other standards in PHP and in the package itself. What I wanted to do was to make these results closer to the Sinhala language. Sadly looks like that's not possible. 

So I  believe it is good to close this PR without merging.

Sorry for your time.

kylekatarnls:
So removing the space is not better in every contexts?

MalindaWMD:
That's right.

kylekatarnls:
OK, thank you for you help. What I can recommend is handling context in an upper function in this case such as:

```php
function contextualizeSinhalaCarbonString(string $carbonDateString, string $context)
{
    switch ($context) {
        case 'START_OF_SENTENCE': // or any kind of grammar identifier
            // Then replace with str_replace, strtr or preg_replace:
            return strtr($carbonDateString, [
                'පස් වරු' => 'පෙරවරු',
            ]);
    }
}
```

PR #2094, opened 2020-05-23
Fix #2092 Support yesterday and tomorrow in Carbon::parseFromLocale()

### LINKED ISSUE #2092: Support today, yesterday and tomorrow in Carbon::parseFromLocale()
With the following code:
```php
echo Carbon::parseFromLocale('demain', 'fr'); // "demain" is "tomorrow" in French
```

Carbon version: **2.34.2**

PHP version: **7.4**

I expected to get:

```
2020-05-22 00:00:00
```

But I actually get:

```
Error: DateTime::__construct(): Failed to parse time string (demain) at position 0 (d): The timezone could not be found in the database
```

It happens because "tomorrow" is not part of the words translated in `translateTimeString()`
https://github.com/briannesbitt/Carbon/blob/06d48c6f4204564399a1886ed91046aba4392488/src/Carbon/Traits/Localization.php#L384

While it's actually available in our translation files for many languages:
https://github.com/briannesbitt/Carbon/blob/06d48c6f4204564399a1886ed91046aba4392488/src/Carbon/Lang/fr.php#L56

Based on the available translations, we could understand in `parseFromLocale`:
- yesterday
- today (may be a problem, we actually don't have the exact translation of "today" in our languages file, maybe it can be calculated from calendar.sameDay)
- tomorrow
- before yesterday
- after tomorrow
- just now

Original idea: https://github.com/jenssegers/date/issues/216

This feature is open to pull-requests if someone want to give a try.

Doudou-A:
With the following code:

```php
echo Carbon::parseFromLocale('demain', 'fr');
```

I get `2020-05-24 00:00:00` as expected

With the code :
```php
echo Carbon::parseFromLocale('demain', 'en');
```
I get an error as expected

But now the code :
```php
Carbon::parseFromLocale('tomorrow', 'fr');
```
I don't get error, I get :
`2020-05-24 00:00:00`

Fix #2092


kylekatarnls:
Thank you, I will review this very soon.

kylekatarnls:
About the English `"tomorrow"` which does not fail with locale `"fr"`, it's actually due to the way it's done under the hood. `parseFromLocale` calls `translateTimeString` which translate words in the given language in the string to the English equivalent word, then the English string is used to create the new date using the native constructor of `DateTime`. So with any language, words in English will still work. It's the same with other words you can put either "dimanche" or "Sunday" when parsing with "fr".

kylekatarnls:
[This commit](https://github.com/briannesbitt/Carbon/commit/2a983e545f096aa1eca401f7ba6e8acd2152cc4c) may help you to use `'diff_now'`, `'diff_before_yesterday'` and `'diff_after_tomorrow'` in translations too.

PR #2107, opened 2020-06-08
Fix #2106 Add full somali locale

### LINKED ISSUE #2106: [Translation tool] so Improved
|English|so|comments|
|--|--|--|
|2 years|2 sanad|2 sano|
|2yrs|2 sanad|2 sanadood|
|1 year|1 sanad|1 sano|
|a year|1 sanad|sanad|
|2 months|2 Bilaha|2 bilood|
|2mos|2 Bilaha|2 bil|
|1 month|1 Bilaha|1 bil|
|a month|1 Bilaha|bil|
|2w|2 todobaad|2t|
|a week|1 todobaad|todobaad|
|2d|2 maalin|1m|
|a day|1 maalin|maalin|
|2 hours|2 saacad|2 saac|
|2h|2 saacad|2s|
|an hour|1 saacad|saacad|
|2m|2 daqiiqad|2d|
|a minute|1 daqiiqad|daqiiqad|
|2s|2 ilbiriqsi|2 il|
|1 second|1 ilbiriqsi|ilbiriqsi|
|a few seconds|1 ilbiriqsi|ilbiriqsiyo|
|January|Bisha Koobaad|Janaayo|
|Jan|Kob|Jan|
|February|Bisha Labaad|Febraayo|
|Feb|Lab|Feb|
|March|Bisha Saddexaad|Maarso|
|Mar|Sad|Mar|
|April|Bisha Afraad|Abriil|
|Apr|Afr|Abr|
|May|Bisha Shanaad|Maajo|
|May|Sha|Mjo|
|June|Bisha Lixaad|Juun|
|Jun|Lix|Jun|
|July|Bisha Todobaad|Luuliyo|
|Jul|Tod|Lyo|
|August|Bisha Sideedaad|Agoosto|
|Aug|Sid|Agt|
|September|Bisha Sagaalaad|Sebteembar|
|Sep|Sag|Seb|
|October|Bisha Tobnaad|Oktoobar|
|Oct|Tob|Okt|
|November|Bisha Kow iyo Tobnaad|Nofeembar|
|Nov|KIT|Nof|
|December|Bisha Laba iyo Tobnaad|Diseembar|
|Dec|LIT|Dis|
|2 hours ago|2 saacad ago|2 saac kahor|
|2 hours from now|2 saacad from now|2 saac ayaa kasoo wareegtay|
|2 hours before|2 saacad before|2 saac kohor|
|2 hours after|2 saacad after|2 saac kedib|

Abdifatahz:
I added full somali translation, and it fully works for diffFOrHumans time format.

Fix #2106


kylekatarnls:
Thanks a lot for your help.

So I guess so_SO is wrong too?
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/so_SO.php

Abdifatahz:
@kylekatarnls Thank you so much
Yeah so_So is wrong.

byjml:
Brilliant stuff, @Abdifatahz 

Two things I want to point out.

1.  As Somali is mainly spoken in the same region, `so_SO`, `so_ET`, `so_DJ` and `so_KE` do not make difference. `so` is enough as @Abdifatahz mentioned.

2. `ar_SO` does not exist. Wikipedia and other sources include Somalia in the Arabic speaking countries due to being politically aligned to Arab League countries but the reality is that arabbic is not local language.

kylekatarnls:
The locale is not just about the language. It's language+region. en_US and en_GB speak the same language (ignoring some s/z and minor differences) but they don't have the same first day of week, not the same date format. And as per the data we currently use (extracted mainly from CLDR), the `first_day_of_week` is not the same for every regions. If this is correct, then we need to keep all those locales, but all `so_*` can extends the `so` locale, so they would change only formats and the words could be fixed globally in `so.php`.

I will do that and update the unit tests then merge this PR.

The locale list can include historical or no longer spoken languages and are kept for backward-compatibility. For instance "la" locale for "Latin" (no longer spoken for centuries) still exists in the ISO 639-1. It can contain non-official dialects, it's nothing political, "ab_XY" is just convenient translations and settings for someone speaking/reading "ab" language in the "XY" region. It could potentially assemble any region with any language.

Our list is based on http://cldr.unicode.org/ which aims to:
> provides key building blocks for software to support the world's languages, with the largest and most extensive standard repository of locale data available

src/Carbon/Lang/so.php
```diff
@@ -10,6 +10,50 @@
...
 
 /*
- * Unknown default region, use the first alphabetically.
+ * Author:
+ * - Abdifatah Abdilahi(@abdifatahz)
  */
-return require __DIR__.'/so_DJ.php';
+return [
+    'year' => '{1}:count sanad|{0}:count sanadood|]1,Inf[:count sanadood',
+    'a_year' => '{1}a sanad|{0}:count sanadood|]1,Inf[:count sanadood',
```
kylekatarnls, line 18: According to what you wrote in #2106 "a year" should be "sanad", so I will remove all the "a" which looks like coming from the English version.

src/Carbon/Lang/so.php
```diff
@@ -10,6 +10,50 @@
...
+    'month' => '{1}:count bil|{0}:count bilood|]1,Inf[:count bilood',
+    'a_month' => '{1}a bil|{0}:count bilood|]1,Inf[:count bilood',
+    'm' => '{1}:countbil|{0}:countbil|]1,Inf[:countbil',
+    'week' => '{1}:count isbuuc|{0}:count isbuuc|]1,Inf[:count isbuuc',
+    'a_week' => '{1}a week|{0}:count isbuuc|]1,Inf[:count isbuuc',
+    'w' => ':countis',
+    'day' => '{1}:count maalin|{0}:count maalmood|]1,Inf[:count maalmood',
+    'a_day' => '{1}maalin|{0}:count maalmood|]1,Inf[:count maalmood',
+    'd' => ':countml',
+    'hour' => '{1}:count saac|{0}:count saac|]1,Inf[:count saac',
```
kylekatarnls, line 29: When the translation is the same for any number, you don't need to use the pluralization syntax:
```php
'hour' => ':count saac',
```

src/Carbon/Lang/so.php
```diff
@@ -10,6 +10,50 @@
  */
 
 /*
- * Unknown default region, use the first alphabetically.
+ * Author:
+ * - Abdifatah Abdilahi(@abdifatahz)
  */
-return require __DIR__.'/so_DJ.php';
+return [
+    'year' => '{1}:count sanad|{0}:count sanadood|]1,Inf[:count sanadood',
```
kylekatarnls, line 17: This is the default pluralization for "so" from symfony/translation, it means you can write:
```php
'year' => ':count sanad|:count sanadood',
```
It output `sanad` for `1` and `sanadood` for any other number.

kylekatarnls:
@byjml @Abdifatahz I updated all the so_* regions and updated unit tests + added missing translations for period stuff.

Could you please take a look at [the resulting SoTest file](https://github.com/briannesbitt/Carbon/pull/2107/files#diff-2ca4e1bc10dbb46b0ec4fbb08251eafc)

which show in comment the code tested and in red the result before, in green the new result.

For instance `Carbon::now()->subDays(2)->diffForHumans()` was `'2 maalin  ago'` and becomes `'2 maalmood kahor'`.

It would be great if you could if all green lines are better than their red line.

Thank you.

src/Carbon/Lang/so.php
Abdifatahz: Yes that's right.

tests/Localization/SoDjTest.php
```diff
@@ -17,210 +17,210 @@ class SoDjTest extends LocalizationTestCase
 
     const CASES = [
         // Carbon::parse('2018-01-04 00:00:00')->addDays(1)->calendar(Carbon::parse('2018-01-04 00:00:00'))
-        'Tomorrow at 12:00 subaxnimo',
+        'Tomorrow at 12:00 AM',
```
Abdifatahz, line 20: Tommorow is 'Beri' in somali.
any AM is "GH"
any PM is 'GD'

kylekatarnls: I guess "at" should be translated too. Here is the default (English) translations:
```php
    'calendar' => [
        'sameDay' => '[Today at] LT',
        'nextDay' => '[Tomorrow at] LT',
        'nextWeek' => 'dddd [at] LT',
        'lastDay' => '[Yesterday at] LT',
        'lastWeek' => '[Last] dddd [at] LT',
        'sameElse' => 'L',
    ],
```
Could you please add and translate it in https://github.com/Abdifatahz/Carbon/blob/patch-1/src/Carbon/Lang/so.php

I can't find "GH" / "GD" anywhere, all the reference I found use AM or subaxnimo. Can you show some official sources?

Abdifatahz: Added Calendar.
Thank you @kylekatarnls .

Abdifatahz: You can skip GH and GD it's just shorthand for GelinHore < 12PM and gelinDanbe >= 12PM.

kylekatarnls: OK

PR #2124, opened 2020-06-21
Document 2.36

kylekatarnls:
- [x] PHPStan extension
- [x] New interval methods

### DIFF
--- a/docs/index.src.html
+++ b/docs/index.src.html
@@ -2098,8 +2098,8 @@ $dt->isMidday(); // check if hour is 12:00:00 (or other midday hour set with Car
         <code>modify()</code>, <code>add()</code> and <code>sub()</code>. <code>change()</code> is an enhanced version
         of <code>modify()</code> that can take <i>magical</i>
         date/time format string, <code>'last day of next month'</code>, that it parses and applies the modification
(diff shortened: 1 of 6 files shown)


docs/index.src.html
```diff
@@ -3442,6 +3437,20 @@ <h1 id="api-macro">Macro</h1>
 </code></pre>
     </p>
 
+    <p id="phpstan-support">We provide a PHPStan extension out of the box so it can understand macros and mixins.</p>
+
+    <p>If you're using Laravel, you should consider using <a href="https://github.com/nunomaduro/larastan">larastan</a>
+        which provides a complete support of Laravel features in PHPStan (including Carbon macros).
+        Else you can install <strong>phpstan/extension-installer</strong> if you don't have it:</p>
```
spawnia, line 3444: This installation method is pretty much unusable for me and many others, see https://github.com/phpstan/extension-installer/issues/17

How about putting the non-magical installation method first (even before hinting to use larastan)?

```html
Enable the extension by adding it to your `phpstan.neon` configuration:

<pre>code>
includes:
- vendor/nesbot/carbon/extension.neon
</pre></code>
```

kylekatarnls: I will change that.

PR #2128, opened 2020-07-02
Fix "Cannot call abstract method Carbon\CarbonInterface::hasMacro()" error

leo108:
If the variable `$className` is `CarbonInterface::class` itself or an abstract class that hasn't implemented the `getMethod` method yet, it will throw an error `Cannot call abstract method Carbon\CarbonInterface::hasMacro()`

### DIFF
--- a/src/Carbon/PHPStan/MacroScanner.php
+++ b/src/Carbon/PHPStan/MacroScanner.php
@@ -20,19 +20,9 @@ final class MacroScanner
      */
     public function hasMethod(string $className, string $methodName): bool
     {
-        $reflectionClass = new ReflectionClass($className);
-
-        if (!$reflectionClass->implementsInterface(CarbonInterface::class)) {
-            return false;
-        }
-
-        $hasMacroMethod = $reflectionClass->getMethod('hasMacro');
-
-        if ($hasMacroMethod->isAbstract()) {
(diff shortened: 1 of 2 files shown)


src/Carbon/PHPStan/MacroScanner.php
```diff
@@ -20,7 +20,19 @@ final class MacroScanner
...
+            return false;
+        }
+
+        $hasMacroMethod = $reflectionClass->getMethod('hasMacro');
+
+        if ($hasMacroMethod->isAbstract()) {
+            return false;
+        }
+
+        return $className::hasMacro($methodName);
```
kylekatarnls, line 35: Don't you think the following would be equivalent:
```suggestion
        return is_a($className, CarbonInterface::class, true) &&
            is_callable([$className, 'hasMacro']) &&
            $className::hasMacro($methodName);
```

leo108: Looks a better solution, thanks

leo108:
Hope this can be released soon, this issue breaks one of our projects.

kylekatarnls:
Released in [2.36.1](https://github.com/briannesbitt/Carbon/releases/tag/2.36.1)

PR #2130, opened 2020-07-03
Add constants for months (fixes #2127)

### LINKED ISSUE #2127: Constants for months just like we have for days?
Hello,

This is not a bug report but more of a question/suggestion.
Apologies upfront if maybe this isn't the correct place to reach out.

I'm wondering if you have considered having constants for months, just like `CarbonInterface` has for days.
Personally, I've come across several situations where it would be useful to have human readable constants for the months, just because it makes the code slightly more readable and understandable, eg:

```php
if ($month >= 2 && $month <= 7) {
    // apply some special business logic that should only happen between Feb and July.
}
```
vs
```php
if ($month >= CarbonInterface::FEBRUARY && $month <= CarbonInterface::JULY) {
    // apply some special business logic that should only happen between Feb and July.
}
```

I know theoretically you could achieve the same with something like `date_parse('July')['month']` or similar, but expressions can't (sadly) be defined as class constants and having those in `CarbonInterface` would be extremely helpful, imo.

Thanks!

nocive:


kylekatarnls:
Thank you. 👍

kylekatarnls:
Released in [2.36.1](https://github.com/briannesbitt/Carbon/releases/tag/2.36.1)

PR #2153, opened 2020-08-08
[Tryout][WIP] 3.x sf4.4

c33s:
just to use travis

### DIFF
--- a/src/Carbon/Traits/Localization.php
+++ b/src/Carbon/Traits/Localization.php
@@ -17,16 +17,10 @@ use Carbon\Language;
 use Carbon\Translator;
 use Closure;
 use Symfony\Component\Translation\TranslatorBagInterface;
-use Symfony\Component\Translation\TranslatorInterface;
 use Symfony\Contracts\Translation\LocaleAwareInterface;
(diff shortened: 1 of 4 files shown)


src/Carbon/Traits/Localization.php
```diff
@@ -17,16 +17,9 @@
...
 use Symfony\Contracts\Translation\LocaleAwareInterface;
-use Symfony\Contracts\Translation\TranslatorInterface as ContractsTranslatorInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
-if (!interface_exists('Symfony\\Component\\Translation\\TranslatorInterface')) {
-    class_alias(
-        'Symfony\\Contracts\\Translation\\TranslatorInterface',
-        'Symfony\\Component\\Translation\\TranslatorInterface'
-    );
-}
```
kylekatarnls, line 29: I though it was yet removed but it's actually still on an other branch:
https://github.com/briannesbitt/Carbon/compare/feature/issue-1967-week-based-on-locale?expand=1

Maybe it can be easier starting from this branch.

PR #2162, opened 2020-08-17
Use DateTime::ISO8601 for iso8601 instead of ATOM

mtawil:


mtawil:
> Hello. Thank you for your contribution.
> 
> HH:mm format for timezone is valid regarding the iso 8601 norm, and breaking change to master branch (2.x) is not possible. This can be discussed for 3.x branch. Can you please explain the benefit of HHmm over HH:mm? Thank you.

Hi @kylekatarnls,

This change is not about hh:mm, it's about timezone after hh:mm, and the different between iso8601 and atom is just the colon; it does not make any sense to get atom when we want iso8601.

kylekatarnls:
Re-read my message I talked about the timezone and go check the iso8601 norm, the colon is optional.

kylekatarnls:
From the PHP doc:

> Note: This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.

We should promote the norm, not the legacy incorrect format.

kylekatarnls:
For the record, there were a discussion about it (a long time ago) #976

PR #2164, opened 2020-08-19
🌎 Fix Estonian date formats based on CLDR data

ragulka:
This PR updates Estonian ISO date formats based on [CLDR data](https://github.com/unicode-cldr/cldr-dates-full/blob/204e99e8b28a52239bb6de1bbc89986cb19c0d50/main/et/ca-gregorian.json#L309-L314) current at the time of this PR. 

I'm submitting a PR because formats cannot be suggested using the translation tool.

### DIFF
--- a/src/Carbon/Lang/et.php
+++ b/src/Carbon/Lang/et.php
@@ -30,6 +30,7 @@
  * - Zahhar Kirillov
  * - João Magalhães
  * - Ingmar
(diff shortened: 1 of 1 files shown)


kylekatarnls (approved):
I checked and this sounds correct. I waiting for unit tests to run on Travis and it can be merged.

PR #2184, opened 2020-09-08
fix return type of createFromFormat()

YAhiru:
I fixed return types of `Carbon::createFromFormat()` and `CarbonImmutable::createFromFormat()`.

https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Traits/Creator.php#L551
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/CarbonInterface.php#L1064
https://www.php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-returnvalues


kylekatarnls (approved):
Thanks!

PR #2189, opened 2020-09-11
Use constants instead of 3600

szepeviktor:
...which is a _magic number_

### DIFF
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -593,7 +593,6 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
     public const DAYS_PER_WEEK = 7;
     public const HOURS_PER_DAY = 24;
     public const MINUTES_PER_HOUR = 60;
-    public const SECONDS_PER_HOUR = 3600;
     public const SECONDS_PER_MINUTE = 60;
     public const MILLISECONDS_PER_SECOND = 1000;
     public const MICROSECONDS_PER_MILLISECOND = 1000;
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Hello,

Thanks for the contribution. This is indeed not good and not consistent with the rest of the code which uses constants.

In `addRealUnit()`, we can currently find `static::MINUTES_PER_HOUR * static::SECONDS_PER_MINUTE` used to convert hours to seconds. And as it's using `static::` this allows users to extend the `Carbon` and override those constants. I can't see any practical reason to do this but still they can.

And the `->offsetHours` actually return:
```php
$this->getOffset() / static::SECONDS_PER_MINUTE / static::MINUTES_PER_HOUR
```

So as it's here the exact reverse operation: getting time zone seconds offset from hours offset, it would be symmetric to use:
```php
$originalTz * static::SECONDS_PER_MINUTE * static::MINUTES_PER_HOUR
```

And so having the same override possibility.

Side note: the factor from any unit to any other can be get with:
```php
CarbonInterval::getFactor('second', 'hour')
```

And if a user would need something like 2.5 hours in seconds, it could be:
```php
CarbonInterval::make('2.5 hours')->totalSeconds
```

Sure `2.5 * Carbon::SECONDS_PER_HOUR` is about 100 times faster as it doesn't involve any interval instance. But I'm not sure we should provide it as a util for user-land.

Else why not SECONDS_PER_DAY, MINUTES_PER_DAY, HOURS_PER_WEEKS, etc.? It would make a lot to cover all the possibly useful cases.

szepeviktor:
I've learned a lot.
Thank you.

kylekatarnls (approved):

kylekatarnls:
Thank you too.

szepeviktor:
@kylekatarnls Please be friends with https://github.com/povils/phpmnd

PR #2191, opened 2020-09-16
Adds a testing method called withTestNow

johncongdon:
This is similar to the internal testing helper "wrapWithTestNow" but meant to be publicly accessible to outside tests

### DIFF
--- a/src/Carbon/Traits/Test.php
+++ b/src/Carbon/Traits/Test.php
@@ -10,8 +10,6 @@
  */
 namespace Carbon\Traits;
 
-use Carbon\Carbon;
-use Carbon\CarbonImmutable;
 use Closure;
 use DateTimeImmutable;
 
@@ -68,11 +66,9 @@ trait Test
      */
     public static function withTestNow($testNow = null, $callback = null)
     {
-        Carbon::setTestNow($testNow);
-        CarbonImmutable::setTestNow($testNow);
+        self::setTestNow($testNow);
         $callback();
-        Carbon::setTestNow();
-        CarbonImmutable::setTestNow();
+        self::setTestNow();
     }
 
     /**
--- a/src/Carbon/Traits/Test.php
+++ b/src/Carbon/Traits/Test.php
@@ -55,6 +55,23 @@ trait Test
     }
 
     /**
+     * Temporarily sets a static date to be used within the callback.
+     * Using setTestNow to set the date, executing the callback, then
+     * clearing the test instance.
+     *
+     * /!\ Use this method for unit tests only.
+     *
+     * @param Closure|static|string|false|null $testNow real or mock Carbon instance
+     * @param Closure|null $callback
+     */
+    public static function withTestNow($testNow = null, $callback = null)
+    {
+        self::setTestNow($testNow);
+        $callback();
+        self::setTestNow();
+    }
+
+    /**
      * Get the Carbon instance (real or mock) to be returned when a "now"
      * instance is created.
      *
--- a/src/Carbon/Traits/Test.php
+++ b/src/Carbon/Traits/Test.php
@@ -10,6 +10,8 @@
  */
 namespace Carbon\Traits;
 
+use Carbon\Carbon;
+use Carbon\CarbonImmutable;
 use Closure;
 use DateTimeImmutable;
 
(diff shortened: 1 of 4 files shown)


kylekatarnls (changes requested):
Hello,

Thanks for your contribution. I completely see the need for it and this surely can help.

Still I do not feel comfortable with the fact to mutate both `Carbon` and `CarbonImmutable` from a method you call on either the first or the second.

It's close to the [`fakeAsync()`](https://github.com/kylekatarnls/carbonite#fakeasync-for-php) method of **Carbonite**. This library might be a better place for such testing util, mostly if you want to get both `Carbon` and `CarbonImmutable` mocked as that's what happen with `Carbonite::freeze()`

I think, the `Test` trait should provide a way to change only itself (consistent with `Carbon::setTestNow()` and other static setters like `Carbon::setLocale()`) and in `Carbonite` there could be a method to set both at once.

I also wonder if this should not need a try/catch for a better isolation, because:

```php
    public function test1()
    {
        Carbon::withTestNow('2000-01-01', function () {
            $this->fail('Test failed');
        });
    }

    public function test2()
    {
        echo Carbon::now(); // Still 2000-01-01
```

src/Carbon/Traits/Test.php
```diff
@@ -54,6 +56,25 @@ public static function setTestNow($testNow = null)
...
+     * @param Closure|static|string|false|null $testNow real or mock Carbon instance
+     * @param Closure|null $callback
+     */
+    public static function withTestNow($testNow = null, $callback = null)
+    {
+        Carbon::setTestNow($testNow);
+        CarbonImmutable::setTestNow($testNow);
+        $callback();
+        Carbon::setTestNow();
+        CarbonImmutable::setTestNow();
```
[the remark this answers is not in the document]
kylekatarnls: Thanks!

johncongdon:
@kylekatarnls good call on `static::` I forgot that setTestNow is already being called statically, so we know what class we are in.

The goal is to minimize the confusion in our tests and having to remember to clear out the test time between tests.

tests/Carbon/TestingAidsTest.php
```diff
@@ -269,4 +269,17 @@ public function testSetTestNow()
...
+    {
+        $self = $this;
+        $test_now = '2020-09-16 10:20:00';
+        Carbon::withTestNow($test_now, static function () use ($self, $test_now) {
+            $current_time = Carbon::now();
+            $self->assertSame($test_now, $current_time->format('Y-m-d H:i:s'));
+        });
+
+        $current_time = Carbon::now();
+        $this->assertNotEquals($test_now, $current_time->format('Y-m-d H:i:s'));
```
kylekatarnls, line 283: Actually I was expecting here to get the real now.

+ Please use camelCase for consistency: `$currentTime`, `$testNow`

johncongdon: I'll update variable names.  That's an habit from our local style guide.

And this does get the actual now.  I am testing that it doesn't equal the test time any more.  This would only fail one second if we ran this in the past. ;-)

kylekatarnls: 👍

tests/Carbon/TestingAidsTest.php
```diff
@@ -269,4 +269,17 @@ public function testSetTestNow()
...
+    {
+        $self = $this;
+        $testNow = '2020-09-16 10:20:00';
+        Carbon::withTestNow($testNow, static function () use ($self, $testNow) {
+            $currentTime = Carbon::now();
+            $self->assertSame($testNow, $currentTime->format('Y-m-d H:i:s'));
+        });
+
+        $current_time = Carbon::now();
+        $this->assertNotEquals($testNow, $current_time->format('Y-m-d H:i:s'));
```
kylekatarnls, line 283:
```suggestion
        $currentTime = Carbon::now();
        $this->assertNotEquals($testNow, $currentTime->format('Y-m-d H:i:s'));
```

tests/CarbonImmutable/TestingAidsTest.php
```diff
@@ -257,4 +257,17 @@ public function testCreateFromPartialFormatWithMicroseconds()
...
+    {
+        $self = $this;
+        $testNow = '2020-09-16 10:20:00';
+        Carbon::withTestNow($testNow, static function () use ($self, $testNow) {
+            $currentTime = Carbon::now();
+            $self->assertSame($testNow, $currentTime->format('Y-m-d H:i:s'));
+        });
+
+        $current_time = Carbon::now();
+        $this->assertNotEquals($testNow, $current_time->format('Y-m-d H:i:s'));
```
kylekatarnls, line 271:
```suggestion
        $currentTime = Carbon::now();
        $this->assertNotEquals($testNow, $currentTime->format('Y-m-d H:i:s'));
```

kylekatarnls (approved):

PR #2193, opened 2020-09-18
Fix for formats that contains microseconds, as in informix databases

mkrohn:
Hi Brian !!
I want to congratulate you for this great piece of software.
I had to patch to work with Laravel 5.8 + abram odbc + informix, because informix sends microseconds in timestamps.
In Creator.php
createFromFormat fails when $format is Y-m-d H:i:s.u and $time does not include microseconds.
Mi change appends zero microseconds (.000) to $time

Regarding phpunit, it gave me errors even without changing anything, so I think this little patch is inocuous and useful.

I hope my contribution is useful to the comunity.
Best regards 

Martin

### DIFF
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -659,6 +659,10 @@ trait Creator
         $function = static::$createFromFormatFunction;
 
         if (!$function) {
+            if ($format === 'Y-m-d H:i:s.u' && strstr($time, '.') == null )
+            {
+                $time .= '.000';
+            }
             return static::rawCreateFromFormat($format, $time, $tz);
         }
 


kylekatarnls:
Hi @mkrohn this has been fixed in Laravel 7.0:
https://github.com/laravel/framework/blob/43bea00fd27c76c01fd009e46725a54885f4d2a5/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L918

And this fix is the relevant one. Carbon should not try to guess that the given format is wrong. The purpose of createFromFormat() is to ensure the given format is respected and create the date only if the string match it. To create a date without format constraint, `Carbon::parse()` should be used instead.

So please upgrade to a recent version of Laravel.

Thanks for your proposal.

PR #2200, opened 2020-09-23
#2199 create from microtime

### LINKED ISSUE #2199: add create methods from float and string types microtime result
Add create methods from float and string types microtime result

```php
Carbon::createFromFloatMicrotime(float $microtime);
Carbon::createFromWhitespaceMicrotime(string $microtime);
```

Carbon version: **2.40.1**

PHP version: **7.4**

userator:
Fix #2199

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ coverage
 phpunit.xml
 Carbon-*.zip
 *.phpunit.result.cache
+/nbproject/
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -1165,22 +1165,24 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
 
     /**
      * Create a Carbon instance from a microtime as float, format XXXXXXXXXX.XXXX.
-     * 
+     *
      * @param float $microtime
      * @param \DateTimeZone|string|null $tz
+     *
      * @return static
      */
     public static function createFromMicrotimeFloat(float $microtime, $tz = null);
 
     /**
      * Create a Carbon instance from a microtime as string, format X.XXXXXXXX XXXXXXXXXX.
-     * 
+     *
      * @param string $microtime
      * @param \DateTimeZone|string|null $tz
+     *
      * @return static
      */
     public static function createFromMicrotimeWhitespace(string $microtime, $tz = null);
-    
+
     /**
      * Create a Carbon instance from just a date. The time portion is set to midnight.
      *
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -1164,6 +1164,24 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
     public static function createFromTimestampUTC($timestamp);
 
     /**
+     * Create a Carbon instance from a microtime as float, format XXXXXXXXXX.XXXX.
+     * 
+     * @param float $microtime
+     * @param \DateTimeZone|string|null $tz
+     * @return static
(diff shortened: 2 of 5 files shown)


kylekatarnls:
Hello,

Thanks for your contribution. This sounds good to me. With this change, we'll have 4 methods to create date from different kinds of timestamps, this can make the API a bit confusing at a first sight. I will take the time to see if there is any way to improve/unify a bit those methods.

userator:
Hi Kyle, I wanted to add two methods for two views of the php microtime () function result
Thank you for participating :)

kylekatarnls:
I did get why you propose this adding. But I'm not sure creating `Carbon` instance from `microtime()`'s output is something we should promote. `microtime()` produces a reversed string (decimal first, integer then) with extra-zeros, this is not standard a way less meaningful than the ISO8601 which is the standard to promote to exchange date/time/timestamp of any precision. And `microtime(true)` produces a `float` that is not able to store the complete precision of timestamp microseconds (before PHP 8).

That's why I want to take the time thinking about naming of those methods and if this is widely needed or more like an edge-case use that could be better fulfilled with a mixin/macros before mergin.

kylekatarnls:
I analyse this deeper. First thing is I still consider it's a good feature we can provide in a way or an other.

Here are some issues to address/investigate:

## Rounding

I just checked with a 32-bit version of PHP and there is not the rounding I suspected. But number_format does an rounding I did not expect when giving a string:

```php
number_format('1600887164.88951247', 6, '.', '')
```
Output:
```
1600887164.889513
```
I was expecting 2 instead of 3 at the end. (This actually seems to reach the maximum precision and 47 becomes 53...). When we get a `float` we no longer can recover the lost precision, but it would be awesome if we could pass a U.u string with any precision loss.

I will check if we can do a better rounding from string.

## Specific methods vs generic method

I also wonder if we can have something more generic which would accept either X.X (as string or float) or "0.X X" with any number of decimal. It could be achieved with:

```php
    public static function createFromMicrotime($microtime, $tz = null)
    {
        if (!is_float($microtime) && !is_int($microtime)) {
            $microtime = preg_replace('/^0\.(\d+)\s+(\d+)$/', '$2.$1', $microtime);
        }

        return static::rawCreateFromFormat('U.u', number_format($microtime, 6, '.', ''))
                ->setTimezone($tz);
    }
```

And we could push further the abstraction by allowing all those in `createFromTimestamp` and `createFromTimestampUTC`.

## Timezone

Timestamp is UTC by definition (e.g. Timestamp 0 is 1970-01-01 00:00 UTC), that means it results in a different date-time depending on the timezone.

When the default timezone (set with `date_default_timezone_set`) is not UTC, `Carbon::createFromTimestamp()` actually does a trick to set the timezone (if not specified) after creating the instance. The `->setTimezone($tz)` actually does the same (if `$tz` is `null`, it falls back to the default timezone).

If we merge `createFromMicrotime` with existing method, it could also benefit from this UTC option. And while we're there, we could in the same move make `createFromTimestampMs` benefit of both (decimal part and UTC option).

kylekatarnls:
As I made many changes, I finally restored your PR in its initial state and created an other PR which still includes your commits: #2205

The idea is:
- to implement your feature right inside `createFromTimestampUTC` and `createFromTimestamp` which could then get either int, float or string.
- allow to passe U.u string and preserve the precision
- allow a more generic multi-numbers string by proceeding an addition with every numbers of a string which supports the output of `microtime()` but also strings with more than 2 numbers and more or less decimal digits
- make every timestamp method including the setters benefit of the `float|int|string` support

Your review on it would be very welcome.

PR #2216, opened 2020-10-08
Error diffFiltered(): "Unknown getter 'locale'"

FabrizioCafolla:
When $property equal 'locale' not found class attribute. 
Not sure if the 'localTranslator' property is correct. But surely 'locale' doesn't work.


kylekatarnls:
This breaks the unit tests. So it requires explainations for breaking changes I will also need a unit test that explains the feature/fix. So it may be better to first fill an issue with all required informations such as a minimal code to reproduce your case and your expectation, so we can understand the point to solve then discuss the appropriate solution.

PR #2235, opened 2020-11-09
Update .gitattributes

bytestream:
Exclude `extension.neon` from composer

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,6 +6,7 @@
 /.styleci.yml       export-ignore
 /.travis.yml        export-ignore
 /build.php          export-ignore
+/extension.neon     export-ignore
 /phpdoc.php         export-ignore
 /phpstan.neon       export-ignore
 /phpunit.xml.dist   export-ignore


kylekatarnls:
Please give details about the problem you have and why you opted for this solution.

AFAIK this will break #1916

bytestream:
> 
> 
> Please give details about the problem you have and why you opted for this solution.
> 
> AFAIK this will break #1916

I naively assumed that this file is not intended for production use and could be removed. If it's intended for production use then I apologise for my assumption and you can close this PR.

kylekatarnls:
This is not meant for production. But still this is part of the distributed package. Actually libraries like ours has no such thing like a production stage/environment.

As you may notice there is a `phpstan.neon` file we already exclude from the package. this file is PHPStan configuration to develop the Carbon library. You need it only to contribute to Carbon, which imply to download the source, so it's not useful in the package.

The `extension.neon` is a file to be used in development environment that **use** Carbon. So it's useful for you as it will help PHPStan (if you use it) to understand Carbon macros and so to get relevant reports when you run phpstan command. This mean we should distribute it. So you can have it in your development environment without download every package as source (which would be very heavy and slow).

In production, this file will not be used. It will just be dead code. It can't do any harm as it's nothing secret and not executable and likely your whole vendor directory is not accessible directly in production. So it's just 1.5KB of dead code. You still can "blacklist" with some production cleanup optimization if you like. But there is probably more efficient ways to save space if this is an issue for you such as removing after installation all the language files you does not need or the "tests" directories some of your packages was possibly not excluded via .gitattributes.

Thanks for your contribution.

bytestream:
Gotcha - thanks for the explanation. I just assumed it was similar to `phpstan.neon` which can be excluded as you mentioned.

PR #2242, opened 2020-11-28
Add GitHub Workflows

kylekatarnls:
Replace TravisCI with GitHub Actions

Result of the experiment:

| Criterion | TravisCI | GitHub Actions |
|---|---|---|
| Speed |   |   |
| Result readability |   |   |
| Featuring |   |   |


kylekatarnls:
The result of the test is that GitHub Actions is a bit faster for the same job. But most importantly TravisCI waiting time increased a lot recently while GitHub Actions seems to be able (for now) to run more jobs in parallel so waiting time is way shorter and we get the test results much faster.

Issue #2135, opened 2020-07-21
PHPStan complains about Carbon::create suddenly returning false?


ruudk:
Hello,

I encountered an issue with the following code:
```php
        Carbon::setTestNow(Carbon::create(2017, 4, 4, 13, 58));
```

Carbon version: 2.36.1
PHP version: 7.4

PHPStan complains everywhere I use the code snippet above:
```
  24     Parameter #1 $testNow of static method Carbon\Carbon::setTestNow()
         expects Carbon\Carbon|Closure|string|null, Carbon\Carbon|false given.
```

This has been introduced in https://github.com/briannesbitt/Carbon/pull/1916.

I don't get why `create` can 1) return a date, 2) throw an exception or 3) return false? It's not in the interface, but only in the implementation.

kylekatarnls:
Hello, first, you can write your date right in `setTestNow()`:
```php
Carbon::setTestNow('2017-04-04 13:58');
```
That's probably the more readable way to define it anyway.

> It's not in the interface, but only in the implementation

This should be in the interface too because `create()` can actually return `false`, for instance with:
```php
Carbon::useStrictMode(false);
var_dump(Carbon::create(2017, 914, 4, 13, 58));
```

With the strict mode, it would throw an exception. In the normal case, it return the date, that's for the answers to your 1/2/3.

And `setTestNow()` wouldn't handle `false` properly,

So PHPStan seems to do its job rather fairly. And while we obviously know `Carbon::create(2017, 4, 4, 13, 58)` is static and so will never return `false`, PHPStan can't guess it and we can't send this info to PHPStan neither.

Indeed this was not a problem before #1916 but only because PHPStan was completely unaware of the `false` possibility.

We don't have too many options to make things better.
- revert the typing => this actually would be a falsification (hidden what can happen in the reality just to make things to look more compatible than they actually are).
- handling false in `setTestNow()` => does not make sense and would imply to do the same in many other methods. But for this particular case it would just imply to replace `return static::getTestNow() !== null;` with `return !!static::getTestNow();` to handle `false` as an equivalent for `null` and that could be acceptable.

ruudk:
Would be great if strict mode could be enabled by default (in a new major) and then get rid of the `false` returns.

kylekatarnls:
strict mode **is** enabled by default. If you didn't change it, `create()` will throw exceptions and so will never return `false`, but PHPStan can't know if the strict mode had been disabled or not.

[...]

PR #2254, opened 2020-12-16
Implemented Create Strict API

i-bajrai:
Realising that this new version has the same issue as 2.x, this method might be useful for creating a Carbon interface only.

### DIFF
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -1065,7 +1065,6 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
      */
     public static function createStrict($year = 0, $month = 1, $day = 1, $hour = 0, $minute = 0, $second = 0, $tz = null);
 
-
     /**
      * Create a Carbon instance from just a date. The time portion is set to now.
      *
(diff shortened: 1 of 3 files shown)


i-bajrai:
Referencing https://github.com/briannesbitt/Carbon/issues/2135

src/Carbon/CarbonInterface.php
```diff
@@ -1038,6 +1038,33 @@ public function copy();
...
+     * @param int|null                 $month
+     * @param int|null                 $day
+     * @param int|null                 $hour
+     * @param int|null                 $minute
+     * @param int|null                 $second
+     * @param DateTimeZone|string|null $tz
+     *
+     * @throws InvalidFormatException
+     *
+     * @return static|false
```
i-bajrai, line 1064:
```suggestion
     * @return static
```

kylekatarnls (commented):
Strict mode applies to several methods and I'm not sure we should have a special handling for `create`.

Actually using `Carbon::useStrictMode(true);` globally for your app is recommended. From there if wanting strict mode, is there any particular reason to want it on `create` and not on other methods? it does not sound very consistent to me.

kylekatarnls:
> Realising that this new version has the same issue as 2.x

This version (master) is 2.x. 3.x is the next version and has its own branch, it will became the master when a first alpha version will be released.

kylekatarnls:
Closing for now. If particular need for `createStrict` (specific to `create`) in 2.x version is clarified, I would re-open.

kylekatarnls:
For the record, a macro for this can be found here:
https://github.com/briannesbitt/Carbon/issues/2135#issuecomment-745940619

PR #2258, opened 2020-12-25
Dedupe versions in CI tests

enomotodev:


kylekatarnls:
Thanks for the contribution. I also see that now, PHPUnit supports PHP 8 so we may be able to get rid of the ignore-platform-reqs. I'll take the occasion to clean it up.

PR #2274, opened 2021-02-01
Fix #2091 Latvian updates

### LINKED ISSUE #2091: 🌎 [Translation tool] lv - Latvian
|English|lv|comments|
|--|--|--|
|2 years|2 gadiem|2 gadi / Because '2 gadiem' is Dativum and it means 'For whom' .. 'For 2 years'|
|2 months|2 mēnešiem|2 mēneši .. Because '2 mēnešiem' means 'for 2 months'.|
|1 month|1 mēneša|1 mēnesis|
|a month|mēneša|mēnesis|
|2 weeks|2 nedēļām|2 nedēļas|
|1 week|1 nedēļas|1 nedēļa|
|2 days|2 dienām|2 dienas|
|2d|2 dienām|2 d.|
|1 day|1 dienas|1 diena|
|2 hours|2 stundām|2 stundas|
|1 hour|1 stundas|1 stunda|
|2 minutes|2 minūtēm|2 minūtes|
|2 seconds|2 sekundēm|2 sekundes|
|1 second|1 sekundes|1 sekunde|
|January|janvārī|janvāris. Because 'janvārī' means where (Lokativum) - 'in January'|
|February|februārī|februāris|
|March|martā|marts|
|April|aprīlī|aprīlis|
|May|maijā|maijs|
|June|jūnijā|jūnijs|
|July|jūlijā|jūlijs|
|August|augustā|augusts|
|September|septembrī|septembris|
|October|oktobrī|oktobris|
|November|novembrī|novembris|
|December|decembrī|decembris|
|2 hours before|2 stundas pirms|pirms 2 stundām|

----
In Latvian prepositions like 'in' are shown at the end of the noun. For instance, 'January' is in Nominativum - 'Janvāris'. However, if you would like to say 'I started a project in January' - you would have to say '.. janvārī', and 'ī' shows 'in'. In the January'.

If it helps, I'm ready to assist with Latvian translation also further. Just write me an email.

jurchiks:
Fixing Latvian localization.
Closes #2091.

I'm not entirely sure if master is the right branch for this.


jurchiks:
After fixing this, I can see that test coverage is quite lacking.
`a_{unit}_before/after` is not covered in the case of `{1}`, month names and days of week are not covered at all, so it's imposible to say if it actually works as expected.

src/Carbon/Lang/lv.php
```diff
@@ -31,72 +32,111 @@
...
-    'ago' => 'pirms :time',
-    'from_now' => 'pēc :time',
+    'a_second' => '{1} sekunde|0 sekunžu|:count sekunde|:count sekundes',
 
     'after' => ':time vēlāk',
-    'before' => ':time pirms',
-
     'year_after' => '0 gadus|:count gadu|:count gadus',
-    'a_year_after' => '{1}gadu|0 gadus|:count gadu|:count gadus',
+    'a_year_after' => '{1} gadu|0 gadus|:count gadu|:count gadus',
```
jurchiks, line 99: Can't tell if the space should be here or not, because there are no tests for this and I don't know how to make Carbon output this.

kylekatarnls (approved):

tests/Localization/LvLvTest.php
```diff
@@ -21,119 +21,119 @@ class LvLvTest extends LocalizationTestCase
     public const CASES = [
         // Carbon::parse('2018-01-04 00:00:00')->addDays(1)->calendar(Carbon::parse('2018-01-04 00:00:00'))
         // 'Tomorrow at 12:00 AM',
-        'Tomorrow at 00:00',
+        'rīt plkst. 00:00',
```
jurchiks, line 24: lv_LV will never be different from just lv, it is the same exact dialect. Why the duplication?

kylekatarnls: Multiple reasons:
- Consistency with all other locales (either we refactor them all or none, but I want to avoid a mix)
- When future changes will apply, we'll be able to see the before/after difference for each impacted locale.
- "will never", let's say Latvia is an immutable country with immutable borders and immutable languages, but the generic idea is to consider languages and regions evolve. And while there are countries having only 1 official language and this language being official in only 1 country, we also have plenty of languages being official in multiple countries and countries having multiple official languages. For instance, while `de_DE` represent the official German language as spoken in Germany, `de` represent the German more common base across all regions speaking German. If Germany officially change its date format, we'll update `de_DE` but not `de`.
- All localization tests follow the same pattern, so you can easily compare outputs of 2 locales, and regenerate them by scripting.
- Last generic concept is to avoid to mirroring source code in tests code, as it increases the risk of mirroring errors. For instance: `assert(4, sum(2, 2))` over `assert(2 + 2, sum(2, 2))`

PR #2276, opened 2021-02-04
Proposed change for fixing #2217

### LINKED ISSUE #2217: Exception: Unknown getter 'locale'
Hello,

```      
$holidays = ['0' => '2020-12-25', '1' => '2020-12-24']; //etc..
            
return $start_date->diffInDaysFiltered(function (Carbon $d) use ($holidays) {
   return $d->isWeekday() && !array_search($d->format("Y-m-d"), $holidays,true);
}, $end_date);
```

Carbon version: **2.41.0**

PHP version: **7.3.5**

I expected to get:

```
number of diff days between start_date and end_date
```

But I actually get:

```
Exception: Unknown getter 'locale'
```

**With version 2.40.1 work!**

makurtu:
Well, if using `property_exists` before trying to acces prop, we can assure that it exists and no exception is triggered.
Closes  #2217

### DIFF
--- a/src/Carbon/Traits/Options.php
+++ b/src/Carbon/Traits/Options.php
@@ -413,7 +413,8 @@ trait Options
             'localFormatFunction' => 'formatFunction',
         ];
         foreach ($map as $property => $key) {
-            $value = $this->$property ?? null;
+            if (property_exists($this, $property)) $value = $this->$property ?? null;
+                else $value = null;
             if ($value !== null) {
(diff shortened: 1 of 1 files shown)


kylekatarnls (changes requested):
It won't do it, As it retrieve some from magic getter, the properties don't exist as declared on the object. That's why this PR break the unit tests, it's basically almost the same than removing the line.

PR #2288, opened 2021-02-11
Makes PHPDoc consistent

alterphp:
In order to make static analysis tools happier (PHPStan at least)


kylekatarnls:
Thanks. Indeed it was never added since it's available.

PR #2290, opened 2021-02-18
[Docs] Add toDateTimeString() to common formats

danilopolani:


kylekatarnls:
`toDateTimeString()` is already mentioned 21 times in the doc including once just above this section, I'm not sure we need one more. Then this section is about methods for the formats related to https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.types (so `toDateTimeString()` is not in those ones).

Eventually, if adding something about `toDateTimeString()` and `toTimeString()` it would be nice to mention they have a optional `$precision` parameter:
```php
$dt->toDateTimeString('microsecond')
$dt->toTimeString('minute')
```

danilopolani:
You're right, didn't notice that section were just about DateTime format wrappers. I think that mention `toDateTimeString()` inside `String Formatting` is enough. 

Sorry for the waste of time :eyes:

PR #2300, opened 2021-03-10
fix #2299 Fatal Error occurs when calling msgpack_unpack() function

### LINKED ISSUE #2299: Fatal Error occurs when calling msgpack_unpack() function in PHP 7.4 or higher environment.
Hello,
Fatal Error occurs when calling msgpack_unpack() function in PHP 7.4 or higher environment.

The msgpack_unpack() function can be used if pecl msgpack is enabled.
https://pecl.php.net/package/msgpack

```php
$now = \Carbon\Carbon::parse('2018-06-01');
$message = @msgpack_pack($now);
$copy = msgpack_unpack($message);
```

Carbon version: **2.46.0**

PHP version: **7.4.14**

pecl msgpack version:**msgpack-2.1.2**
https://pecl.php.net/package/msgpack

I expected to get:

```
PHP Fatal error:  Uncaught Error: Invalid serialization data for DateTime object in ./vendor/nesbot/carbon/src/Carbon/Traits/Serialization.php:131
Stack trace:
#0 ./vendor/nesbot/carbon/src/Carbon/Traits/Serialization.php(131): DateTime->__wakeup()
#1 [internal function]: Carbon\Carbon->__wakeup()
#2 ./test.php(6): msgpack_unpack('\x84\xC0\xADCarbon\\Carbo...')
#3 {main}
  thrown in ./vendor/nesbot/carbon/src/Carbon/Traits/Serialization.php on line 131
```

But I actually get:

```
$now  = $copy;
```

Thanks!

suzunone:
Fixing  Fatal Error occurs when calling msgpack_unpack() function.
Closes #2299.

### DIFF
--- a/src/Carbon/Traits/Serialization.php
+++ b/src/Carbon/Traits/Serialization.php
@@ -131,7 +131,7 @@ trait Serialization
         // FatalError occurs when calling __wakeup method in PHP 7.4 or later.
         // @codeCoverageIgnoreStart
         if (version_compare(PHP_VERSION, '7.4.0-dev', '>=')) {
-            parent::__construct($this->date, $this->timezone !== null ? new DateTimeZone($this->timezone) : null);
+            parent::__construct($this->{$this->dumpProperties[0]}, $this->timezone !== null ? new DateTimeZone($this->timezone) : null);
         } elseif (get_parent_class() && method_exists(parent::class, '__wakeup')) {
             parent::__wakeup();
         }
--- a/src/Carbon/Traits/Serialization.php
+++ b/src/Carbon/Traits/Serialization.php
@@ -11,6 +11,7 @@
 namespace Carbon\Traits;
 
 use Carbon\Exceptions\InvalidFormatException;
+use DateTimeZone;
 
 /**
  * Trait Serialization.
@@ -127,9 +128,14 @@ trait Serialization
      */
     public function __wakeup()
     {
-        if (get_parent_class() && method_exists(parent::class, '__wakeup')) {
+        // FatalError occurs when calling __wakeup method in PHP 7.4 or later.
+        // @codeCoverageIgnoreStart
+        if (version_compare(PHP_VERSION, '7.4.0-dev', '>=')) {
+            parent::__construct($this->date, $this->timezone !== null ? new DateTimeZone($this->timezone) : null);
+        } elseif (get_parent_class() && method_exists(parent::class, '__wakeup')) {
             parent::__wakeup();
         }
+        // @codeCoverageIgnoreEnd
 
         $this->constructedObjectId = spl_object_hash($this);
 


src/Carbon/Traits/Serialization.php
```diff
@@ -127,9 +128,14 @@ public function __sleep()
...
     {
-        if (get_parent_class() && method_exists(parent::class, '__wakeup')) {
+        // FatalError occurs when calling __wakeup method in PHP 7.4 or later.
+        // @codeCoverageIgnoreStart
+        if (version_compare(PHP_VERSION, '7.4.0-dev', '>=')) {
+            parent::__construct($this->{$this->dumpProperties[0]}, $this->timezone !== null ? new DateTimeZone($this->timezone) : null);
+        } elseif (get_parent_class() && method_exists(parent::class, '__wakeup')) {
             parent::__wakeup();
         }
+        // @codeCoverageIgnoreEnd
```
kylekatarnls, line 138: We have multiple little drawbacks with this.

`DateTime` has an explicitly [documented `__wakeup` method](https://www.php.net/manual/en/datetime.wakeup): so we should keep it to be future-proof and consistent with parent class.

The condition is too strict. [As per documentation](https://www.php.net/manual/en/language.oop5.magic.php),  `__sleep()` and `__wakeup()` are for `serialize()` and `unserialize()` and `unserialize(serialize(Carbon::parse('2018-06-01')))` works with no issue. Work-around should be at least scoped to `extension_loaded('msgpack')` and at best called only when calling `msgpack_unpack()` but not when calling `unserialize()`

Last (but that's not expected to be a standard usage, this will be a BC for usages of the trait like this:

```php
class Foo
{
    public $date = '2021-07-07 12:00:00';
    public $timezone_type = 3;
    public $timezone = 'America/Toronto';

    public function __construct()
    {
    }

    public function __wakeup()
    {
        echo "Hello";
    }
}

class Bar extends Foo
{
    use \Carbon\Traits\Serialization;
}
```
Output "Hello" before this PR, will output nothing after.
And worse if we don't have a `__construct` method:
Output "Hello" before this PR, will throw a fatal exception after.

kylekatarnls, line 138:
```suggestion
       if (get_parent_class() && method_exists(parent::class, '__wakeup')) {
            // @codeCoverageIgnoreStart
            try {
                parent::__wakeup();
            } catch (Throwable $exception) {
                // FatalError occurs when calling msgpack_unpack() in PHP 7.4 or later.
                parent::__construct(
                    $this->date,
                    $this->timezone !== null ? new DateTimeZone($this->timezone) : null
                );
            }
            // @codeCoverageIgnoreEnd
        }
```

src/Carbon/Traits/Serialization.php
```diff
@@ -127,9 +128,14 @@ public function __sleep()
      */
     public function __wakeup()
     {
-        if (get_parent_class() && method_exists(parent::class, '__wakeup')) {
+        // FatalError occurs when calling __wakeup method in PHP 7.4 or later.
+        // @codeCoverageIgnoreStart
+        if (version_compare(PHP_VERSION, '7.4.0-dev', '>=')) {
+            parent::__construct($this->{$this->dumpProperties[0]}, $this->timezone !== null ? new DateTimeZone($this->timezone) : null);
```
kylekatarnls, line 134: As per my tests when running your code, at this point `$this->date` is null:
```php
var_dump($this->{$this->dumpProperties[0]}, $this->date); // null null
```
And so we actually create a instance with the current datetime instead of getting a copy of the original object.

So unless I'm missing something, this does not fix #2299

PR #2314, opened 2021-04-13
Remove unnecessary check within `make` method

andrzejkupczyk:
This is just a minor readability improvement.  
We already know that `$var` is a string ([`trim`](https://www.php.net/manual/en/function.trim.php) always returns a string), so there's no need for additional check.


kylekatarnls (approved):
Thanks.

PR #2318, opened 2021-04-20
Update Vietnamese (vi.php) locale

sakuriki:
Capitalize first letter of months and weekdays


kylekatarnls:
According to https://linguapedia.info/en/phrases/days-of-the-week-in-vietnamese.html it's not capitalized.

Please provide a grammar reference for this change we capitalize months or weeks only if they are always capitalized in the given language even in the middle of a sentence such as in German or English but not else (Spannish, French, Italian, etc.).

sakuriki:
> only if they are always capitalized in the given language even in the middle of a sentence

Well, never mind, it just happened that I switch from English to Vietnamese and thought there are some typo, guess I use CSS now.

PR #2323, opened 2021-04-26
#2322 Fix regular expression to also match numbers in locale 

### LINKED ISSUE #2322: Invalid translation when using locale `es-419`  (works with es_419)
Hello,

```php
// Does Not work
 $date = \Carbon\Carbon::now();
 $date->locale('es-419');
 
 echo( $date->monthName );  -> April 👎 
 
 // Work 
 $date = \Carbon\Carbon::now();
 $date->locale('es_419');
 
 echo( $date->monthName ); -> abril   👍 
```

Carbon version: 2.46.0

PHP version: 7.4

I expected to get:

```
abril  (the spanish version)
```

But I actually get:

```
Avril (english version - the default)
```

Thanks!

I can't guess one of the question:
Why dont you use es_419 then ?
because I am trying to adhere to the BCP 47 format  https://www.ietf.org/rfc/bcp/bcp47.txt
which use `-` instead of `_`

Also it works for other locales when using '-'  like `de-de` for example so for consistency sake I think this bug should be fixed

cgaube:
Fix #2322

### DIFF
--- a/src/Carbon/Translator.php
+++ b/src/Carbon/Translator.php
@@ -313,7 +313,7 @@ class Translator extends Translation\Translator
      */
     public function setLocale($locale)
     {
-        $locale = preg_replace_callback('/[-_]([a-z]{2,})/', function ($matches) {
+        $locale = preg_replace_callback('/[-_]([a-z0-9]{2,})/', function ($matches) {
             // _2-letters or YUE is a region, _3+-letters is a variant
             $upper = strtoupper($matches[1]);
 
--- a/src/Carbon/Translator.php
+++ b/src/Carbon/Translator.php
@@ -313,7 +313,7 @@ class Translator extends Translation\Translator
      */
     public function setLocale($locale)
     {
-        $locale = preg_replace_callback('/[-_]([a-z0-9]{2,})/', function ($matches) {
+        $locale = preg_replace_callback('/[-_]([a-z]{2,}|[0-9]{2,})/', function ($matches) {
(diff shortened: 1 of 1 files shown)


src/Carbon/Translator.php
```diff
@@ -313,7 +313,7 @@ public function getMessages($locale = null)
      */
     public function setLocale($locale)
     {
-        $locale = preg_replace_callback('/[-_]([a-z]{2,})/', function ($matches) {
+        $locale = preg_replace_callback('/[-_]([a-z0-9]{2,})/', function ($matches) {
```
kylekatarnls, line 316: I would like to be a bit more precise on this, e.g. only letter or only digits to be closer to the norm:
```suggestion
        $locale = preg_replace_callback('/[-_]([a-z]{2,}|[0-9]{2,})/', function ($matches) {
```

kylekatarnls (changes requested):
Thanks for your help. It makes sense to normalize numeric locale as we do for regions.

Can you give your thought on the suggestion below?

src/Carbon/Translator.php
cgaube: This is actually how i started doing things then decided to go the other way `[a-z0-9]`   but i am happy with your suggestion

kylekatarnls (approved):

cgaube:
@kylekatarnls  Merci 
Looks like i don't not have permissions to merge this - I assume this is something that somebody else has to do ?

Issue #2329, opened 2021-05-17
Strict instantiation method with time


spawnia:
I am looking for a strict way to instantiate a Carbon instance for a statically known date. My current code looks like this:

```php
Carbon::create(1997, 12, 8, 8, 56, 20);
```

We have recently increased the level of strictness in our static validation tool. Using the result of the above call now gives me an error, because apparently `Carbon::create()` can return `false`: https://github.com/briannesbitt/Carbon/blob/d3c447f21072766cddec3522f9468a5849a76147/src/Carbon/Traits/Creator.php#L365-L367

I have switched to `Carbon::createFromDate()` in cases where the time is not relevant, but have not found an equivalent method that allows me instantiate a date with time. None of the other methods for instantiation do exactly what I want. I guess I could use the constructor, but that would invite possible format errors.

Could we add a strict specialization of `create`? Perhaps like this:

```php
    /**
     * @throws InvalidFormatException
     */
    public static function createStrict(int $year = null, int $month = null, int $day = null, int $hour = null, int $minute = null, int $second = null, $tz = null): static
    {
        return static::create($year, $month, $day, $hour, $minute, $second, $tz);
    }
```

kylekatarnls:
Hello 👋 Can you check if solutions provided in #2135 (useStrictMode(), macro, or `@var` annotation) can help?

spawnia:
Hey, it has been a while since we last talked 👋

Thanks for the swift response. Great discussion in that issue! I attempted to look for existing issues but apparently missed that one.

`useStrictMode()` definitely aligns with my preferences in regards to error handling (it looks like `settings()` should be used instead now). I will definitely look into enabling that in our project. However, while it does change the behaviour, it has no impact on static analysis. Perhaps a fancy PHPStan extension might solve that?

The `@var` type assertion is certainly a possible workaround and one that I considered as well. I view it as kind of a last resort though, as it can be quite harmful when refactoring. We tend to use them sparingly, and have a rule that they must be accompanied by a descriptive comment that justifies its use. Given we have hundreds of calls to `Carbon::create()` throughout our codebase, I am not too keen on converting them all to this.

I do like the macro solution. It seems that we have independently come up with `createStrict()`, that indicates to me that this method would be generically useful for many users. Would you be open towards a PR that adds it?

kylekatarnls:
[...]

PR #2332, opened 2021-05-18
Add instantiation method Carbon::createStrict()

spawnia:
Resolves https://github.com/briannesbitt/Carbon/issues/2329

### DIFF
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -478,6 +478,37 @@ trait Creator
     }
 
     /**
+     * Create a new Carbon instance from a specific date and time using strict validation.
+     *
+     * @see create()
+     *
+     * @param int|null                 $year
+     * @param int|null                 $month
+     * @param int|null                 $day
+     * @param int|null                 $hour
+     * @param int|null                 $minute
+     * @param int|null                 $second
+     * @param DateTimeZone|string|null $tz
+     *
+     * @throws InvalidFormatException
+     *
+     * @return static
+     */
+    public static function createStrict(?int $year = 0, ?int $month = 1, ?int $day = 1, ?int $hour = 0, ?int $minute = 0, ?int $second = 0, $tz = null): self
+    {
+        $initialStrictMode = static::isStrictModeEnabled();
+        static::useStrictMode(true);
+
+        try {
+            $date = static::create($year, $month, $day, $hour, $minute, $second, $tz);
+        } finally {
+            static::useStrictMode($initialStrictMode);
+        }
+
+        return $date;
+    }
+
+    /**
      * Create a Carbon instance from just a date. The time portion is set to now.
      *
      * @param int|null                 $year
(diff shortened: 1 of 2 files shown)


spawnia:
Given the implementation is fine, I can definitely expand the test cases. What else should it test for?

kylekatarnls (approved):

src/Carbon/Traits/Creator.php
```diff
@@ -477,6 +477,37 @@ public static function createSafe($year = null, $month = null, $day = null, $hou
...
+     *
+     * @return static
+     */
+    public static function createStrict(?int $year = 0, ?int $month = 1, ?int $day = 1, ?int $hour = 0, ?int $minute = 0, ?int $second = 0, $tz = null): self
+    {
+        $initialStrictMode = static::isStrictModeEnabled();
+        static::useStrictMode(true);
+
+        try {
+            $date = static::create($year, $month, $day, $hour, $minute, $second, $tz);
```
spawnia, line 503: Is there a point to using `createSafe()` here?

kylekatarnls:
```php
Carbon::useStrictMode(true);

echo Carbon::create(2021, 2, 30); // 2021-03-02
echo Carbon::createSafe(2021, 2, 30); // throw InvalidDateException Error: day : 30 is not a valid value.
```
Basically `createSafe` disallow overflow while `create()` actually allow 0 to 99 for any unit from month to second.

While `createSafe` is nice to prevent falling in unexpected result (as it ensure you either get a datetime with the numbers you pass or an exception in any other case), it's actually a bit slower as it makes the extra-checking around the `DateTime` constructor which allow overflow and it prevents convenient tricks like adding a fixed interval right in the `create()`:
```php
Carbon::create($year, $month, $day + 2); // $year-$month-$day + 2 days
```

kylekatarnls:
For the record, the PHPStan error making "Tests / Style check - PHP 7.4" CI test fail is now fixed on master, you can pull the upstream master to get the CI green on this PR.

src/Carbon/Traits/Creator.php
spawnia: So - stick with the current code?

kylekatarnls: So much regrets, please help me going back in time preventing myself adding this method.

We need to stick with `create()`, here are the blockers:
- `create()` is can also be used as an alias of `parse()`, e.g.: `Carbon::create('2021-01-02', 'Europe/London')`, stupid, but at some point judged convenient. And this is not allowed in `createSafe()`
- `createSafe()` doesn't allow overflow (such as February 30th overflowing to March 2nd) but default value for `$day` is `null` and `create()` use current value when passing `null`, convenient for time but a big mistake for the day, as mean `Carbon::createSafe(2020, 2)` actually throw an exception if today is a 29th, 30th or 31th. Actually this mistake is there since the very first release.

So rework needed for next major and I think `createSafe()` shouldn't be used internally by other methods meanwhile.

PR #2340, opened 2021-05-26
Return null instead of false

kylekatarnls:
Use stronger typing and `null` instead of `false`


spawnia:
I like where this is going, `null` is definitely more convenient then `false`, but why shouldn't methods such as `create()` always return a `Carbon` instance (or throw in case of invalid inputs)?

kylekatarnls:
This is the planned progression: strict mode becomes `true` by default from 3.0.0 and disappears (to be always `true`) in 4.0.0, this means most of the method won't return `null` any longer (except `make()` or such which are very dedicated to that purpose).

Actually with the default strict mode already in 3.0.0 you can't get `null` from `create()`

spawnia:
Got it, that makes a lot of sense to me. Good to see this library turning away from PHP's error swallowing legacy swamp.

kylekatarnls:
This is hoping it will also evolve on PHP side because providing methods like `createFromFormat(): Carbon|null` while `DateTime` has `createFromFormat(): DateTime|false` is actually violating the Liskov Substitution Principle making `Carbon` a bit less compatible with its parent. :(

So it will still need careful checking with dependent packages to validate we can keep it for the future stable release but I have good hope.

PR #2344, opened 2021-05-30
Add Doctrine Types Annotations

AlexandruGG:
## Description

This PR's purpose is to allow [phpstan-doctrine](https://github.com/phpstan/phpstan-doctrine) to recognise `Carbon`|`CarbonImmutable` and `carbon`|`carbon_immutable` usages in entities. The main beneficiaries of this are those using the library in a Symfony project with Doctrine.

## Features / Changes
- [x] adding type hints using generics to the needed types in `src/Carbon/Doctrine`
- [x] ignoring [these files](https://github.com/briannesbitt/Carbon/compare/master...AlexandruGG:feature/doctrine-types-annotations?expand=1#diff-644bd23a8e99e566f73e70a02a0d1a387c088927c32ef3cb75f3367e72f4334bR55-R56) in PHP-CS-FIXER - this is because the rule [phpdoc_to_comment](https://cs.symfony.com/doc/rules/phpdoc/phpdoc_to_comment.html) currently breaks the `@use` annotation for traits. It's been reported in various forms in multiple issues, one example [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4446#issuecomment-561703207). Sadly there is no way to ignore a single line with this tool, only an entire file; should not be a big issue though, as those files are very small 

## Testing
- [x] Manually tested that phpstan-doctrine works as expected in a separate project with these changes
- [x] Ran PHPStan in this project at max level to ensure the typings added are correct

## Notes
- this was recommended as the best approach by @ondrejmirtes here: https://github.com/phpstan/phpstan-doctrine/pull/192#issuecomment-850258319
- I tested and this will work with both ways of using this library with Doctrine if all types are registered in phpstan-doctrine, i.e:

A)
```
datetime_immutable: \Carbon\Doctrine\DateTimeImmutableType
datetime: \Carbon\Doctrine\DateTimeType
```

B)
```
carbon_immutable: \Carbon\Doctrine\CarbonImmutableType
carbon: \Carbon\Doctrine\CarbonType
```

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -49,6 +49,11 @@ $rules = array(
 );
 
 return Config::create()->setRules($rules)
-             ->setFinder(Finder::create()->in(__DIR__.'/src'))
+             ->setFinder(
+                Finder::create()
+                    ->in(__DIR__.'/src')
+                    ->notPath('Carbon/Doctrine/DateTimeType.php')
+                    ->notPath('Carbon/Doctrine/DateTimeImmutableType.php')
+             )
              ->setUsingCache(true)
              ->setRiskyAllowed(true);
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -36,3 +36,8 @@ enabled:
   - unneeded_control_parentheses
   - unused_use
   - whitespacy_lines
+
+finder:
+  exclude:
+    - "src/Carbon/Doctrine/DateTimeType.php"
+    - "src/Carbon/Doctrine/DateTimeImmutableType.php"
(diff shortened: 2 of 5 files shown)


kylekatarnls (approved):
LGTM Thanks 🙏

AlexandruGG:
hey @kylekatarnls, thanks for merging this. Since this is a small change do you think this could be released as part of a minor version `2.48.2`?  I'm just weary that it'll take a bit until the `2.49` [milestone](https://github.com/briannesbitt/Carbon/milestone/82) is ready.

kylekatarnls:
PHP 8.1 fixes postponed, 2.49.0 is out.

PR #2349, opened 2021-06-09
Fix polish translation causing a runtime error

maszczyn:
It seems that PL translations provide 3 variants, depending on `:count` param value, but the "short year" translation provides only two, which causes a runtime error.

I don't know PHP, so I am not able to provide (in reasonable time) reproduction steps other than the actual use where I spotted the problem. Sample manifestation of the bug is described in the following issue:
https://github.com/firefly-iii/firefly-iii/issues/4845


kylekatarnls (approved):
Thanks 🙏

kylekatarnls:
Indeed, no translation should have only 2 plural form in PL files.

PR #2354, opened 2021-06-14
Prefer https secure scheme.

lucasmichot:
Prefer `https` secure scheme over `http` scheme.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "time",
         "DateTime"
     ],
-    "homepage": "http://carbon.nesbot.com",
+    "homepage": "https://carbon.nesbot.com",
     "support": {
         "issues": "https://github.com/briannesbitt/Carbon/issues",
         "source": "https://github.com/briannesbitt/Carbon"
@@ -17,11 +17,11 @@
         {
             "name": "Brian Nesbitt",
             "email": "brian@nesbot.com",
-            "homepage": "http://nesbot.com"
+            "homepage": "https://nesbot.com"
         },
         {
             "name": "kylekatarnls",
-            "homepage": "http://github.com/kylekatarnls"
+            "homepage": "https://github.com/kylekatarnls"
         }
     ],
     "prefer-stable": true,
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         {
             "name": "Brian Nesbitt",
             "email": "brian@nesbot.com",
-            "homepage": "https://nesbot.com"
+            "homepage": "https://markido.com"
         },
         {
             "name": "kylekatarnls",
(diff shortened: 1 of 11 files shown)


src/Carbon/CarbonInterface.php
```diff
@@ -4818,7 +4818,7 @@ public function translateTimeStringTo($timeString, $to = null);
     public static function translateWith(\Symfony\Component\Translation\TranslatorInterface $translator, string $key, array $parameters = [], $number = null): string;
 
     /**
-     * Format as ->format() do (using date replacements patterns from http://php.net/manual/fr/function.date.php)
+     * Format as ->format() do (using date replacements patterns from https://php.net/manual/en/function.date.php)
```
lucasmichot, line 4821: An opportunity to change the default documentation locale too.

src/Carbon/Lang/bn_BD.php
```diff
@@ -9,10 +9,6 @@
  * file that was distributed with this source code.
  */
 
-/*
- * Authors:
- * - Ankur Group, http://www.ankurbangla.org, http://www.bengalinux.org Taneem Ahmed, Jamil Ahmed taneem@bengalinux.org, jamil@bengalinux.org
```
lucasmichot, line 14: Both those URLs return 404s.

src/Carbon/Lang/shn_MM.php
```diff
@@ -11,7 +11,7 @@
 
 /*
  * Authors:
- * - ubuntu Myanmar LoCo Team http://www.ubuntu-mm.net Bone Pyae Sone bone.burma@mail.com
+ * - ubuntu Myanmar LoCo Team https://ubuntu-mm.net Bone Pyae Sone bone.burma@mail.com
```
lucasmichot, line 14: `www.` redirects to non-www

src/Carbon/Lang/bn_BD.php
mfn: But just because the URLs don't work, there's no reason to remove the groups/names?

lucasmichot: Re-added

composer.json
```diff
@@ -17,11 +17,11 @@
         {
             "name": "Brian Nesbitt",
             "email": "brian@nesbot.com",
-            "homepage": "http://nesbot.com"
+            "homepage": "https://nesbot.com"
```
kylekatarnls, line 20: This website does not have a valid HTTPS certificate (it maybe never did).

Hello @briannesbitt, please check if you want to update your website link and give HTTPS link if you have. Thank you.

lucasmichot: Makes sense @kylekatarnls

kylekatarnls (approved):

PR #2356, opened 2021-06-14
Use proper tests assertions

lucasmichot:
Always use the proper tests assertions

### DIFF
--- a/tests/CarbonInterval/ToDateIntervalTest.php
+++ b/tests/CarbonInterval/ToDateIntervalTest.php
@@ -15,7 +15,7 @@ class ToDateIntervalTest extends AbstractTestCase
     {
         $interval = CarbonInterval::days(5)->hours(3)->minutes(50)->microseconds(123456)->invert()->toDateInterval();
 
-        $this->assertSame('DateInterval', get_class($interval));
+        $this->assertInstanceOf(DateInterval::class, $interval);
 
         $this->assertSame(1, $interval->invert);
         $this->assertSame(5, $interval->d);
@@ -29,7 +29,7 @@ class ToDateIntervalTest extends AbstractTestCase
     {
         $interval = CarbonInterval::days(5)->hours(3)->minutes(50)->microseconds(123456)->invert()->cast(DateInterval::class);
 
-        $this->assertSame('DateInterval', get_class($interval));
+        $this->assertInstanceOf(DateInterval::class, $interval);
 
         $this->assertSame(1, $interval->invert);
         $this->assertSame(5, $interval->d);
--- a/tests/CarbonInterval/ToDateIntervalTest.php
+++ b/tests/CarbonInterval/ToDateIntervalTest.php
@@ -15,7 +15,7 @@ class ToDateIntervalTest extends AbstractTestCase
     {
         $interval = CarbonInterval::days(5)->hours(3)->minutes(50)->microseconds(123456)->invert()->toDateInterval();
 
-        $this->assertInstanceOf(DateInterval::class, $interval);
+        $this->assertSame(DateInterval::class, get_class($interval));
 
         $this->assertSame(1, $interval->invert);
         $this->assertSame(5, $interval->d);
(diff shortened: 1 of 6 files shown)


tests/CarbonInterval/ToDateIntervalTest.php
```diff
@@ -15,7 +15,7 @@ public function testConvertToDateInterval()
     {
         $interval = CarbonInterval::days(5)->hours(3)->minutes(50)->microseconds(123456)->invert()->toDateInterval();
 
-        $this->assertSame('DateInterval', get_class($interval));
+        $this->assertInstanceOf(DateInterval::class, $interval);
```
kylekatarnls, line 18: Please revert this and the similar, it's not heuristic.

```php
'DateInterval' === get_class(new CarbonInterval()) // false
new CarbonInterval() instanceof DateInterval // true
```

kylekatarnls (changes requested):
Please revert non-heuristic chnges

count/array/equals checks are fine.

tests/CarbonInterval/ToDateIntervalTest.php
lucasmichot: Thanks for the look  @kylekatarnls 
I feel the previous assertion was actually a bit misleading and would love to enforce a bit more meaning
What do you think of `$this->assertTrue(is_a($interval, DateInterval::class));` ?

kylekatarnls: `is_a` has the same problem, it also makes the check pointless, where checking that we got a raw `DateInterval` and we don't want any kind of sub-class of `DateInterval`.

It's not a matter of code style, this test must succeed with a `DateInterval` instance and fail with a `CarbonInterval` or other classes that may extends `DateInterval`.

lucasmichot: > `is_a` has the same problem

I don't think so
```php
$this->assertInstanceOf(DateInterval::class, $interval);
$this->assertTrue(is_a($interval, DateInterval::class));
$this->assertFalse(is_a($interval, CarbonInterval::class));
```

lucasmichot: Nevertheless, I do not really care

lucasmichot: See c99a2c1bd8355c41f875abaeac0b5b32925241df

kylekatarnls: Again, `is_a(new CarbonInterval(), DateInterval::class)` returns true, which is wrong regarding the meaning of the test.

tests/CarbonPeriod/ToDatePeriodTest.php
```diff
@@ -22,7 +23,7 @@ public function testToArrayIsNotEmptyArray()
         $result = $period->toDatePeriod();
 
         $this->assertFalse($period->isEndExcluded());
-        $this->assertSame('DatePeriod', get_class($result));
+        $this->assertInstanceOf(DatePeriod::class, $result);
```
kylekatarnls, line 26: This test is still broken. `assertInstanceOf` is not what this test aims to verify.

tests/CarbonPeriod/ToDatePeriodTest.php
```diff
@@ -35,7 +36,7 @@ public function testToArrayIsNotEmptyArray()
         $newInstance = CarbonPeriod::instance($result);
 
         $this->assertTrue($period->isEndExcluded());
-        $this->assertSame('DatePeriod', get_class($result));
+        $this->assertInstanceOf(DatePeriod::class, $result);
```
kylekatarnls, line 39:
```suggestion
        $this->assertSame(DatePeriod::class, get_class($result));
```

tests/CarbonPeriod/ToDatePeriodTest.php
```diff
@@ -47,7 +48,7 @@ public function testToArrayIsNotEmptyArray()
         $result = $period->toDatePeriod();
         $newInstance = CarbonPeriod::instance($result);
 
-        $this->assertSame('DatePeriod', get_class($result));
+        $this->assertInstanceOf(DatePeriod::class, $result);
```
kylekatarnls, line 51:
```suggestion
        $this->assertSame(DatePeriod::class, get_class($result));
```

PR #2357, opened 2021-06-14
Directly extend AbstractTestCaseWithOldNow

lucasmichot:
Directly extend `AbstractTestCaseWithOldNow`


tests/AbstractTestCaseWithOldNow.php
```diff
@@ -29,10 +29,9 @@ abstract class AbstractTestCaseWithOldNow extends AbstractTestCase
     protected function tearDown(): void
     {
         Carbon::resetMacros();
-        Carbon::setTestNow();
```
lucasmichot, line 32: Both `*:: setTestNow();` can safely be removed as they are already call in `\Tests\AbstractTestCase::tearDown`

kylekatarnls (approved):
Thanks 🙏

PR #2360, opened 2021-06-14
Always use stricter string assertions

lucasmichot:
Always use stricter string assertions when possible.


kylekatarnls (approved):
Thanks 🙏

PR #2337, opened 2021-05-25
Fix passing null to non-nullable arguments

jrmajor:
Passing null to non-nullable arguments of internal functions has been deprecated in PHP 8.1 ([RFC](https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg)).

PR #2364, opened 2021-06-26
Fixes for PHP 8.1

jrmajor:
Since my PR #2338 return types have also been added to `JsonSerializable::jsonSerialize()`.

I've also fixed a few cases of passing null to non-nullable arguments that I've missed in #2337.

### DIFF
--- a/src/Carbon/Lang/nl.php
+++ b/src/Carbon/Lang/nl.php
@@ -71,9 +71,9 @@ return [
     'period_recurrences' => ':count keer',
     'period_interval' => function ($interval) {
         /** @var string $output */
-        $output = preg_replace('/^(een|één|1)\s+/', '', $interval);
+        $output = preg_replace('/^(een|één|1)\s+/', '', $interval ?? '');
 
-        if (preg_match('/^(een|één|1)( jaar|j| uur|u)/', $interval)) {
+        if (preg_match('/^(een|één|1)( jaar|j| uur|u)/', $interval ?? '')) {
             return "elk $output";
         }
 
--- a/src/Carbon/Lang/nl.php
+++ b/src/Carbon/Lang/nl.php
@@ -69,11 +69,11 @@ return [
     'diff_after_tomorrow' => 'overmorgen',
     'diff_before_yesterday' => 'eergisteren',
     'period_recurrences' => ':count keer',
-    'period_interval' => function ($interval) {
+    'period_interval' => function (string $interval) {
         /** @var string $output */
-        $output = preg_replace('/^(een|één|1)\s+/', '', $interval ?? '');
+        $output = preg_replace('/^(een|één|1)\s+/', '', $interval);
 
-        if (preg_match('/^(een|één|1)( jaar|j| uur|u)/', $interval ?? '')) {
+        if (preg_match('/^(een|één|1)( jaar|j| uur|u)/', $interval)) {
             return "elk $output";
         }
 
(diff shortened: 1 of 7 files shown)


src/Carbon/Lang/oc.php
```diff
@@ -86,7 +86,7 @@
         $ordinal = [1 => 'èr', 2 => 'nd'][(int) $number] ?? 'en';
 
         // feminine for year, week, hour, minute, second
-        if (preg_match('/^[yYwWhHgGis]$/', $period)) {
+        if (preg_match('/^[yYwWhHgGis]$/', $period ?? '')) {
```
kylekatarnls, line 89: I think we could just skip the `preg_match` call if `$period` is falsy.
```suggestion
        if ($period && preg_match('/^[yYwWhHgGis]$/', $period)) {
```

src/Carbon/Lang/nl.php
```diff
@@ -71,9 +71,9 @@
     'period_recurrences' => ':count keer',
     'period_interval' => function ($interval) {
```
kylekatarnls, line 72: Actually we could enforce `string` here. In the Carbon context, it will always receive a string:
```suggestion
    'period_interval' => function (string $interval) {
```

jrmajor: `TypeError : Carbon\Translator::{closure}(): Argument #1 ($interval) must be of type string, null given, called in src/Carbon/Translator.php on line 168`

The tests pass when you replace `null` with `''` in this code: https://github.com/briannesbitt/Carbon/blob/27a1ff3b2daf97a1c9d033fe0b1233abbbb8930a/src/Carbon/Translator.php#L166-L169

But it might be a breaking change for people using this translator (or is it only used internally?)

src/Carbon/Lang/oc.php
jrmajor: I've added type hint to this closure, is it OK like that, or do you want this check to be there anyway?

src/Carbon/Lang/nl.php
kylekatarnls: The only point where we use `period_interval` translation is:
```php
$this->translate('period_interval', [':interval' => $this->dateInterval->forHumans([
    'join' => true,
])], null, $translator)
```
So it's only breaking for people using:
```php
->translate('period_interval', [':interval' => null])
```
But indeed it could be smoother to just cast it. Then I guess we could cast only once at the beginning.

jrmajor: `CarbonInterval::forHumans()` always returns a string. Please see my latest commit (0a396b5), the problem was caused by the test that checks whether translation key exists and passes no arguments. No need to cast, I've fixed the test.

src/Carbon/Traits/Localization.php
```diff
@@ -643,7 +643,7 @@ public static function localeHasPeriodSyntax($locale)
         return static::executeWithLocale($locale, function ($newLocale, TranslatorInterface $translator) {
             return $newLocale &&
                 $translator->trans('period_recurrences') !== 'period_recurrences' &&
-                $translator->trans('period_interval') !== 'period_interval' &&
+                $translator->trans('period_interval', ['']) !== 'period_interval' &&
```
kylekatarnls, line 646: Is it needed? `':period' => (string) $period,` should convert `null` into `""` already.

src/Carbon/Lang/oc.php
kylekatarnls: No it's fine like this, `':period' => (string) $period,` to ensure the string type for text interpolations make complete sense.

src/Carbon/Traits/Localization.php
kylekatarnls: Checked, it is. I can't see any relevant use case for using translator directly and translate such `period_interval` without parameter, but I will double-check.

kylekatarnls:
Thank you 🙏

PR #2370, opened 2021-06-29
Enforce consistency and stop using is_null.

lucasmichot:
`null` comparisons are checked with `===` and `!==` everywhere in the code, which is faster than using `is_null` function.

Only those `is_null` calls were remaining.
 
 This PR enforces consistency and remove them.


kylekatarnls (approved):
🙏

PR #2372, opened 2021-06-29
Improve readability by using strpos.

lucasmichot:
Improve readability by using `strpos` for comparing the start of a string.

### DIFF
--- a/phpdoc.php
+++ b/phpdoc.php
@@ -442,7 +442,7 @@ foreach ($tags as $tag) {
 
         $description = trim($vars->description);
         $variable = $vars->name;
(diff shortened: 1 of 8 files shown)


kylekatarnls:
Thanks. I plan to use `str_starts_with()` in 4.x version (PHP >= 8), but for now I find the benefit of `strpos()` quite subjective and it's actually slower in most cases (as it reads the whole string when not found).

kylekatarnls:
Good news about this, finally dropped PHP 7.4 for Carbon 3, so even in 3.x we'll be able to switch to `str_starts_with`

PR #2378, opened 2021-06-30
Add getTimestampMs method to output millisecond-precise timestamps

aryehraber:
Hi there! Firstly, thank you for your continued work on Carbon, I don't know what I'd do without it! 

This PR adds a new `getTimestampMs()` method to compliment the already existing `createFromTimestampMs()` method. 

This is mostly a helper method to easily return a millisecond-precise timestamp as an `integer`. It also reads more naturally than the currently available methods (which return a `float`):

**Before**
```php
$timestamp = (int) $carbonDate->getPreciseTimestamp(3);

// OR

$timestamp = (int) $carbonDate->valueOf();
```

**After**
```php
$timestamp = $carbonDate->getTimestampMs();
```

Please let me know if there's any questions or feedback. I'm also happy to submit a PR to the docs if this gets merged.

Thanks!

### DIFF
--- a/src/Carbon/Traits/Timestamp.php
+++ b/src/Carbon/Traits/Timestamp.php
@@ -138,6 +138,16 @@ trait Timestamp
     }
 
     /**
+     * Returns the timestamp with millisecond precision.
+     *
+     * @return int
+     */
+    public function getTimestampMs()
+    {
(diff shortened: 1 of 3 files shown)


kylekatarnls (approved):

kylekatarnls:
Thanks. 👍

PR #2388, opened 2021-07-09
Stop displaying coverage report for each PR

lucasmichot:
This PR disables the coverage report comment report for each PR.
IMO having access to hit via the PR status and the README badge is already sufficient.


kylekatarnls (approved):
LGTM

PR #2392, opened 2021-07-09
Optimize conditions

lucasmichot:
This PR optimises conditions.
For instance a simple value comparison should always be done before calling a ( potential costly ) function

### DIFF
--- a/src/Carbon/CarbonPeriod.php
+++ b/src/Carbon/CarbonPeriod.php
@@ -651,10 +651,12 @@ class CarbonPeriod implements Iterator, Countable, JsonSerializable
                 $parsed = @CarbonInterval::make($argument)
             ) {
                 $this->setDateInterval($parsed);
-            } elseif ($this->startDate === null && $parsed = Carbon::make($argument)) {
-                $this->setStartDate($parsed);
-            } elseif ($this->endDate === null && $parsed = Carbon::make($argument)) {
-                $this->setEndDate($parsed);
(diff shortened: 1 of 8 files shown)


src/Carbon/Translator.php
```diff
@@ -361,7 +361,9 @@ public function setLocale($locale)
             parent::setLocale($macroLocale);
         }
 
-        if ($this->loadMessagesFromFile($locale) || $this->initializing) {
+        $loaded = $this->loadMessagesFromFile($locale);
```
lucasmichot, line 364: This code has been changed on purpose: `loadMessagesFromFile` method should always be called.

src/Carbon/CarbonPeriod.php
```diff
@@ -651,10 +651,12 @@ public function __construct(...$arguments)
...
-            } elseif ($this->startDate === null && $parsed = Carbon::make($argument)) {
-                $this->setStartDate($parsed);
-            } elseif ($this->endDate === null && $parsed = Carbon::make($argument)) {
-                $this->setEndDate($parsed);
+            } elseif ($parsed = Carbon::make($argument)) {
+                if ($this->startDate === null) {
+                    $this->setStartDate($parsed);
+                } elseif ($this->endDate === null) {
+                    $this->setEndDate($parsed);
+                }
```
kylekatarnls, line 659: This makes the behavior change, that's why the unit test fails. I'm closing this but I would re-open it if you can revert this part and rebase.

PR #2404, opened 2021-07-29
Improve exceptions testing

lucasmichot:
Add one test class for each exception, where the message and the getters are all tested.
By adding those tests, the coverage surface of the testsuite should improve.

See: `namespace Tests\Carbon\Exceptions;`

---

Also this PR enforces testing expected exception, with `expectExceptionObject` as what should be tested is not the expected type of exception and its message, but with which argument a specific exception type as been called.

### DIFF
--- a/tests/Carbon/CreateSafeTest.php
+++ b/tests/Carbon/CreateSafeTest.php
@@ -26,10 +26,7 @@ class CreateSafeTest extends AbstractTestCase
 
     public function testCreateSafeThrowsExceptionForSecondLowerThanZero()
     {
-        $this->expectException(InvalidDateException::class);
-        $this->expectExceptionMessage(
-            'second : -1 is not a valid value.'
-        );
+        $this->expectExceptionObject(new InvalidDateException('second', -1));
 
         Carbon::createSafe(null, null, null, null, null, -1);
     }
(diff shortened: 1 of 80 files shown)


tests/Carbon/CreateSafeTest.php
```diff
@@ -26,10 +26,7 @@ public function testInvalidDateExceptionProperties()
 
     public function testCreateSafeThrowsExceptionForSecondLowerThanZero()
     {
-        $this->expectException(InvalidDateException::class);
-        $this->expectExceptionMessage(
-            'second : -1 is not a valid value.'
-        );
+        $this->expectExceptionObject(new InvalidDateException('second', -1));
```
kylekatarnls, line 29: Sounds to me like we testing here `2 + 2 = 2 + 2` instead of testing `2 + 2 = 4` when doing this we mirror a little more the source code and we'll no longer detect if the `InvalidDateException` code/message broke/changed. I have to double-think of that.

lucasmichot: You will because, because there is a custom `\Tests\Carbon\Exceptions\InvalidDateExceptionTest::testInvalidCastException` will fail.

I like to think this is a command bus testing approach.
- we test that the command is called with the expected argument (how is the exception constructed?)
- we ensure that the handler does the expected thing (what will be the built message of the exception depending on its arguments)

kylekatarnls: Nice, indeed it's a good approach.

kylekatarnls (approved):

PR #2405, opened 2021-07-29
Typehint all functions in tests

lucasmichot:
This PR typehints all functions in `/tests`


kylekatarnls:
This will be a pain to merge in 3.x, could we do this instead on 3.x branch and let it go for 2.x versions?

lucasmichot:
> This will be a pain to merge in 3.x, could we do this instead on 3.x branch and let it go for 2.x versions?

Absolutely

Issue #2399, opened 2021-07-28
str_starts_with() 


aryala7:
Hello,
i encountered an issue after updating to the latest version

Carbon version: **2.51.0**

PHP version: **7.4**

![carbon_bug](https://user-images.githubusercontent.com/56627259/127313640-ab796b48-3475-4378-b2dc-5a37f7df4ec2.png)Thanks!

kylekatarnls:
Hello :wave

I need the complete stack trace to solve this and the code you run that called it. Thanks.

drewwynne0:
str_starts_with is a PHP8 only function.  With this release, Carbon has effectively just dropped support for PHP7.x

https://www.php.net/manual/en/function.str-starts-with.php

kylekatarnls:
😄 Hopefully not [symfony/polyfill-php80](https://packagist.org/packages/symfony/polyfill-php80)

trevorgehman:
@drewwynne0 This package uses the php8 polyfill by Symfony, so it will work fine on PHP 7. However, I just ran into this issue as well after updating my dependencies:

<img width="885" alt="CleanShot 2021-07-28 at 14 52 15@2x" src="https://user-images.githubusercontent.com/3209864/127387007-32fb178f-f2ea-4669-820d-cf76e6a1610c.png">

I believe the cause is this pull request: https://github.com/briannesbitt/Carbon/pull/2393

The issue is that `substr()` accepts `null` but `str_starts_with()` does not.

kylekatarnls:
Thanks @trevorgehman stack strace is a good start but I still miss the code called to reproduce the error, here you mention: a usage that could for instance be called with `Carbon::createFromTimestampMsUTC(null)` but it's not a valid usage as this methods expect `@param float|int|string $timestamp` (does not allow `null`) I will fix it if it's a regression for a valid case: a usage where having `null` is relevant, if not then it will be better to keep it strict and fix your code to not pass `null` in the method rather than allowing it because converting silently `null` to something else (such as `0` for `createFromTimestampMsUTC` will lead to unexpected and hard to debug behavior, so you better have a fatal error here IMO).

I can give more accurate feedback and determine if it needs to be fixed or not only if you can provide legit example of usage where `null` get passed in.

drewwynne0:
@kylekatarnls @trevorgehman fair point, though still somewhat of a frustrating change for a minor release.

trevorgehman:
[...]

PR #2408, opened 2021-07-29
Enforce strict types

lucasmichot:
Enforce strict types

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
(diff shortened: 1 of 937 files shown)


kylekatarnls:
First I'm completely favorable to use strict types.

I'm just a bit worried about surprises like in #2399 where users have usages outside what is expected as per our unit tests and PHPDoc. I wonder if it would be better to put this in 3.x branch for next major version.

Also, I don't think we should add those in the README. Code examples should be the very minimum code chunk the developer needs and I don't think it's relevant for the Carbon library to push user rather to use or not strict_types.

PR #2428, opened 2021-08-21
Add missing u regexp modifier

lucasmichot:
Add missing `u` regexp modifier

(as the regexp contains unicode characters)

### DIFF
--- a/src/Carbon/Lang/si.php
+++ b/src/Carbon/Lang/si.php
@@ -33,7 +33,7 @@ return [
     'a_second' => '{1}තත්පර කිහිපයකට|තත්පර :count',
     'ago' => ':time කට පෙර',
     'from_now' => function ($time) {
-        if (preg_match('/දින \d+/u', $time)) {
+        if (preg_match('/දින \d/u', $time)) {
             return $time.' න්';
         }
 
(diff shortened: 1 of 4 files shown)


src/Carbon/Lang/si.php
```diff
@@ -33,15 +33,15 @@
     'a_second' => '{1}තත්පර කිහිපයකට|තත්පර :count',
     'ago' => ':time කට පෙර',
     'from_now' => function ($time) {
-        if (preg_match('/දින \d+/', $time)) {
+        if (preg_match('/දින \d+/u', $time)) {
```
kylekatarnls, line 36: Those was converted as is from moment.js but I realize in our case we just want to detect the presence without using the captured number so we can even drop the `+` as we're optimizing it.
```suggestion
        if (preg_match('/දින \d/u', $time)) {
```

src/Carbon/Lang/si.php
```diff
@@ -33,15 +33,15 @@
...
             return $time.' න්';
         }
 
         return $time.' කින්';
     },
     'before' => ':time කට පෙර',
     'after' => function ($time) {
-        if (preg_match('/දින \d+/', $time)) {
+        if (preg_match('/දින \d+/u', $time)) {
```
kylekatarnls, line 44:
```suggestion
        if (preg_match('/දින \d/u', $time)) {
```

kylekatarnls (approved):

kylekatarnls:
For the record, Style Check failure sounds unrelated. More likely a new version of `php-cs-fixer` is now capable to detect errors that were missing earlier.

PR #2440, opened 2021-09-06
Split style check GitHub Actions scripts

kylekatarnls:


lucasmichot:
@kylekatarnls 
I think that StyleCI should be disabled,

it still runs for some PR 

image

PR #2445, opened 2021-09-10
Enforce tests consistency with some extra rules

lucasmichot:
This PR enforces tests consistency with some extra rules


kylekatarnls (approved):
👍

PR #2454, opened 2021-09-15
Use Generators for the tests dataProviders

lucasmichot:
Use Generators for the tests dataProviders


tests/Carbon/WeekTest.php
```diff
@@ -13,5305 +13,5312 @@
...
+        yield ['2020-01-10', [
             2020,
             2020,
             2,
             2,
             52,
             53,
-        ],
-    ];
+        ]];
```
kylekatarnls, line 5302: For this one, we will have one more tearDown() and setUp() call for each item, so I wonder if it won't actually slow down significantly the test due to the size of the sample.

lucasmichot: It actually does not, duration for php 8 stable are still around 35-45 seconds with GA
So not much difference

lucasmichot: And memory wise it will always be more efficient

kylekatarnls: Not true, because PHPUnit immediately convert this Generator into an array. before it re-iterate on it to run tests.

PR #2456, opened 2021-09-15
Enforce header declaration

lucasmichot:
Enforce header declaration


kylekatarnls:
Thanks. Merged via #2461

PR #2457, opened 2021-09-15
Enforce PSR12 code-style

lucasmichot:
Enforce PSR12 code-style


kylekatarnls:
Thanks, merged via #2460

PR #2467, opened 2021-09-28
Conflicts with symfony/translation 5.3.8

poplach:


kylekatarnls:
Not wanted, I'm pretty confident that symfony/translation 5.3.9 will fix the issue, so we could mark 5.3.8 only as conflicting but it's actually not a conflict, this version is just broken.

So we should mark all the previous versions of Carbon (and many other packages) are incompatible with symfony/translation 5.3.8.

Obviously this is not possible as they are already tagged. So this PR is pointless as you can still get any of the nesbot/carbon versions that currently exist installed with symfony/translation 5.3.8 by composer.

PR #2486, opened 2021-11-03
Add random method.

lucasmichot:
This PR introduces a `random` method.

this method allows to create a random of Carbon or CarbonImmutable within a specific range.
None of the boundaries are mandatory. If not provided, the minimum and maximum timestamps will be used.

This method also accepts a timezone.


kylekatarnls:
Hello, interesting idea. I see few points we might consider.

I can find this handful for testing purpose, but don't see concret application for real life. If we are on a test-only feature, I wonder if it wouldn't be better to have it as an external until (macro/mixin advertised in the documentation) or maybe in  https://github.com/kylekatarnls/carbonite

With `createFromTimestamp()` the result always have `microseconds = 0` while one might expect it to be a random value between 0 and 999999 when calling `random()`.

lucasmichot:
> With createFromTimestamp() the result always have microseconds = 0 while one might expect it to be a random value between 0 and 999999 when calling random().

👍

PR #2492, opened 2021-11-04
Fix past declensions in Polish

### LINKED ISSUE #2480: [Translation tool] pl
## Additional fixes
Something tricky:

- `2 hours from now` is `za 2 godziny` - that's correct
- `1 hour` is `1 godzina` - that's also correct
- `1 hour from now` is `za 1 godzina` - that's incorrect, should be `za 1 godzinę`

also:

- `1 minute from now` should be `za 1 minutę` (not `za 1 minuta`)
- `1 second from now` should be `za 1 sekundę` (not `za 1 sekunda`)

It should be corrected only for difference value of `1` and applies to `hour`, `minute` and `second` if the checked time is in future.

kylekatarnls:
Fix #2480

### DIFF
--- a/src/Carbon/Lang/pl.php
+++ b/src/Carbon/Lang/pl.php
@@ -53,7 +53,21 @@ return [
     'a_second' => '{1}kilka sekund|:count sekunda|:count sekundy|:count sekund',
     's' => ':count sek.',
     'ago' => ':time temu',
-    'from_now' => 'za :time',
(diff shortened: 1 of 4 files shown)


maciek-szn:
Does it cover usage of `'aUnit' => true` (so `za godzinę` instead of `za 1 godzinę`)?

kylekatarnls:
Thanks, I added it.

PR #2496, opened 2021-11-09
Unify data providers method names

lucasmichot:
This PR ensure that all tests data providers are prefixed with `provider`.
This PR also enforces consistency and ensures that all all the methods are fully-qualified.

### DIFF
--- a/tests/Carbon/AddMonthsTest.php
+++ b/tests/Carbon/AddMonthsTest.php
@@ -33,7 +33,7 @@ class AddMonthsTest extends AbstractTestCase
         $this->carbon = $date;
     }
 
-    public function providerTestAddMonthNoOverflow(): Generator
+    public function dataForTestAddMonthNoOverflow(): Generator
     {
         yield [-2, 2015, 11, 30];
         yield [-1, 2015, 12, 31];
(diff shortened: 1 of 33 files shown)


lucasmichot:
> Another suggestion maybe for the future: make data provider method `static`. They're effectively this, because they are evaluated upfront before the test suite is ran.

Absolutely @mfn
Probably part of another PR

kylekatarnls:
I would be OK to harmonize but it's the first time I see "provider" as a prefix. It makes it grammatically wrong for instance for "providerLocales" I would prefer a verb such as "provideLocales" or a suffix "localesProvider" so we can get an English readable method name.

mfn:
I found this tweet https://twitter.com/rectorphp/status/1169171714510602240?lang=en which suggests

> provideDataFor

and, unclear from just the tweet, I assume the test name (I guess with creative freedom if a single dataProvider is used for multiple different tests).

In private projects I've seen just `public static function dataFor<NameOfTestMethod>`; sweet'n'short (the prefix).

HTH

lucasmichot:
I like `dataFor` a lot

kylekatarnls:
👍 OK for `dataFor`

kylekatarnls (approved):

PR #2498, opened 2021-11-09
Ensure dataProviders return Generators.

lucasmichot:
Ensure that `dataProviders` return `Generator`.


kylekatarnls:
Returning generators is not helpful if you're looking for memory saving since PHPUnit will convert it again into an array before running the tests, and this will also not be easy to merge this in 3.x branch.

Issue #2425, opened 2021-08-19
Date::formatLocalized vs PHP 8.1


drbyte:
PHP 8.1 deprecates `strftime()` per https://wiki.php.net/rfc/deprecations_php_8_1#strftime_and_gmstrftime

This affects all formats beginning with `%`, which triggers `Date::formatLocalized()`, which calls `strftime()`:

https://github.com/briannesbitt/Carbon/blob/369c0e2737c56a0f39c946dd261855255a6fccbe/src/Carbon/Traits/Date.php#L893

https://github.com/briannesbitt/Carbon/blob/369c0e2737c56a0f39c946dd261855255a6fccbe/src/Carbon/Traits/Date.php#L1819-L1829

kylekatarnls:
Hello,

For the same reason than `strftime()` should not be used (it is platform-dependent) then you should not use `formatLocalizzed()` and `isoFormat()` exists for that purpose:
https://carbon.nesbot.com/docs/#api-localization

Carbon embed all the translations you need.

From our side it's the preferred way for a long time and `formatLocalized()` is just legacy. `formatLocalized()` is purely mirroring `strftime` using the current Carbon instance date-time. Consider it deprecated in PHP 8.1 too.

About the getters, use the `isoFormat` instead too:
`->localeDayOfWeek` => `->isoFormat('dddd')`
`->shortLocaleDayOfWeek` => `->isoFormat('ddd')`
`->localeMonth` => `->isoFormat('MMMM')`
`->shortLocaleMonth` => `->isoFormat('MMM')`

I think, we'll change those getters to use `isoFormat()` but `formatLocalized()` itself does not worth to be fixed.

drbyte:
Thanks for the feedback.
I'll leave this issue open; feel free to close it when you've decided how you intend to address the points you mentioned.

Stanislas-Poisson:
Thanks. It save my day

rulatir:
Where can I find a 100% exhaustive, Rosetta-stone guide for migrating from `formatLocalized()` format strings to `isoFormat()` format strings?

kylekatarnls:
Formats in `formatLocalized` are the ones of `strftime` as documented in PHPDoc:
https://www.php.net/manual/en/function.strftime

Formats in `isoFormat()` are listed in the 2 tables here:
https://carbon.nesbot.com/docs/#iso-format-available-replacements

They are not 1:1 equivalent so the mapping between both depends on your business logic.

[...]

PR #2504, opened 2021-11-26
Fix strftime PHP 8.1 depreciations.

lucasmichot:
Fix `strftime` PHP 8.1 depreciations.


kylekatarnls:
`strftime` use the format with `%` and OS translations, while `date_format` use ISO en English only. So it's not 1:1 interchangeable.

For the record the issue is handled for 3.x #2425 and #2491. For 2.x I just believe we should propagate the deprecation.

PR #2511, opened 2021-12-03
Make AbstractTranslator class actually abstract

lucasmichot:
This PR provides some adjustment for `AbstractTranslator`


src/Carbon/AbstractTranslator.php
```diff
@@ -68,18 +70,18 @@ public static function get($locale = null)
         $locale = $locale ?: 'en';
 
         if (!isset(static::$singletons[$locale])) {
-            static::$singletons[$locale] = new static($locale ?: 'en');
```
lucasmichot, line 71: Already checked 2 lines above

kylekatarnls (approved):

PR #2528, opened 2022-01-03
CarbonPeriod Handle infinite recurrences when parsing period string

### LINKED ISSUE #2527: Handling infinite recurrences on parsing from ISO 8601 string
Hello,

```php
$period = CarbonPeriod::create('2018-05-31', 'P3M', INF);

echo CarbonPeriod::create($period->toIso8601String())->toIso8601String();
```

Carbon version: **2.55.2**

PHP version: **8.0.14**

I expected to get:

```
RINF/2018-05-31T00:00:00+00:00/P3M
```

But I actually get:

```
Error: Could not parse 'RINF': DateTime::__construct(): Failed to parse time string (RINF) at position 0 (R): The timezone could not be found in the database
```

I expected, that the ISO8601 serialization created by CarbonPeriod could be used to create a new CarbonPeriod object, that is similarly constructed. I assume, that INF as a recurrence parameter might be a deviation from the standard format. Is this behaviour intended, that these periods can be serialized to an ISO8601 string but not deserialized again?

Thanks!

dbohn:
Fixes #2527.
The regular expression to parse the recurrence is augmented with the INF value. Based on this, the recurrence value is set to INF, if RINF is found.

### DIFF
--- a/src/Carbon/CarbonPeriod.php
+++ b/src/Carbon/CarbonPeriod.php
@@ -482,8 +482,8 @@ class CarbonPeriod implements Iterator, Countable, JsonSerializable
         $end = null;
 
         foreach (explode('/', $iso) as $key => $part) {
-            if ($key === 0 && preg_match('/^R([0-9]*)$/', $part, $match)) {
-                $parsed = \strlen($match[1]) ? (int) $match[1] : null;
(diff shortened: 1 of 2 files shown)


kylekatarnls (approved):

kylekatarnls:
Thanks for your help 🙏 

This will land in 2.56.0 and can be tested for now with `composer require "nesbot/carbon:2@dev"`

PR #2534, opened 2022-01-09
Update nn.php

asmaa13295:
update weekdays translation in the Norwegian language


kylekatarnls:
Hello,

According to the examples we can find on glosbe.com `sundag` is the primary translation for "Sunday" in "nn" (nynorsk):
https://glosbe.com/en/nn/sunday

Maybe you're looking for "nb" (bokmål):
https://glosbe.com/en/nb/sunday

Which has `søndag` in its weekdays:
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/nb.php

asmaa13295:
@kylekatarnls thanks for this clarification, it's so helpful. 
we'll use the other one.

PR #2537, opened 2022-01-13
Allow precision 0 in Carbon DBAL types

kdaniel95:
Allow explicitly defined 0 precision for DBAL types without explicitly setting Datetime precision through `DateTimeDefaultPrecision::set()` because right now explicit 0 precision is treated same as `null`.

### DIFF
--- a/tests/Doctrine/CarbonTypesTest.php
+++ b/tests/Doctrine/CarbonTypesTest.php
@@ -64,9 +64,17 @@ class CarbonTypesTest extends AbstractTestCase
         $precision = DateTimeDefaultPrecision::get();
         $this->assertSame(6, $precision);
 
+        $this->assertSame('DATETIME', $type->getSQLDeclaration([
+            'precision' => 0,
+        ], new MySQL57Platform()));
+
         $this->assertSame('DATETIME(3)', $type->getSQLDeclaration([
             'precision' => 3,
         ], new MySQL57Platform()));
+
+        $this->assertSame('TIMESTAMP(0)', $type->getSQLDeclaration([
+            'precision' => 0,
+        ], new DB2Platform()));
 
         $this->assertSame('TIMESTAMP(6)', $type->getSQLDeclaration([
             'precision' => null,
--- a/tests/Doctrine/CarbonTypesTest.php
+++ b/tests/Doctrine/CarbonTypesTest.php
@@ -65,7 +65,8 @@ class CarbonTypesTest extends AbstractTestCase
         $this->assertSame(6, $precision);
 
         $this->assertSame('DATETIME', $type->getSQLDeclaration([
-            'precision' => 0,
+            'precision' => null,
+            'secondPrecision' => true,
         ], new MySQL57Platform()));
 
         $this->assertSame('DATETIME(3)', $type->getSQLDeclaration([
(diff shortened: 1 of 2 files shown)


TamasSzigeti (approved):

kylekatarnls:
Sadly, `@ORM\Column()` is assigning `0` as property default value:
https://github.com/doctrine/orm/blob/2.11.x/lib/Doctrine/ORM/Mapping/Column.php#L32

which means this is a breaking change.

If an application is currently doing:
```php
/**
 * @ORM\Column(type="datetime")
 */
```
without specifying any precision value, they are currently interpreted as `precision=10` so it uses the maximum precision the DB engine can provide. If we merge this MR as is, those will suddenly become `0`, it will lead to all second fractional data to be lost and this precision is often critical (that's why initially we provided only `precision=6` with no option to change it, then in a next iteration, we allowed to customize it so it can be increased but without breaking loosing precision for existing apps).

I will check the other possibilities. But I would generally advocate for keeping maximum precision for datetimes. You won't regret having precision you don't use, on the opposite side, settle for second-precision at a given moment means the data you're saving now will never have second fraction, and if you need it later, it will be too late for the existing data. Also it's one less thing to bother if have all your datetime/timestamp with constant precision in your DB and don't have to determine which ones will need more or less.

TamasSzigeti:
@kylekatarnls What about an extra argument to explicitly set 0 then? That shouldn't hurt BC

kylekatarnls:
Indeed, some secondPrecision as a boolean could be a backward compatible solution.

kdaniel95:
So in this case, is it enough that I check if there's a key `secondPrecision` in `$fieldDeclaration` (and if its also true), then I set `$precision` to 0?

So something like this:

`$precision = isset($fieldDeclaration['secondPrecision']) && $fieldDeclaration['secondPrecision']?0:($fieldDeclaration['precision'] ?: 10) === 10`

kdaniel95:
@kylekatarnls Could you please confirm if above approach is correct or if I have to supply the flag from somewhere else? Thank you!

kylekatarnls:
Getting the flag from here would be fine, then as an alternative to nested ternary, the code could be like:

```php
$precision = $fieldDeclaration['precision'] ?: 10;

if ($fieldDeclaration['secondPrecision'] ?? false) {
    $precision = 0;
}

if ($precision === 10) {
    $precision = DateTimeDefaultPrecision::get();
}
```

kdaniel95:
@kylekatarnls Okay, got it, changed it and the tests as well. Could you check it out and if all is good (there's no breaking change or things of sort), run the other workflows (and hopefully at the end, merge the changes)?

Thank you!

tests/Doctrine/CarbonTypesTest.php
```diff
@@ -64,10 +64,20 @@ public function testGetSQLDeclaration(string $name)
...
         ], new MySQL57Platform()));
 
+        $this->assertSame('TIMESTAMP(0)', $type->getSQLDeclaration([
+            'precision' => null,
+            'secondPrecision' => true,
+        ], new DB2Platform()));
+
         $this->assertSame('TIMESTAMP(6)', $type->getSQLDeclaration([
             'precision' => null,
         ], new DB2Platform()));
```
kylekatarnls, line 83:
```suggestion
        ], new DB2Platform()));

        $this->assertSame('TIMESTAMP(6)', $type->getSQLDeclaration([
            'precision' => 0,
        ], new DB2Platform()));

        $this->assertSame('DATETIME(6)', $type->getSQLDeclaration([
            'precision' => 0,
        ], new MySQL57Platform()));
```

kdaniel95: Okay, I applied suggested change.

kylekatarnls (approved):

tests/Doctrine/CarbonTypesTest.php
kylekatarnls: 🙏

PR #2550, opened 2022-02-17
Lang/En: Ordinal function: int === float, Fixes #2549

### LINKED ISSUE #2549: PHP 8.1 Implicit conversion from float to int loses precision
Hello,

```php
echo Carbon::parse('2018-06-15')->isoFormat('Do');
```

Carbon version: 2.56.0

PHP version: 8.1.2

I expected to get:

```
15th
```

But I actually get:

```
"Error: Implicit conversion from float 1.5 to int loses precision"
```

The issue is in src/Carbon/Lang/en.php on line 75:

```php
    'ordinal' => function ($number) {
        $lastDigit = $number % 10;

        return $number.(
            (~~($number % 100 / 10) === 1) ? 'th' : (
                ($lastDigit === 1) ? 'st' : (
                    ($lastDigit === 2) ? 'nd' : (
                        ($lastDigit === 3) ? 'rd' : 'th'
                    )
                )
            )
        );
    },

```

Specifically in:

```php
~~($number % 100 / 10) === 1
```

Thanks!

levu42:
Compare float to float, not float to int

Fix #2549

### DIFF
--- a/src/Carbon/Lang/en.php
+++ b/src/Carbon/Lang/en.php
@@ -72,7 +72,7 @@ return [
         $lastDigit = $number % 10;
 
         return $number.(
-            (~~($number % 100 / 10) === 1) ? 'th' : (
+            (~~($number % 100 / 10) === 1.0) ? 'th' : (
                 ($lastDigit === 1) ? 'st' : (
                     ($lastDigit === 2) ? 'nd' : (
                         ($lastDigit === 3) ? 'rd' : 'th'


kylekatarnls (changes requested):
Thanks for the report and the work on it🙏 

As per tests we now have: `'11st'` which is not correct, we should have `'11th'`.

For a proper fix for this it would need:
- a unit test to show the case we fix (so we can ensure it does not regress and we see how the "Implicit conversion" error can be triggered and be sure the fix actually avoid this error to occur)
- fix all languages using the same principle `~~` or `% a / b`, I think it's only used for selecting numbers from 10 to 19 (maybe this can be done is a more explicit way)
- keep compatibility from PHP 7.1.8

Would you be possible for you to update this pull-requests with those items?

levu42:
@kylekatarnls Thanks for your feedback! I'll work on fixing all the mentioned points in the next days!

kylekatarnls:
Deprecations are now converted into exception in the unit tests so they will be detected by the build. Then those int cast are now fixed, you can try `composer require "nesbot/carbon:dev-master as 2.58.0"` if you want to test the next version.

Issue #2529, opened 2022-01-03
Comparisons show wrong result


tflori:
Hello,

I encountered an issue with the following code:
```php
Carbon::setToStringFormat('d.m.Y \a\t h:i a');
$dt = Carbon::parse('2022-05-03');
var_dump($dt->gt('2021-05-03'));
```

Carbon version: 2.55.2

PHP version: PHP 8.0.13

I expected to get:

```
bool(true)
```

But I actually get:

```
bool(false)
```

Ok, enough of your template. Let me explain. Someone in our company thought it works. And indeed it might work most of the time because noone is touching the to string format and php is smart enough to compare `'2022-05-03' > '2021-05-03'`. Also the documentation does not say it is save to use strings.

PHP itself also is very broken in this regard as the DateTime cannot be converted to string:
```
>>> $dt = new DateTime('2022-05-03')
=> DateTime @1651536000 {#189
     date: 2022-05-03 00:00:00.0 UTC (+00:00),
   }
>>> $dt > '2022-06-03'
=> true
>>> $dt > '2021-06-03'
=> true
>>> $dt < '2021-06-03'
=> false
>>> $dt < '2022-06-03'
=> false
```

The comparisons with strings seem to work more accendtially than anything else. Nevertheless the inline documetnations tells us it is working with strings (see: https://github.com/briannesbitt/Carbon/blob/2.55.2/src/Carbon/Traits/Comparison.php#L147). But that is only true if you don't change the toStringFormat.

My suggestion is to check if the given parameter in the comparison methods is a `DateTime` object and if not call `::resolveCarbon($date)` or directly call `::resolveCarbon($date)`. Another option would be to throw but that's a breaking change.

Thanks!

kylekatarnls:
Hello,

I agree. We should use `resolveCarbon` as we do with most of the methods. Here we want to get any string that `parse()` is able to handle and don't want to rely to `setToStringFormat`.

Meanwhile it get changed and released, I suggest to use instead:
```php
$dt = Carbon::parse('2022-05-03');
var_dump($dt->gt(Carbon::parse('2021-05-03')));
```
Or simply:
```php
var_dump($dt > Carbon::parse('2021-05-04'));
```

Issue #2544, opened 2022-01-27
Comparison breaking changes in 2.56.0


TMMeijers:
Hello,

I encountered an issue with the following code:
```php
Carbon\Carbon::now()->lessThan(null);
```

Breaking changes were introduced with #2530 as `null` is now resolved into a Carbon instance `lessThan(null)` returns a different result.

Carbon version: **2.56.0**

PHP version: **8.0.13**

I expected to get:

```
false
```

But I actually get:

```
true
```

Thanks!

kylekatarnls:
Hello 👋 

I understand it can be handy to support nullable values here, but it was never really intended to.

I would recommend to avoid comparing `null` to any kind of thing.

There is no consensus of what should be the expected result:

```php
var_dump(new DateTime() < null);
> false

var_dump(new DateTime() > null);
> true
```

```sql
SELECT NOW() < NULL
> NULL

SELECT NOW() < NULL
> NULL
```

And the same applies to the next person reading your code. This person will have no particular reason to think it will return `false`.

You better can give the answer explicitly to the next reader or the future you with no ambiguity:
```php
$myVariable = null; // or a Carbon instance

if ($myVariable !== null && Carbon::now()->lessThan($myVariable)) {
    // This is a date in the future
}
```

Or even better:
```php
$myVariable = null; // or a Carbon instance

if ($myVariable instanceof CarbonInterface && $myVariable->isFuture()) {
    // This is a date in the future
}
```
And you get a code really clear about what it's doing without having to take any assertion or doing retro-engineering.

I will weight if it's worth to change it as it would be breaking for people that started to use the new version and may have asserted that `null` resolve to `now`. I would even be tempted to say the correct behavior should be to raise an exception when passing to such method a value that is not comparable to a date.

kylekatarnls:
Yes, when those BC can be noticed shortly and show a high number of users impacted, I typically revert it as per the policy described here:
https://github.com/briannesbitt/Carbon/blob/master/contributing.md#note-about-semantic-versioning-and-breaking-changes

If a revert comes after few days and just fixes an edge-case, it just does more arm than good, it's actually an other BC for the people who started to use the new feature. And honestly I think the previous behavior was actually wrong and inconsistent with the rest of the methods, sometimes you would just not increase the major version, when your lib says 2 + 2 = 5, you just consider it a bug and hope no one took this result for granted.

[...]

PR #2553, opened 2022-02-23
LessThan comparisons with null or empty string

zencodex:
the issue was introduced from 2.56.0 to fix https://github.com/briannesbitt/Carbon/issues/2529

I saw https://github.com/briannesbitt/Carbon/issues/2544 conversations, and I met same issue.

There is another issue isn't mentioned,

```
>>> Carbon::now()->lessThanOrEqualTo('')
=> false

>>> Carbon::now()->lessThan('')
=> true
```

`lessThanOrEqualTo` and `lessThan` with different result was confused, I think it should be better if carbon deal with, and keep same result with version < 2.56.0, it would be appreciated if we don't need to change original code in project.

@kylekatarnls so, would you pls double check? and I also add test with null or empty, thanks.

### DIFF
--- a/src/Carbon/Traits/Comparison.php
+++ b/src/Carbon/Traits/Comparison.php
@@ -155,7 +155,7 @@ trait Comparison
      */
     public function greaterThan($date): bool
     {
-        return $this > $this->resolveCarbon($date);
+        return !$date || $this > $this->resolveCarbon($date);
     }
 
     /**
@@ -256,7 +256,7 @@ trait Comparison
      */
     public function lessThan($date): bool
     {
-        return $this < $this->resolveCarbon($date);
+        return $date && $this < $this->resolveCarbon($date);
     }
 
     /**
--- a/tests/Carbon/ComparisonTest.php
+++ b/tests/Carbon/ComparisonTest.php
@@ -360,4 +360,17 @@ class ComparisonTest extends AbstractTestCase
 
         $this->assertSame('06.300000', $baseDate->farthest($closestDate, $farthestDate)->format('s.u'));
     }
+
+    public function testWithNullOrEmpty()
+    {
+        $dt = Carbon::now();
+        $this->assertFalse($dt->lessThan(null));
+        $this->assertFalse($dt->lessThan(''));
+
+        $this->assertFalse($dt->lessThanOrEqualTo(null));
+        $this->assertFalse($dt->lessThanOrEqualTo(''));
+
+        $this->assertTrue($dt->greaterThan(null));
+        $this->assertTrue($dt->greaterThan(''));
+    }
 }


kylekatarnls:
Indeed `lessThanOrEqualTo` and `greaterThanOrEqualTo` are inconsistent. We should be able to pass string here. And to be consistent with the other methods, it should consider `''` or `null` as "now".

I must concede this change was a bad move, but sadly working around like this to restore the behavior we had one month ago won't undo the harm, it will just make a second breaking change and more complains to deal with after the release.

I still think that if your code is doing `$date->lessThan("")` or `$date->lessThan(null)` it's just unclear for who will read it to understand what does mean the question "Is a date less than an empty string? Less than null?" It makes actually no sense to compare a date to something that isn't a date with `<`, `>`, `<=` or `>=` operators. And either it should return `true` or `false` in such case is arbitrary and one or the other can be randomly more convenient for each different case in applications.

Honestly if you have things like this in your code, replacing them is the better option anyway, your code base readability will be improved, to match behavior of `< 2.56.0`:
```php
$date->lte($mixed)
$date->lessThanOrEqualTo($mixed)
// =>
$mixed instanceof \DateTimeInterface && $date <= $mixed


$date->isBefore($mixed)
$date->lt($mixed)
$date->lessThan($mixed)
// =>
$mixed instanceof \DateTimeInterface && $date < $mixed


$date->gte($mixed)
$date->greaterThanOrEqualTo($mixed)
// =>
!($mixed instanceof \DateTimeInterface) || $date >= $mixed


$date->isAfter($mixed)
$date->gt($mixed)
$date->greaterThan($mixed)
// =>
!($mixed instanceof \DateTimeInterface) || $date > $mixed


$date->eq($mixed)
$date->equalTo($mixed)
// =>
$mixed instanceof \DateTimeInterface && $date == $mixed
```

`$mixed instanceof \DateTimeInterface` can be simplified to truthy-checks (e.g. just `$mixed`) if `$mixed` is guaranteed to be `\DateTimeInterface|null|false` or completely skipped if `$mixed` is guaranteed to be `\DateTimeInterface` which is the best case if you can get rid of null/false values earlier.

For instance for a validation check, if you have:
```php
if (Carbon::now()->lessThan($nullableDate)) {
  throw new \Exception('Date is null or in the future');
}
```

I recommend to transform it into;
```php
if (!$nullableDate) {
  throw new \Exception('Date is null');
}

if (Carbon::now() < $nullableDate) {
  throw new \Exception('Date is in the future');
}
```

kylekatarnls:
In 2.61, `lessThanOrEqualTo` and `greaterThanOrEqualTo` will both consider `""` is `"now"`, for the record, this behavior comes from the parent PHP `DateTime` class: `new DateTime('')`

But we'll now discourage to pass `false` and `null`:
https://github.com/briannesbitt/Carbon/commit/d24f53f6316eae8dd8d26f2255989d5d90f7b9de

Even if it will still work to pass empty string `""` I really discourage to do that, it would be really not clear for next developer reading it or future what to expect from a comparison between a date and empty string. (e.g. what is it supposed to mean?)

So for both current and next versions it's recommended that you use 2 conditions for that in order to explicitly tell if `""` should be `true` or `false` in your context:

`false`:
```php
if ($myString !== '' && Carbon::now()->lessThan($myString))
```

`true`:
```php
if ($myString === '' || Carbon::now()->lessThan($myString))
```

PR #2554, opened 2022-02-23
Grammar fixes

sumonst21:
French needs to be capitalized when not used as verb.
Past participle issue or typo in same line "be" should be "been"

### DIFF
--- a/src/Carbon/Lang/fr.php
+++ b/src/Carbon/Lang/fr.php
@@ -90,7 +90,7 @@ return [
     'weekdays_min' => ['di', 'lu', 'ma', 'me', 'je', 've', 'sa'],
     'ordinal' => function ($number, $period) {
         switch ($period) {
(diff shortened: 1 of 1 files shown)


kylekatarnls (approved):

src/Carbon/Lang/fr.php
```diff
@@ -90,7 +90,7 @@
     'weekdays_min' => ['di', 'lu', 'ma', 'me', 'je', 've', 'sa'],
     'ordinal' => function ($number, $period) {
         switch ($period) {
-            // In french, only the first has be ordinal, other number remains cardinal
+            // In French, only the first has been ordinal, other number remains cardinal
```
kylekatarnls, line 93: This is what I meant actually:
```suggestion
            // In French, only the first has to be ordinal, other number remains cardinal
```

PR #2560, opened 2022-03-09
Improve float comparison precision in tests

rgson:
PHPUnit 9.5.14 increased the precision of float comparisons from
0.0000000001 to PHP_FLOAT_EPSILON, causing some tests to fail.


kylekatarnls (approved):

kylekatarnls:
Thanks 🙏

PR #2563, opened 2022-03-10
Fix CreateFromTimeTest

crhg:
In the test when timezone is specified, `Carbon::now()` is used to calculate the expected date. If the test is run in an environment that defaults to a timezone other than the 'Europe/London' specified in the test, the test may fail because the dates obtained will be different for some hours of the day.
Specifying the time zone when asking for the expected date solves this problem.

### DIFF
--- a/tests/Carbon/CreateFromTimeTest.php
+++ b/tests/Carbon/CreateFromTimeTest.php
@@ -97,14 +97,14 @@ class CreateFromTimeTest extends AbstractTestCase
     public function testCreateFromTimeWithDateTimeZone()
     {
         $d = Carbon::createFromTime(12, 0, 0, new DateTimeZone('Europe/London'));
-        $this->assertCarbon($d, Carbon::now()->year, Carbon::now()->month, Carbon::now()->day, 12, 0, 0);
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Default timezone is enforced in test by `AbstractTestCase::setUp`:

```php
date_default_timezone_set('America/Toronto');
```

And now is mocked:
```php
/** @var Carbon $now */
$now = $this->oldNow
    ? Carbon::create(2017, 6, 27, 13, 14, 15, 'UTC')
    : Carbon::now();

/** @var CarbonImmutable $immutableNow */
$immutableNow = $this->oldImmutableNow
    ? CarbonImmutable::create(2017, 6, 27, 13, 14, 15, 'UTC')
    : CarbonImmutable::now();

Carbon::setTestNowAndTimezone($this->now = $now);
CarbonImmutable::setTestNowAndTimezone($this->immutableNow = $immutableNow);
```

GitHub Actions does not run on Europe/London timezone neither. Can you please demonstrate how you make it fail?

kylekatarnls (approved):

Issue #2536, opened 2022-01-13
Carbon\CarbonInterval::setLocalTranslator(): Argument #1 ($translator) must be of type Symfony\Component\Translation\TranslatorInterface, Carbon\Translator given


techmahedy:
I am facing this issue with Laravel 8 version and my php version is 7.4

kylekatarnls:
Please provide:

- [ ] your Carbon version;
- [ ] your Composer version;
- [ ] minimal steps to reproduce your case;

`Carbon\Translator` does implement this interface. So my wild guess is that the vendor installation is incomplete.

Be sure you have in your vendor directory at least:
- symfony/translation
- nesbot/carbon/src
- nesbot/carbon/lazy

Thanks.

techmahedy:
Composer version 2.0.8 and nesbot\carbon version is ^2.0. 

Be sure you have in your vendor directory at least:

- symfony/translation
- nesbot/carbon/src
- nesbot/carbon/lazy

All the folders are there. But still facing this error

kylekatarnls:
Hello `^2.0` can be anything from 2.0.0 to 2.55.2. As stated in the issue template, please run the command `composer show nesbot/carbon` to get the exact version.

Please also try to rebuild your autoload with `php artisan dump-autoload` or a reinstall with `composer install`

fisharebest:
I have the same error, and I have been able to identify the cause.

My projects creates a distribution file (.zip) containing all the vendor files.  Users upgrade by unpacking the latest zip file.  This means that old files from previous versions not be deleted.  These are generally harmless.

[Yes, we should instruct users to delete old files, but upgrades have been kept intentionally simple to encourage users to upgrade regularly...]

Recently, my project upgraded from symfony 4 to symfony 5, which involved an interface moving from one namespace to another: `Symfony\Component\Translation\TranslatorInterface` -> `Symfony\Contracts\Translation\TranslatorInterface`.
            
Carbon uses the Symfony 4 class, and creates an alias from the symfony 5 class.
The code is here: https://github.com/briannesbitt/Carbon/blob/2.56.0/src/Carbon/Traits/Localization.php#L26-L31

Now, if the you are using symfony 5 **and** the old symfony 4 file still exists, then this logic fails.

The ideal solution is for carbon to reverse the logic.  i.e. to code against the symfony 5 class, and create the alias to the symfony 4 class when the symfony 5 class does not exist.

[...]

PR #2566, opened 2022-03-12
Fix: #2536 - improve test for Symfony v4/5 contracts

fisharebest:
Only one of the two interface files should exist.  Hence these two statements should be equivalent:

* `if (!interface_exists('Symfony\\Component\\Translation\\TranslatorInterface'))`
* `if (interface_exists('Symfony\\Contracts\\Translation\\TranslatorInterface'))`

If both interface files exist (i.e. the old one is still present), then replacing the first test with the second one solves #2536 - but only for PHP 7.2 and upwards.  For PHP 7.1, we need the original test.

The following code will fix this problem and will pass the unit tests. 
I also tested on my own application using PHP 7.1 - 8.1.

```
if (
    PHP_VERSION_ID < 72000 && !interface_exists('Symfony\\Component\\Translation\\TranslatorInterface') ||
    PHP_VERSION_ID >= 72000 && interface_exists('Symfony\\Contracts\\Translation\\TranslatorInterface')
) {
    class_alias(
        'Symfony\\Contracts\\Translation\\TranslatorInterface',
        'Symfony\\Component\\Translation\\TranslatorInterface'
    );
}
```

I am not completely happy with this - I haven't managed to work out exactly why the PHP dependency is relevant.

### DIFF
--- a/src/Carbon/Traits/Localization.php
+++ b/src/Carbon/Traits/Localization.php
@@ -23,7 +23,10 @@ use Symfony\Component\Translation\TranslatorInterface;
 use Symfony\Contracts\Translation\LocaleAwareInterface;
 use Symfony\Contracts\Translation\TranslatorInterface as ContractsTranslatorInterface;
 
-if (!interface_exists('Symfony\\Component\\Translation\\TranslatorInterface')) {
+if (
+    PHP_VERSION_ID < 72000 && !interface_exists('Symfony\\Component\\Translation\\TranslatorInterface') ||
+    PHP_VERSION_ID >= 72000 && interface_exists('Symfony\\Contracts\\Translation\\TranslatorInterface')
+) {
     class_alias(
         'Symfony\\Contracts\\Translation\\TranslatorInterface',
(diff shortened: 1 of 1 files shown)


kylekatarnls:
> these two statements should be equivalent

This is incorrect because `interface_exists` has `$autoload = true` by default, which means it tries to load from composer the dependency you pass (so Contracts vs. Component will make a different autoload attempt).

Also it will behave differently if the file is included/required multiple time (`!interface_exists()` never fails in case of multiple loads), `interface_exists()` allows the file to be loaded only once. This means `!interface_exists('A') && interface_exists('B')` and `interface_exists('B') && !interface_exists('A')` can also give different results.

There is very few chances that `PHP_VERSION_ID >= 72000` is a root cause for a difference, more likely different version of the composer dependencies are installed before/after this version, which would be the real cause. And in this case, it can not be the correct trigger in real applications (if those applications have other packages locking down some package, platform config or other settings that may enforce different versions of the packages.

I will try few things.

kylekatarnls:
Would #2586 work for you as it sounds to me safer.

PR #2577, opened 2022-03-27
Always use strict in_array comparison.

lucasmichot:
Always use strict `in_array` comparison.

### DIFF
--- a/src/Carbon/Lang/cy.php
+++ b/src/Carbon/Lang/cy.php
@@ -60,7 +60,7 @@ return [
     'ordinal' => function ($number) {
         return $number.(
             $number > 20
-                ? (\in_array($number, [40, 50, 60, 80, 100]) ? 'fed' : 'ain')
+                ? (\in_array($number, [40, 50, 60, 80, 100], true) ? 'fed' : 'ain')
                 : ([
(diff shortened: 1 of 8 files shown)


src/Carbon/Lang/cy.php
```diff
@@ -60,7 +60,7 @@
     'ordinal' => function ($number) {
         return $number.(
             $number > 20
-                ? (\in_array($number, [40, 50, 60, 80, 100]) ? 'fed' : 'ain')
+                ? (\in_array($number, [40, 50, 60, 80, 100], true) ? 'fed' : 'ain')
```
kylekatarnls, line 63:
```suggestion
                ? (\in_array((int) $number, [40, 50, 60, 80, 100], true) ? 'fed' : 'ain')
```

src/Carbon/Traits/Units.php
```diff
@@ -167,7 +167,7 @@ public static function isModifiableUnit($unit)
             'weekday',
         ];
 
-        return \in_array($unit, $modifiableUnits) || \in_array($unit, static::$units);
+        return \in_array($unit, $modifiableUnits) || \in_array($unit, static::$units, true);
```
kylekatarnls, line 170:
```suggestion
        return \in_array($unit, $modifiableUnits, true) || \in_array($unit, static::$units, true);
```

kylekatarnls (approved):

PR #2587, opened 2022-04-09
Allow grammar rule customization for unit declensions

kylekatarnls:
This will allow declension in Mongolian to apply only to the last part when the interval is a sequence of units:

When last unit can be shorten, we just get no declensions:
```php
Carbon::now()
    ->subYears(2)
    ->subMonths(3)
    ->subDay()
    ->subSecond()
    ->locale('mn')
    ->diffForHumans(null, null, true, 4)
```
```
2 жил 3 сар 1 өдөр 1с өмнө
```

When last unit is entire, we get declension only on this last unit:
```php
Carbon::now()
    ->subYears(2)
    ->subMonths(3)
    ->subDay()
    ->subSecond()
    ->locale('mn')
    ->diffForHumans(null, null, false, 4)
```
```
2 жил 3 сар 1 өдөр 1 секундын өмнө
```

👋  @lucifer-crybaby @Batmandakh your review is welcome. 🙏

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -1620,18 +1620,23 @@ class CarbonInterval extends DateInterval implements CarbonConverterInterface
         $unit = $short ? 's' : 'second';
         $isFuture = $this->invert === 1;
         $transId = $relativeToNow ? ($isFuture ? 'from_now' : 'ago') : ($isFuture ? 'after' : 'before');
+        $declensionMode = null;
 
         /** @var \Symfony\Component\Translation\Translator $translator */
         $translator = $this->getLocalTranslator();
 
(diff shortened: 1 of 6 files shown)


src/Carbon/CarbonInterval.php
```diff
@@ -2752,4 +2763,14 @@ public function ceil($precision = 1)
     {
         return $this->round($precision, 'ceil');
     }
+
+    private function needsDeclension(string $mode, int $index, int $parts): bool
+    {
```
lucasmichot, line 2768: or just `return $mode !== 'last' || $index === $parts - 1;` ?

kylekatarnls: I think the switch will keep easier to add more modes and more clear about what is the grammar rule.

PR #2590, opened 2022-04-15
Add \DateTime and \DateTimeImmutable support to the Test::setTestNow()

yuraplohov:
This PR provides the ability to use \DateTimeInterface as the argument type for the Test::setTestNow() method.

Currently this method can receive instances of CarbonInterface, but cannot receive instances of \DateTime and \DateTimeImmutable php-classes exactly.

Support for \DateTime and \DateTimeImmutable types will provide more flexibility for client code.

### DIFF
--- a/src/Carbon/Traits/Test.php
+++ b/src/Carbon/Traits/Test.php
@@ -63,7 +63,7 @@ trait Test
      */
     public static function setTestNow($testNow = null)
     {
-        static::$testNow = $testNow instanceof self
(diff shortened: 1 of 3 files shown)


kylekatarnls (approved):
Nice idea, I also realize we can simply rely on make() so what can be passed in setTestNow() will simply be consistent with make/parse/constructor and the behavior with currently supported values DateTimeInterface/CarbonInterface/Closure/string/null/false is still the same.

PR #2599, opened 2022-04-20
Simplify some conditions

lucasmichot:
This PR simplifies some conditions and promote early returns.

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -1823,7 +1823,7 @@ class CarbonInterval extends DateInterval implements CarbonConverterInterface
     {
         $format = $this->localToStringFormat;
 
-        if ($format === null) {
+        if (!$format) {
             return $this->forHumans();
         }
 
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -1435,11 +1435,9 @@ class CarbonInterval extends DateInterval implements CarbonConverterInterface
             ];
         }
 
-        if ($altNumbers) {
-            if ($altNumbers !== true) {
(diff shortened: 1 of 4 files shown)


src/Carbon/CarbonInterval.php
```diff
@@ -1435,11 +1435,9 @@ protected function getForHumansParameters($syntax = null, $short = false, $parts
             ];
         }
 
-        if ($altNumbers) {
-            if ($altNumbers !== true) {
-                $language = new Language($this->locale);
-                $altNumbers = \in_array($language->getCode(), (array) $altNumbers, true);
-            }
+        if ($altNumbers !== true) {
```
kylekatarnls, line 1438:
```suggestion
        if ($altNumbers && $altNumbers !== true) {
```
Normally it's not equivalent if we allow anything not `true` instead of any truthy value except `true`.

Tests did not fail by luck, because if it's false/null/[] then \in_array() will return `false` as a language code string will never match. Still we execute some code and assignation that are not needed in this case.

src/Carbon/CarbonInterval.php
```diff
@@ -1825,15 +1823,15 @@ public function __toString()
     {
         $format = $this->localToStringFormat;
 
-        if ($format) {
-            if ($format instanceof Closure) {
-                return $format($this);
-            }
+        if ($format === null) {
```
kylekatarnls, line 1826: This is maybe not a case we need to support, but this will change the behavior when `$format` is an empty string (e.g. for `->settings(['toStringFormat' => ''])`)

For now I would stick to:
```suggestion
        if (!$format) {
```
So the behavior is the same and consistent with the date classes.

PR #2600, opened 2022-04-20
Remove redundant initializers

lucasmichot:
This PR removes redundant initializers.

### DIFF
--- a/src/Carbon/TranslatorImmutable.php
+++ b/src/Carbon/TranslatorImmutable.php
@@ -18,7 +18,7 @@ use Symfony\Component\Translation\Formatter\MessageFormatterInterface;
 class TranslatorImmutable extends Translator
 {
     /** @var bool */
-    private $constructed = false;
+    private $constructed;
 
     public function __construct($locale, MessageFormatterInterface $formatter = null, $cacheDir = null, $debug = false)
     {


src/Carbon/TranslatorImmutable.php
```diff
@@ -18,7 +18,7 @@
 class TranslatorImmutable extends Translator
 {
     /** @var bool */
-    private $constructed = false;
+    private $constructed;
```
kylekatarnls, line 21: `parent::__construct()` must be allowed to call `$this->setLocale()`, at this moment it's not yet `true`, so `$this->constructed` is "not set". As the "not set"-state is still handled weirdly by PHP when it's not typed (because of backward compatibility), then it's just `null` when we do `if ($this->constructed)` in `disallowMutation`.

`null` is falsy so it's behaving the same. But it's actually a mismatch with `@var bool` typehint, and as soon as we'll pass to real type `private bool $constructed;` in Carbon 3 (as we drop PHP 7 support), 💥 we'll get:

```
Typed property Carbon\TranslatorImmutable::$constructed must not be accessed before initialization
```

So it will need either to be `private bool $constructed = false;` again or `private ?bool $constructed = null;` those are also marked as "redundant initializers" by static analyzers but actually they're not as they allow us to check if `__construct` method is complete or not.

lucasmichot:
True @kylekatarnls I did not get in consideration the Carbon 3 move 

Thanks for the in-deep review though 👍

PR #2601, opened 2022-04-20
Remove useless casting

lucasmichot:
This PR removes useless casting

### DIFF
--- a/src/Carbon/Traits/Difference.php
+++ b/src/Carbon/Traits/Difference.php
@@ -472,7 +472,7 @@ trait Difference
      */
     public function floatDiffInSeconds($date = null, $absolute = true)
     {
-        return $this->diffInMicroseconds($date, $absolute) / static::MICROSECONDS_PER_SECOND;
+        return (float) ($this->diffInMicroseconds($date, $absolute) / static::MICROSECONDS_PER_SECOND);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Carbon/Traits/Difference.php
```diff
@@ -472,7 +472,7 @@ public function diffInRealMilliseconds($date = null, $absolute = true)
      */
     public function floatDiffInSeconds($date = null, $absolute = true)
     {
-        return $this->diffInMicroseconds($date, $absolute) / static::MICROSECONDS_PER_SECOND;
+        return (float) ($this->diffInMicroseconds($date, $absolute) / static::MICROSECONDS_PER_SECOND);
```
lucasmichot, line 475: I also force this cast there as `diffInMicroseconds` always return `int`

src/Carbon/Traits/Localization.php
```diff
@@ -235,7 +235,7 @@ public static function translateWith(TranslatorInterface $translator, string $ke
             : $translator->transChoice($key, $number, $parameters);
         // @codeCoverageIgnoreEnd
 
-        return (string) $choice;
+        return $choice;
```
kylekatarnls, line 238: Not guaranteed with symfony/translation 4

lucasmichot: reverted this one

kylekatarnls (approved):

PR #2609, opened 2022-05-14
No need to use version_compare when the constant is sufficient.

lucasmichot:
No need to use `version_compare` when the constant is sufficient.

### DIFF
--- a/tests/Carbon/IsTest.php
+++ b/tests/Carbon/IsTest.php
@@ -977,7 +977,7 @@ class IsTest extends AbstractTestCase
     {
         $output = Carbon::now()->format($letter);
 
-        if ($output === '1000' && $letter === 'v' && version_compare(PHP_VERSION, '7.2.12', '<')) {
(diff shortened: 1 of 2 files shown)


tests/Carbon/IsTest.php
```diff
@@ -977,7 +977,7 @@ public function testHasFormatWithSingleLetter($letter)
     {
         $output = Carbon::now()->format($letter);
 
-        if ($output === '1000' && $letter === 'v' && version_compare(PHP_VERSION, '7.2.12', '<')) {
+        if ($output === '1000' && $letter === 'v' && PHP_VERSION_ID < 70212) {
```
kylekatarnls, line 980: I might be silly, but I struggle to read it (confuse with 7.0; on 3.x maybe we can use the PHP 7.4 syntax for clarity and write `7_02_12`) and I would prefer to keep the version as readable as possible. Mostly in tests where we don't care to much about being fast to execute it.

PR #2616, opened 2022-05-19
Add exception for endless periods

### LINKED ISSUE #2615: Allowed memory fatal error by count method on CarbonPeriod instance with null end property
Hello,

```php
(new \Carbon\CarbonPeriod())->count();
```

Carbon version: 2.57.0

PHP version: 7.4.29

I expected to get (feature proposal) :

Some kind of exception thrown by count on CarbonPeriod with null `end` property

But I actually get:

```
Fatal error: Allowed memory size of XXXXXXX bytes exhausted
```

Thanks!

kylekatarnls:
Fix #2615


src/Carbon/CarbonPeriod.php
```diff
@@ -1534,13 +1535,50 @@ public function toDatePeriod()
...
+
     /**
      * Convert the date period into an array without changing current iteration state.
      *
      * @return CarbonInterface[]
      */
     public function toArray()
     {
+        if ($this->isUnfilteredAndEndLess()) {
+            throw new EndLessPeriodException("Endless period can't converted to array nor counted.");
```
kylekatarnls, line 1579:
```suggestion
            throw new EndLessPeriodException("Endless period can't be converted to array nor counted.");
```

PR #2623, opened 2022-06-10
Correction of time and date position

swarakaka:

PR #2624, opened 2022-06-16
Update CkbTest.php

swarakaka:


swarakaka:
Please reopen it again, I accidentally deleted my PR.

kylekatarnls:
I can but CkbTest.php should be modified in the same branch/PR than cbk.php so we can actually see what's the result of the change without having the tests job broken.

swarakaka:
Please tell me what I can do? Because all the texts are now identical to the test.

kylekatarnls:
Reopen PR did not restore the branches, do you have a "Restore branch" button? So if you can restore patch-2 and patch-4 I'll can merge them easily.

swarakaka:
According to what GitHub describes [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/deleting-and-restoring-branches-in-a-pull-request), no such action is available.

kylekatarnls:
Ah, actually, it does not matter, as it didn't change since last time I merged both branches.

I can see last commit here is [8db8d64](https://github.com/briannesbitt/Carbon/pull/2624/commits/8db8d645ed8cf606ebeec6ef255f6517540d1431)

And you can see it's already merged in #2623

 Still the tests do not pass on #2623

If CkbTest.php is correct (properly reflect what the output should be for the language, then it would mean that ckb.php still need some adjustments to match.

PR #2627, opened 2022-06-18
Update ckb.php

swarakaka:


swarakaka:
Hello @kylekatarnls ,
I apologize for repeating the PR, because the previous one did not belong to any branch. Please add the test. Then I can do a local test?

swarakaka:
Hello @kylekatarnls ,
Is there still a problem with the test?

kylekatarnls:
There is no problem with tests, they just fail when actual output differ from expected output, see the logs:
https://github.com/briannesbitt/Carbon/runs/7291475334?check_suite_focus=true#step:11:119

```
1) Tests\Localization\CkbTest::testLanguage
In ckb, Carbon::now()->subSeconds(1)->diffForHumans() should return پێشتر چرکە
--- Expected
+++ Actual
@@ @@
-'پێشتر چرکە'
+'پێش چرکەیەک'
/home/runner/work/Carbon/Carbon/tests/Localization/LocalizationTestCase.php:364
/home/runner/work/Carbon/Carbon/tests/AbstractTestCase.php:232
/home/runner/work/Carbon/Carbon/tests/AbstractTestCase.php:239
/home/runner/work/Carbon/Carbon/tests/Localization/LocalizationTestCase.php:367
```

Which is kind of normal to have `CkbTest` failing when changing `ckb.php`.

So I updated the test according to your change:
https://github.com/briannesbitt/Carbon/pull/2627/commits/6a3404d747ce4591f2ef1ac459fb54eb0382965c

But those tests are useless if I (a non Sorani-speaker) update them blindly.

That's why you should check the commit above, verify it makes sense for the given method calls (given in the comment above each line) and check that green lines are better than red lines.

kylekatarnls (approved):

swarakaka:
I really don't know what to do in this case?!

kylekatarnls:
CkbTest.php are the unit tests for ckb.php file, when you modify a test file, it does not change anything in the library and its behavior, test files are not even included in the final package. They are here to ensure we don't break a feature when we make a change on something else.

So if test pass like it was the case on the previous commit ([6a3404d](https://github.com/briannesbitt/Carbon/pull/2627/commits/6a3404d747ce4591f2ef1ac459fb54eb0382965c)), then changing only the test file will necessarily make the test file, and this is legit as we tell we expect some precise result but this is not the result the library currently output, so it also need ckb.php to be changed to actually produce this new expected behavior. I'm not sure but here you changed it because `و ` is missing? It's the equivalent of the `,` (coma) in English? If so it is driven by the `'list' => ['، ', ' و '],` key and it's optional. It can be added to the result using `->diffForHumans(['join' => true])` so it's expected not to have them when the option is not specified.

Anyway improving translations is something that can be done step by step. It does not need to be perfect at once, if the new version is better than previous (if green lines are better than red lines) then we should merge as it's an improvement, And we can still continue and make other merge-requests for other improvements. It will also be easier we'll have less lines to check.

swarakaka:
No, (و) is not the equivalent of (comma ,), but means (and). In Kurdish, the comma is always (,), as you can see it is written in reverse.

kylekatarnls:
OK, this could be a wider change and it might be the same in other languages. the "list" system is inherited from CLDR, by default we just provide a compact form:

`2 years 3 months 10 days 5 hours` would be clear enough in some "raw" output (data table, summaries, timeline)

The more verbose output that is more natural in a sentence can be get with the `'join' => true` option:

`2 years, 3 months, 10 days and 5 hours` e.g. coma between each unit except the last one separated with " and "

Here is with your change what it gives with and without the option, so for the code:
```php
echo Carbon::now()->addYears(2)->addMonths(3)->addDays(10)->addHours(5)->locale('ckb')->diffForHumans(['parts' => 5]);
echo "\n";
echo Carbon::now()->addYears(2)->addMonths(3)->addDays(10)->addHours(5)->locale('ckb')->diffForHumans(['parts' => 5, 'join' => true]);
```
It would result:
```
دوو ساڵ 3 مانگ هەفتەیەک 3 ڕۆژ 4 کاتژمێر لە ئێستاوە
دوو ساڵ، 3 مانگ، هەفتەیەک، 3 ڕۆژ و 4 کاتژمێر لە ئێستاوە
```
If I understood correctly this is actually correct. Unless this usage of `،` is not correct in Sorani? If not we can fix the list to use only `و` instead between every unit. (Note that it also can be set manually using `->diffForHumans(['parts' => 5, 'join' => ' و '])` if one user wants to change it in a specific case.

Also I would keep the compact variant by default so it like other languages, unless it would be not understandable in Sorani.

That's why I merged your commits in #2647 for the release of 2.60.0 but if it's not good this way, we can still re-open this and make an other patch (2.60.1).

PR #2639, opened 2022-07-14
Watch event for each app

### LINKED ISSUE #2638: Add Laravel Octane Support
Responding to a thread on laravel/octane (https://github.com/laravel/octane/issues/551#issuecomment-1180210039), I discovered a bug in this package.

From what I indicate in the title, this package is not supported when using Laravel Octane.

- What is the problem?
When updating the language of the application using `app()->setLocale('xx')`, the language in carbon is not updated

- Where does the problem originate?
In:
https://github.com/briannesbitt/Carbon/blob/a9000603ea337c8df16cc41f8b6be95a65f4d0f5/src/Carbon/Laravel/ServiceProvider.php#L47
Gets the application from when laravel octane is instantiated, so even if the language is changed, it will continue to return the default language, causing carbon to not update.

- How to solve the problem?
Change the `$this->app` inside `updateLocale` https://github.com/briannesbitt/Carbon/blob/a9000603ea337c8df16cc41f8b6be95a65f4d0f5/src/Carbon/Laravel/ServiceProvider.php#L45
to `Illuminate\Container\Container::getInstance()`

This solution should work whether laravel octane is used or not.

kylekatarnls:
Fix #2638


mcolominas:
I just checked this and this fix doesn't work, it's the same problem, you're getting the app when the provider boot.

PR #2646, opened 2022-07-23
Align ku locale with CLDR data

kylekatarnls:


kylekatarnls:
Hello @swarakaka, if you can read Kurmanji, your review is welcome.

PR #2650, opened 2022-08-02
Updated Simplified Chinese translation

10935336:
It's actually only a few words different from zh_Hans.php, but considering the number of people using Simplified Chinese, that might help.

Opps, maybe i should edit zh_hans.php but it looks good enough.

### DIFF
--- a/src/Carbon/Lang/zh_CN.php
+++ b/src/Carbon/Lang/zh_CN.php
@@ -21,13 +21,87 @@
  * - Chris Hemp
  * - shankesgk2
  */
-return array_merge(require __DIR__.'/zh.php', [
+return [
+    'year' => ':count:optional-space年',
+    'y' => ':count:optional-space年',
+    'month' => ':count:optional-space个月',
+    'm' => ':count:optional-space月',
+    'week' => ':count:optional-space周',
+    'w' => ':count:optional-space周',
+    'day' => ':count:optional-space天',
+    'd' => ':count:optional-space天',
+    'hour' => ':count:optional-space小时',
+    'h' => ':count:optional-space小时',
+    'minute' => ':count:optional-space分钟',
+    'min' => ':count:optional-space分钟',
+    'second' => ':count:optional-space秒',
+    'a_second' => '{1}几秒|]1,Inf[:count:optional-space秒',
+    's' => ':count:optional-space秒',
+    'ago' => ':time前',
+    'from_now' => ':time后',
+    'after' => ':time后',
+    'before' => ':time前',
+    'diff_now' => '现在',
+    'diff_today' => '今天',
+    'diff_yesterday' => '昨天',
+    'diff_tomorrow' => '明天',
     'formats' => [
         'LT' => 'HH:mm',
         'LTS' => 'HH:mm:ss',
         'L' => 'YYYY/MM/DD',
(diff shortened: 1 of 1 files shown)


kylekatarnls:
zh_Hans, zh_Hans_HK, zh_Hans_MO, zh_Hans_SG, yue_Hans are the same (driven by zh_Hans) in Carbon as (according to data from CLDR) there is no difference in date/time stuff.

zh is zh_Hans + formats:
```php
[
    'LT' => 'HH:mm',
    'LTS' => 'HH:mm:ss',
    'L' => 'YYYY/MM/DD',
    'LL' => 'YYYY年M月D日',
    'LLL' => 'YYYY年M月D日 A h点mm分',
    'LLLL' => 'YYYY年M月D日dddd A h点mm分',
]
```

Then zh_CN just remove the spaces from LLL and LLLL formats:
```php
        'LLL' => 'YYYY年M月D日Ah点mm分',
        'LLLL' => 'YYYY年M月D日ddddAh点mm分',
```

I don't say it's correct, but it's what we got and extracted from CLDR.

zh is also inherited by zh_SG and zh_YUE with little differences.

Depending on the scope (if it applies only on zh_CN or if it apply to all zh + inheritence or all zh_Hans + inheritence), the appropriate file zh_CN / zh / zh_Hans should be modified.

If language file changes, unit tests will change too, we can then check what is the new output for various calls to diffForHumans and so check the changes give the expected outcome.

10935336:
> zh_Hans, zh_Hans_HK, zh_Hans_MO, zh_Hans_SG, yue_Hans are the same (driven by zh_Hans) in Carbon as (according to data from CLDR) there is no difference in date/time stuff.
> 
> zh is zh_Hans + formats:
> 
> ```
> [
>     'LT' => 'HH:mm',
>     'LTS' => 'HH:mm:ss',
>     'L' => 'YYYY/MM/DD',
>     'LL' => 'YYYY年M月D日',
>     'LLL' => 'YYYY年M月D日 A h点mm分',
>     'LLLL' => 'YYYY年M月D日dddd A h点mm分',
> ]
> ```
> 
> Then zh_CN just remove the spaces from LLL and LLLL formats:
> 
> ```
>         'LLL' => 'YYYY年M月D日Ah点mm分',
>         'LLLL' => 'YYYY年M月D日ddddAh点mm分',
> ```
> 
> I don't say it's correct, but it's what we got and extracted from CLDR.
> 
> zh is also inherited by zh_SG and zh_YUE with little differences.
> 
> Depending on the scope (if it applies only on zh_CN or if it apply to all zh + inheritence or all zh_Hans + inheritence), the appropriate file zh_CN / zh / zh_Hans should be modified.
> 
> If language file changes, unit tests will change too, we can then check what is the new output for various calls to diffForHumans and so check the changes give the expected outcome.

Yes, I noticed that they are all provided by zh_Hans.php.
`h点mm分` is generally used in spoken language, and `h时mm分` is generally used in written documents.
So I don't think this is very appropriate. If really needed, it should be `h时mm分`, but in the internet it is enough to use `HH:mm`.

<details><summary>Screenshot of government website</summary>
image
</details>

But since this was taken from the CLDR, maybe it represents a wider range of opinions, so I decided to close this PR. 
Sorry for the trouble.

PR #2651, opened 2022-08-05
Add options for microseconds on getDateIntervalSpec()

Nakatox:
https://www.postgresql.org/docs/14/datatype-datetime.html#DATATYPE-INTERVAL-INPUT
Postgres date interval field values can have fractional parts.

https://tc39.es/proposal-temporal/docs/duration.html
Also see the `ECMAScript` Temporal Proposal : `duration.toString` has a `fractionalSecondDigits` option

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -2150,7 +2150,7 @@ class CarbonInterval extends DateInterval implements CarbonConverterInterface
      *
      * @return string
      */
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Note that:

```php
new DateInterval('PT0.012300S')
```

Would throw an exception but duration in ISO 8601 norm allows decimal (by the way with either `.` or `,` separator) for any unit in the interval. So I'm OK to add this flag.

kylekatarnls (approved):

PR #2653, opened 2022-08-14
Added: 'Day, Month Day, Year' as a predefined date format

Axum25:


src/Carbon/Traits/Converter.php
```diff
@@ -157,6 +157,21 @@ public function toFormattedDateString()
     {
         return $this->rawFormat('M j, Y');
     }
+
```
kylekatarnls, line 160:
```suggestion

```

src/Carbon/Traits/Converter.php
```diff
@@ -157,6 +157,20 @@ public function toFormattedDateString()
     {
         return $this->rawFormat('M j, Y');
     }
+    /**
```
kylekatarnls, line 160:
```suggestion

    /**
```

kylekatarnls:
Thanks.

PR #2683, opened 2022-10-09
Allow lowercase macro-format per-language customization

### LINKED ISSUE #2679: Issue with isoFormat('ll') when using 'months_short' as it adds "ta" suffix
Hello,

There seems to be a problem with isoFormat('ll') when using 'months_short' where the "ta suffix" is incorrectly added:

https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/fi.php

I expected to get: tammi', 'helmi', 'maalis', 'huhti', 'touko', 'kesä', 'heinä', 'elo', 'syys', 'loka', 'marras', 'joulu

But I actually get:

tammita', 'helmita', 'maalista', 'huhtita', 'toukota', 'kesäta', 'heinäta', 'elota', 'syysta', 'lokata', 'marrasta', 'jouluta and this makes no sense in Finnish.

Note: If I instead would use 'months' then it would be correct as there "ta" suffix is correct

Note 2: https://en.wiktionary.org/wiki/-ta#Finnish

Thanks!

kylekatarnls:
Fix #2679


cbaconnier:
I think it broke the date format I was using. 
```
$date->isoFormat('Ll');
```

The expected output was `01 janv. 2020` _(French)_ but since the last update I'm getting `01/01/2020`

kylekatarnls:
Hello, mixed case was never intended to provide any particular output, `ll` will give you the same output as `Ll` before this fix:

```
1 janv. 2020
```

(in both case, there is no trailing `0`)

cbaconnier:
Thanks! I suppose it was a typo on our side.

PR #2711, opened 2022-11-25
Fix docblock

michaelnabil230:


src/Carbon/Traits/Date.php
```diff
@@ -641,6 +641,8 @@ public function getTimezone()
     /**
      * List of minimum and maximums for each unit.
      *
+     * @param int $daysInMonth
+     *
```
kylekatarnls, line 645:
```suggestion
     *
```

michaelnabil230:
@kylekatarnls I see many functions need to write the docblock. I can make PR for fix if you need that's

kylekatarnls:
Thanks. 🙏 

I'm fine with fixes for consistency, but while PHPDoc made sense by the past for this project because older PHP versions had a very limited support of the typing, now it's way less useful as IDEs and co. will read the real type hints and `@param` or `@return` annotation add value only when they can be more precise (such as `array<Carbon>` over `array`) or if they have relevant descriptions of the params that may justify to have them.

Also, they are a bit less useful on private and protected methods than on public ones.

So all in all, it does not harm to have docblocks cleaned or to add some descriptions for parameters where the name is not enough to understand what it is. But no pressure, this is not a must-have for 100% of the methods IMO.

Issue #2723, opened 2022-12-11
Incorrect operation type


yarscript:
Hello,

I encountered an issue with the following code:
https://github.com/briannesbitt/Carbon/blob/107ca469b2b6ef6e0b0fc6a6edbb92cbfaf282d9/src/Carbon/Traits/Units.php#L69

Carbon version: 
```json
{
    "name": "nesbot/carbon",
    "version": "2.64.0",
}
```

PHP version: 
```sh
PHP 8.2.0 (cli) (built: Dec  9 2022 16:19:06) (NTS)
```

I expected to get:

```php
   gettype($value); // 'int'
```

But I actually get:

```php
   gettype($value) // 'string'
```

<img width="1340" alt="image" src="https://user-images.githubusercontent.com/26656665/206924836-f395b451-dcad-4a74-a8e9-913ba35b46fe.png">

Thanks!

yarscript:
#2724

yarscript:
Please make sure that is it correct.
Thanks!

kylekatarnls:
Hello, I tried to add a test to reproduce your issue: https://github.com/briannesbitt/Carbon/pull/2726/files but this test is not failing with  PHP 8.2.

Can you please provide a minimum code chunk to trigger this error? So I'll add it in unit tests and it will ensure the fix we merge actually fix it and will prevent regression.

Thanks 🙏

kylekatarnls:
When I try it in PHP 8.2 no error is thrown, it just cast implicitly:
```php
var_dump($value); // "8"
$value *= static::SECONDS_PER_MINUTE;
var_dump($value); // 480
```

I can only trigger the error if I pass an invalid string such as `"rr"` but in this case we clearly want to have an error, we should definitely not silently convert it to `0` which is what happen if we add `(int)` or `(float)` explicit cast.

yarscript:
@kylekatarnls 
In my case I received empty string there but anyway your're right - silently type converting is not for this case.
Thanks a lot for your time!

PR #2724, opened 2022-12-11
Update Units.php

yarscript:
#2723

### DIFF
--- a/src/Carbon/Traits/Units.php
+++ b/src/Carbon/Traits/Units.php
@@ -30,14 +30,14 @@ trait Units
      * Add seconds to the instance using timestamp. Positive $value travels
      * forward while negative $value travels into the past.
      *
-     * @param string $unit
-     * @param int    $value
+     * @param string    $unit
+     * @param int|float $value
      *
      * @return static
      */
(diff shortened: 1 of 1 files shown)


kylekatarnls (approved):

kylekatarnls (approved):

src/Carbon/Traits/Units.php
```diff
@@ -30,13 +30,16 @@ trait Units
...
+     * @param string    $unit
+     * @param int|float $value
      *
      * @return static
      */
     public function addRealUnit($unit, $value = 1)
     {
+        $value = (float) $value;
+        $unit = (string) $unit;
+
```
kylekatarnls, line 42:
```suggestion

```

kylekatarnls:
Hello because of this:
https://github.com/briannesbitt/Carbon/issues/2723#issuecomment-1354704561

We shouldn't make any explicit cast here, or else we should add first an input validation with a specific error such as `"string" is not a valid numeric value`.

Because muting the error and simply replacing any non-numeric string with `0` is just hiding to the developer the incorrect usage and cause bugs later.

PR #2735, opened 2022-12-26
Update SrLatnMeTest.php

perisicnikola37:


kylekatarnls:
Hello, thanks for your contribution, I debugged this and saw that Symfony doesn't pick the correct "sr" pluralization because of the "Latn" or "Cyrl" variant in the locale name.

When I fix it, I see it matches your change but it also modify some other output lines, can you confirm those 4 changes are also correct:
https://github.com/briannesbitt/Carbon/pull/2740/commits/66b9317e43e2bf488b096ddb5a57e0ccec5c20c6#diff-0a9e11e82065a345aa701dbc73fdbda5c3202d13454331cb5d0392788f2adbf0

For instance `1 nedjelja 6 dan` would become `1 nedjelja 6 dana`:
```php
        // Carbon::now()->addWeek()->addDays(6)->diffForHumans(null, true, false, 2)
        // '1 week 6 days'
        '1 nedjelja 6 dana',
```

kylekatarnls:
Merged in #2740

PR #2743, opened 2023-01-01
Update SrLatnMeTest.php

perisicnikola37:


kylekatarnls:
Thanks. 🙏 

Hello @milose @Glavic, invoking Serbian contributors 🧙 here to have a second look and if you can, please proceed the relevant changes on:
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/sr_Latn_ME.php

milose:
Looks good to me @kylekatarnls, thanks @perisicnikola37

milose:
https://github.com/briannesbitt/Carbon/blob/1e60d1b9b3a445a6d450d960019a0a0c2f2a59d3/src/Carbon/Lang/sr_Latn_ME.php#L60 should be changed as per test.

perisicnikola37:
@kylekatarnls Why this is still not merged?

kylekatarnls:
Hello @perisicnikola37 this is only a test file, that does not actually change the actual translations which live in:
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/sr.php
+ the variants Cyrl and Latn

So if I merge this alone, actually test won't pass anymore and it will have no effect for users.

It would need now to change accordingly all language files and tests (if there are other countries that ME that are concerned).

kylekatarnls (approved):

kylekatarnls (changes requested):
Source files need to be updated accordingly:

https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/sr.php

kylekatarnls:
I have another issue, I don't know why I cannot run the tests on this PR, maybe it needs to be rebased on last changes of master branch.

PR #2753, opened 2023-01-14
Remove exit call in unit tests

rgson:
The exit call breaks the test suite by exiting PHPUnit. These calls are presumably just unintentional leftovers of development.


kylekatarnls (approved):

kylekatarnls:
Damn it 🤦

kylekatarnls:
Thanks I will enable PHPMD in the tests folder too #2754 so it will catch dev fragments there too.

PR #2766, opened 2023-02-06
Add Kurdish (Sorani) translation

AlanD20:
Hello, This PR adds the Kurdish (Sorani) language to the primary Kurdish languages and the Turkish version is separated to its own Kurdish version. Unfortunately, the primary Kurdish (Sorani) language was replaced by the Kurdish (Turkish) version a while ago.

Thanks.


kylekatarnls:
Hello, I checked [CLDR Unicode](https://cldr.unicode.org/index/downloads) latest version and it still have ku and ku_TR locales identical. I don't know why Arab and Cyrillic alphabets declensions are missing.

The macrolanguages sometimes don't default to the more common language because of legacy reasons. For instance in this case, if we release this change and someone is already using the `ku` locale in their app, it will suddenly change after updating Carbon, which a quite hard breaking change.

Note than `ckb` locale should be the Sorani language you're looking for.

AlanD20:
Hello, thank you for the clarification.

You are absolutely right. I was looking for the `ckb` locale. I wasn't aware of the existence of CLDR Unicode and I was mostly get used to the more general locale where most packages use `ku` locale as primary Kurdish Sorani.

iIn my previous applications, I used `ku` locale for Carbon instances, and when I looked at them a few weeks ago. All the Kurdish Sorani are changed to the Turkish version, which was the reason I opened this PR.

I guess we should use `ckb` locale instead of `ku` according to CLDR Unicode for Kurdish Sorani.

Thank you for the valuable comment, and thank you for introducing me to CLDR Unicode.

You are free to close this PR if you don't have anything to add.

kylekatarnls:
As a side note if you need to replace la Carbon locale with another in your application for consistency reason you can do it with:

```php
$translator = \Carbon\Translator::get('ku');
$translator->setTranslations(\Carbon\Translator::get('ckb')->getMessages('ckb'));

$date1 = Carbon::create(2018, 1, 1, 0, 0, 0);
$date2 = Carbon::create(2018, 1, 4, 4, 0, 0);

echo $date1->locale('ku')->diffForHumans($date2);
```

Output:
```
پێش 3 ڕۆژ
```

PR #2771, opened 2023-02-28
Lowercase swedish translations

tomsommer:


kylekatarnls:
Hello, as translation for the `calendar` array come from various sources, the casing is not consistent across languages. In version 4, it should be harmonized (so probably everything lowercased everywhere). But in the meantime, I think it's better not to change case for specific language from now on on minor updates, so not to create unexpected change for those who might currently use it and expect uppercase (if they use it as a title or the beginning of a sentence).

As a work-around, you can force case with:
```php
mb_ucfirst($date->calendar(...))
mb_lcfirst($date->calendar(...))
```

PR #2781, opened 2023-04-12
Options - `@method` shiftTimezone returns `static`

ondrejmirtes:
The original return type confuses PHPStan. All the methods in trait-using classes return `static` so it's correct to return `static` in `@method` PHPDoc too.

Thanks.

### DIFF
--- a/src/Carbon/Traits/Options.php
+++ b/src/Carbon/Traits/Options.php
@@ -22,7 +22,7 @@ use Throwable;
  *
  * Depends on the following methods:
  *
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
The build errors are not fault of this PR but are present on master already.

kylekatarnls (approved):
Indeed, it's how we do for `copy()`, `startOfWeek()` etc. so no reason to do it differently for `shiftTimezone()`

PR #2789, opened 2023-05-09
Update en_CH formats

Seeow:
In Switzerland, we will always use the `DD.MM.YYYY` format.

### DIFF
--- a/src/Carbon/Lang/en_CH.php
+++ b/src/Carbon/Lang/en_CH.php
@@ -11,4 +11,12 @@
 
 return array_replace_recursive(require __DIR__.'/en.php', [
     'first_day_of_week' => 1,
+    'formats' => [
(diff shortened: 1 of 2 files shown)


kylekatarnls (approved):

kylekatarnls:
When changing this one, I think we might have some inconsistency, for instance `h:mm A` is not used for time neither in Switzerland, it's typically `HH:mm`

It's a bit tricky for LL, LLL and LLLL as it will contain words in English, but for `LT` and `LTS` I guess we should `'HH:mm'`and `'HH:mm:ss'` since it's common to German, French and Italian.

Seeow:
Hello @kylekatarnls, you're right. I'll check the different formats and update the PR tomorrow.
Thanks.

kylekatarnls (approved):

PR #2791, opened 2023-05-12
FIX for:  A non-numeric value encountered at vendor/nesbot/carbon/src/Carbon/Traits/Units.php:136

raphaelcangucu:


raphaelcangucu (commented):
A non-numeric value encountered at vendor/nesbot/carbon/src/Carbon/Traits/Units.php:136

raphaelcangucu:
Hey sure:

I discovered that the problem is with laravel:

`    /**
     * Create a new notification instance.
     */
    public function __construct(Vip $vip)
    {
        $this->vip = $vip;
        
        **$this->delay($vip->cashback_limit_claim_date);**
        $this->onConnection('database');
    }
`

This delay function would broke the carbon parse ... Still don't know why.

But when I add the cast in the model, it worked.

` /**
     * The attributes that should be cast.
     *
     * @var array
     */
    protected $casts = [
        'rakeback' => 'decimal:8',
        'reload_base' => 'decimal:8',
        'reload_loss' => 'decimal:8',
        'rank_up_bonus' => 'decimal:8',
        'curated_rank_up_bonus' => 'decimal:8',
        'wagered_all_time' => 'decimal:8',
        'wagered_on_level' => 'decimal:8',
        'lost_on_level' => 'decimal:8',
        'profit_on_level' => 'decimal:8',
        'time_based_values' => 'array',
        'cashback_limit_claim_date' => 'datetime',
    ];`

But The code that I did, also fix the problem.

kylekatarnls:
Thank you. Can you please provide the full stack trace of your error and the version of Laravel that you use? Thanks.

kylekatarnls:
I miss infos to be able to reproduce your error.

But assuming `delay()` is from the Laraval trait `Queueable` it needs `\DateTimeInterface|\DateInterval|int|null` as an argument, please check which value has `$vip->cashback_limit_claim_date` when it fails (for instance if it's an empty string, `delay()` is not meant to support it and to fix it correctly, the safe way is to ensure `$vip->cashback_limit_claim_date` always has the expected type and/or `delay()` is not called if `$vip->cashback_limit_claim_date` is not a `\DateTimeInterface`).

If it happens when `$vip->cashback_limit_claim_date instanceof \DateTimeInterface`, then provide the value (and the `now` if you mock it) and the full stack trace so I can run the same `delay()` method with the same value to reproduce.

PR #2830, opened 2023-08-04
Fix PHPUnit 10 compatibility

rgson:
* Adds type hints to tests to deal with errors like:

> .PHP Fatal error:  Declaration of Carbon\Doctrine\CarbonTypeConverter::getSQLDeclaration(array $fieldDeclaration, Doctrine\DBAL\Platforms\AbstractPlatform $platform) must be compatible with Doctrine\DBAL\Types\DateTimeType::getSQLDeclaration(array $column, Doctrine\DBAL\Platforms\AbstractPlatform $platform): string in src/Carbon/Doctrine/CarbonTypeConverter.php on line 37

* Adds a workaround for a type mismatch in a trait shared between two classes in need of different return type signatures.

> .PHP Fatal error:  Declaration of Carbon\Doctrine\CarbonTypeConverter::convertToPHPValue(mixed $value, Doctrine\DBAL\Platforms\AbstractPlatform $platform): ?DateTimeInterface must be compatible with Doctrine\DBAL\Types\VarDateTimeImmutableType::convertToPHPValue(mixed $value, Doctrine\DBAL\Platforms\AbstractPlatform $platform): ?DateTimeImmutable in src/Carbon/Doctrine/CarbonTypeConverter.php on line 69

* Replaces the removed `assertObjectHasAttribute` method.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^0.12.99 || ^1.7.14",
         "phpunit/php-file-iterator": "^2.0.5 || ^3.0.6",
-        "phpunit/phpunit": "^7.5.20 || ^8.5.26 || ^9.5.20",
+        "phpunit/phpunit": "^7.5.20 || ^8.5.26 || ^9.5.20 || ^10.2.6",
         "squizlabs/php_codesniffer": "^3.4"
     },
     "provide": {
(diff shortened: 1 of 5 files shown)


kylekatarnls:
This would drop completely the PHP 7 compatibility. But [3.x](https://github.com/briannesbitt/Carbon/tree/3.x) version is compatible with PHPUnit 10.

PR #2848, opened 2023-09-11
Create dependabot.yml

juherr:


juherr:
Hi @kylekatarnls 
Any chance to have a review or feedback?

kylekatarnls:
Hello, sorry I didn't had time to check dependabot behavior, I would like to see what it would suggest as an example before I actually integrate it, what is the possible fine tunning. I would not add it if it to get suggestions that don't make sense for our library.

I would add new tools only if it has a benefit for the users. As a library, we try to keep our compatibility range wide as long as it's backward-compatible, which is already what happens using `^x.y.z` semver, dropping dependencies (major) versions happen only on major version bump of Carbon and then what version to drop or not is based on current support of those libraries, user needs, so it somehow has to be done manually. Same for supporting new major versions of the dependencies, we have very few and when we add support for a new version, it need testing and most of the time to adapt some code, not just to bump the number in composer.json. This also need human attention and manual care.

So I'm not sure relying on automation would help.

juherr:
Dependabot is just a friend who notify you that one of your dependencies has released a new version and help you to update without effort.

Here, it checks both composer dependencies and github actions.

In case you need more options https://github.com/renovatebot/renovate can be a good alternative.

In fact it is more a tool for the contributors than the end users. 😉

PR #2851, opened 2023-09-19
Use `mt_ ` random functions

lucasmichot:
Use `mt_ ` random functions

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -88,6 +88,7 @@ $rules = [
     'single_quote' => true,
     'space_after_semicolon' => true,
     'standardize_not_equals' => true,
+    'statement_indentation' => false,
     'ternary_operator_spaces' => true,
     'trailing_comma_in_multiline' => true,
     'trim_array_spaces' => true,
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -83,6 +83,7 @@ $rules = [
     'phpdoc_trim' => true,
     'phpdoc_types' => true,
     'phpdoc_var_without_name' => true,
+    'random_api_migration' => true,
     'self_accessor' => true,
     'single_quote' => true,
     'space_after_semicolon' => true,
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -5126,6 +5126,7 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
      * @param Closure(): T                                       $callback
      *
      * @return mixed
+     *
      * @phpstan-return T
      */
     public static function withTestNow($testNow, $callback);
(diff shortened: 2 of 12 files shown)


src/Carbon/Lang/fr.php
```diff
@@ -102,7 +102,7 @@
             case 'd':
                 return $number.($number === 1 ? 'er' : 'e');
 
-            // Words with feminine grammatical gender: semaine
+                // Words with feminine grammatical gender: semaine
```
kylekatarnls, line 105:
```suggestion
            // Words with feminine grammatical gender: semaine
```

src/Carbon/Traits/Date.php
```diff
@@ -913,102 +913,102 @@ public function get($name)
 
                 return is_numeric($value) ? (int) $value : $value;
 
-            // @property-read string long name of weekday translated according to Carbon locale, in english if no translation available for current language
+                // @property-read string long name of weekday translated according to Carbon locale, in english if no translation available for current language
```
kylekatarnls, line 916:
```suggestion
            // @property-read string long name of weekday translated according to Carbon locale, in english if no translation available for current language
```

src/Carbon/Traits/Date.php
```diff
@@ -913,102 +913,102 @@ public function get($name)
...
             case $name === 'dayName':
                 return $this->getTranslatedDayName();
-            // @property-read string short name of weekday translated according to Carbon locale, in english if no translation available for current language
+                // @property-read string short name of weekday translated according to Carbon locale, in english if no translation available for current language
```
kylekatarnls, line 919:
```suggestion
            // @property-read string short name of weekday translated according to Carbon locale, in english if no translation available for current language
```

src/Carbon/Traits/Date.php
```diff
@@ -913,102 +913,102 @@ public function get($name)
...
             case $name === 'shortDayName':
                 return $this->getTranslatedShortDayName();
-            // @property-read string very short name of weekday translated according to Carbon locale, in english if no translation available for current language
+                // @property-read string very short name of weekday translated according to Carbon locale, in english if no translation available for current language
```
kylekatarnls, line 922:
```suggestion
            // @property-read string very short name of weekday translated according to Carbon locale, in english if no translation available for current language
```

kylekatarnls (changes requested):
Thanks.

All random functions are in tests. So I would expect to have only test folder touched in this PR.

At least, I would not touch comment indent in switch witch is actually wrong here.

src/Carbon/Lang/fr.php
lucasmichot: Those are actually `php-cs-fixer` (`vendor/bin/php-cs-fixer fix`) fixes, if I revert those back, [the CS](https://github.com/briannesbitt/Carbon/blob/master/.github/workflows/phpcs.yml) workflow will fail.

In this case, we need to negate the `statement_indentation` rule

lucasmichot: cf. https://github.com/briannesbitt/Carbon/pull/2851/commits/3205687d0c7c6d1db5ce726942a8c7fe9c20ee0c

lucasmichot: I believe some changes might be related to an update of version of PHP-CS-Fixer

kylekatarnls (approved):

PR #2852, opened 2023-09-19
FIX: UnknownGetterException error on Serialization.php

### LINKED ISSUE #2850: UnknownGetterException on serialization
Hello,

I encountered an issue with the following code in Serialization.php file:

```php
    public function __serialize(): array
    {
        // @codeCoverageIgnoreStart
        if (isset($this->timezone_type)) {
            return [
                'date' => $this->date ?? null,
                'timezone_type' => $this->timezone_type,
                'timezone' => $this->timezone ?? null,
            ];
        }
    ...
```
Package Laravel: nesbot/Carbon v2.70.0
Carbon version: 2.70.0
PHP version: 8.1.23
Laravel version: 10.23.0

The 
```php
$this->timezone_type
```
and the
```php
$this->date
```
not work properly and they throws an UnknownGetterException in Date Trait get() method because only 
```php
$this->getTimezone()->getType();
```
returns a timezone_type value and only a 
```php
$this->format("Y-m-d H:i:s.u");
```
returns the full date value with microseconds.

Can you resolve this issue or tell me what i can do?

gjespo85:
$this->timezone and $this->date gets exception

Fix #2850

### DIFF
--- a/src/Carbon/Traits/Serialization.php
+++ b/src/Carbon/Traits/Serialization.php
@@ -145,16 +145,6 @@ trait Serialization
      */
     public function __serialize(): array
     {
-        // @codeCoverageIgnoreStart
-        if (isset($this->timezone_type)) {
-            return [
-                'date' => $this->date ?? null,
-                'timezone_type' => $this->timezone_type,
-                'timezone' => $this->timezone ?? null,
-            ];
-        }
-        // @codeCoverageIgnoreEnd
-
         $timezone = $this->getTimezone();
         $export = [
             'date' => $this->format('Y-m-d H:i:s.u'),
--- a/src/Carbon/Traits/Serialization.php
+++ b/src/Carbon/Traits/Serialization.php
@@ -145,6 +145,16 @@ trait Serialization
      */
     public function __serialize(): array
     {
+        // @codeCoverageIgnoreStart
+        if (isset($this->timezone_type) && isset($this->timezone) && isset($this->date)) {
+            return [
+                'date' => $this->date ?? null,
+                'timezone_type' => $this->timezone_type,
+                'timezone' => $this->timezone ?? null,
+            ];
+        }
+        // @codeCoverageIgnoreEnd
+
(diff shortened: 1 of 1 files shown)


kylekatarnls (changes requested):
Hello, tests need to pass on all versions for backward compatibility.

As you can see:

```php
$reflection = (new ReflectionClass(DateTime::class))->newInstanceWithoutConstructor();

$reflection->date = '1990-01-17 10:28:07';
$reflection->timezone_type = 3;
$reflection->timezone = 'US/Pacific';

$date = unserialize(serialize($reflection));
```

is no longer working in PHP >= 7.4 if we remove this part of the code.

Construction by reflection is needed to support some caching libraries.

Please instead add a test to reproduce your issue and we'll try to find the correct way to fix it without breaking features for other users.

gjespo85:
My issue happens when i dispatch a Job Queue on Laravel application. Maybe it can be resolved with a specific condition at row 149...

src/Carbon/Traits/Serialization.php
```diff
@@ -146,7 +146,7 @@ public function __sleep()
     public function __serialize(): array
     {
         // @codeCoverageIgnoreStart
-        if (isset($this->timezone_type)) {
+        if (isset($this->timezone_type) && isset($this->timezone) && isset($this->date)) {
```
kylekatarnls, line 149:
```suggestion
        if (isset($this->timezone_type, $this->timezone, $this->date)) {
```

kylekatarnls:
> My issue happens when i dispatch a Job Queue on Laravel application. Maybe it can be resolved with a specific condition at row 149...

Thanks. This code change is mergeable and it makes sense to check all properties, so if it solves your problem, I'll merge.

If you are able to give on #2850 a complete stack trace, we could spot which part of the Laravel code triggers this error, and if you can inspect to get an example of which data it fails to unserialize. So we could add it as a unit test and make sure it does not regress in the future.

PR #2856, opened 2023-09-22
isBetweenInclusiveExclusiveRange function

Axum25:
Adds a function that can test if a date is between two dates, where one date can be inclusive and the other date can be exclusive.

### DIFF
--- a/src/Carbon/Traits/Comparison.php
+++ b/src/Carbon/Traits/Comparison.php
@@ -440,7 +440,7 @@ trait Comparison
      * @example
      * ```
      * isBetweenInclusiveExclusiveRange(['2023-09-01', '2023-09-15'], (object)['left_inclusive' => true, 'right_inclusive' => false], '2023-09-15', null); // false
-     * isBetweenInclusiveExclusiveRange(['2023-09-01', '2023-09-15'], (object)['left_inclusive' => false, 'right_inclusive' => true], '2023-09-01',null,); // false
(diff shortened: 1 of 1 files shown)


kylekatarnls:
Hello, thanks for your contribution.

However, I don't think we should add this in the library:

I think calling:
`Carbon::isBetweenInclusiveExclusiveRange(['2023-09-01', '2023-09-15'], (object)['left_inclusive' => true, 'right_inclusive' => false], '2023-09-15', null);`

Is way less clear than simply calling one of the following syntax that all give the same result:
```php
$date = Carbon::parse('2023-09-15');

if ($date->greaterThanOrEqualTo('2023-09-01') && $date->lessThan('2023-09-15')) {}

if ($date >= Carbon::parse('2023-09-01') && $date < Carbon::parse('2023-09-15')) {}

// If strings are strictly formatted as YYYY-MM-DD, you can even do:
if ('2023-09-15' >= '2023-09-01') && '2023-09-15' < '2023-09-15') {}
```

Also `$dateBeingEvaluated` should not be a parameter, it's inconsistant with how method such as `->isBetween` are build as they can be called on a Carbon object, method such as `isSomething` should not be static methods and the subject of them should be `$this` (and for consistency, if the method is called statically, then it should use `now`).

But even that way, the `$bounds` config object will just create more complexity and be less obvious to use than simply 2 conditions using explictly "OrEqual" or not.

Same for the timezone, as each object can be either string or an object (that already has a timezone), it's really not obvious to predict how it should behave when having mixed timezone in the parameters, while on the opposite when the user explicitly creates 3 Carbon objects and compare them, they are in control of what is the timezone and can change it if needed before doing a comparison.

Axum25:
Thank you for your response and also breaking things down as to why this cannot work. Much appreciated.

PR #2857, opened 2023-09-23
Add `var_export()` support for `CarbonInterval`

hettiger:
`CarbonInterval` relies on the inherited `DateInterval::__set_state()` implementation when the interpreter evaluates PHP generated by `var_export()`. This yields an unexpected type change. The reconstructed instance is of type `DateInterval` instead of `CarbonInterval`. This behavior has been surprising for me and caused my code to break in an unexpected way.

I used `CarbonInterval::minutes(10)` in a Laravel config file. Locally I did not cache the config. Everything was working fine. Deploying the application also did not yield an error. However, the build steps included caching the config file. The application did crash at runtime when trying to call `CarbonInterval` exclusive methods on the config value.

Here's an example:

```php
// config/demo.php

return [
    'interval' ⇒ CarbonInterval::minutes(10),
];
```

```php
// routes/console.php

Artisan::command('demo', function () {
    $this->info(config('demo.interval')->greaterThan(now()));
});
```

Executing `php artisan demo` prints the string `'OK'` to the console. However, if you cache the config using `php artisan config:cache` executing `php artisan demo` begins to fail with `Call to undefined method DateInterval::greaterThan()`.

This PR fixes this problem. It makes sure, that `CarbonInterval` is properly reconstructed as a `CarbonInterval` instance.


tests/CarbonInterval/SetStateTest.php
```diff
@@ -0,0 +1,38 @@
...
+use Carbon\CarbonInterval;
+use Tests\AbstractTestCase;
+
+class SetStateTest extends AbstractTestCase
+{
+    public function testEvaluatingVarExportReturnsCarbonIntervalInstance()
+    {
+        $export = var_export(CarbonInterval::minutes(3), true);
+
+        $this->assertInstanceOfCarbonInterval(eval("return $export;"));
```
kylekatarnls, line 25: Maybe we should also check the rebuilt object is the correct interval.
```suggestion
        $export = var_export(CarbonInterval::minutes(3), true);
        $interval = eval("return $export;");

        $this->assertInstanceOfCarbonInterval($interval);
        $this->assertSame('3 minutes', $interval->forHumans());
```

kylekatarnls (approved):

src/Carbon/CarbonInterval.php
```diff
@@ -772,6 +772,20 @@ public static function __callStatic($method, $parameters)
...
+     * Evaluate the PHP generated by var_export() and recreate the exported CarbonInterval instance.
+     *
+     * @param array $an_array
+     * @return static
+     */
+    public static function __set_state($an_array): static
+    {
+        /** @noinspection PhpVoidFunctionResultUsedInspection */
+        /** @var DateInterval $dateInterval */
+        $dateInterval = parent::__set_state($an_array);
```
kylekatarnls, line 785:
```suggestion
     *
     * @param array $dump data as exported by var_export()
     */
    public static function __set_state($dump): static
    {
        /** @noinspection PhpVoidFunctionResultUsedInspection */
        /** @var DateInterval $dateInterval */
        $dateInterval = parent::__set_state($dump);

```

src/Carbon/CarbonInterval.php
```diff
@@ -772,6 +772,22 @@ public static function __callStatic($method, $parameters)
...
     }
 
+    /**
+     * Evaluate the PHP generated by var_export() and recreate the exported CarbonInterval instance.
+     *
+     * @param array $dump data as exported by var_export()
+     *
+     * @return static
+     */
+    public static function __set_state($dump)
```
kylekatarnls, line 782:
```suggestion
    #[ReturnTypeWillChange]
    public static function __set_state($dump)
```

Issue #2867, opened 2023-10-12
Fixed typo in PHPUnit test code.


websuke:
Hello,

Problems occurred at the following locations:
https://github.com/briannesbitt/Carbon/blob/ab853f01af3ef6f78d92df632490ad57579a3b0e/tests/PHPStan/FeaturesTest.php#L50

Carbon version: **3.0.0**

PHP version: **PHP 8.2.11**

I expected to get:

```
OK (2 tests, 2 assertions)
```

But I actually get:

```
There was 1 failure:

1) Tests\PHPStan\FeaturesTest::testAnalysesWithAnError
Failed asserting that ' ------ ------------------------------------------------------ \n
  Line   Fixture.php                                           \n
 ------ ------------------------------------------------------ \n
  :22    Static call to instance method Carbon\Carbon::foo().  \n
 ------ ------------------------------------------------------ \n
\n
\n
 [ERROR] Found 1 error                                                                                                  \n
\n
' contains "22     Static call to instance method Carbon\Carbon::foo().".

/home/.../Carbon/tests/PHPStan/FeaturesTest.php:50

FAILURES!
Tests: 2, Assertions: 2, Failures: 1.
```

 Thank you in advance.

websuke:
https://github.com/briannesbitt/Carbon/pull/2868

We will close it as well.

PR #2868, opened 2023-10-12
Fixed typo in PHPUnit test code. #2867

websuke:
Fixes were made so that the test would pass.
Thank you for your confirmation.


kylekatarnls:
Right now this test pass on both Windows and Linux on all PHP versions in GitHub Actions:
https://github.com/briannesbitt/Carbon/actions/runs/6486348590

(It's using PHPStan 1.10.38 as of now)

It also pass locally in my env.

So I would expect jobs to fail if we merge that.

You might have something different in your env.

websuke:
@kylekatarnls 
I have confirmed that it works on both my Mac and Ubuntu by modifying it, but the environment may be bad in both. We will close it. I apologize for the trouble.

PR #2871, opened 2023-10-19
Update the names of timezones

rgson:
Some timezones have had their names changed. The newer names are recommended in newer versions of PHP [1].

    Asia/Calcutta -> Asia/Kolkata
    Europe/Kiev   -> Europe/Kyiv
    US/Pacific    -> America/Los_Angeles

[1] https://www.php.net/manual/en/timezones.others.php

---

The old names cause test failures on systems where legacy time zone data is not installed.

### DIFF
--- a/tests/Carbon/SerializationTest.php
+++ b/tests/Carbon/SerializationTest.php
@@ -103,7 +103,7 @@ class SerializationTest extends AbstractTestCase
 
             @$reflection->date = '1990-01-17 10:28:07';
             @$reflection->timezone_type = 3;
-            @$reflection->timezone = 'US/Pacific';
+            @$reflection->timezone = 'America/Los_Angeles';
 
             $date = unserialize(serialize($reflection));
         } catch (Throwable $exception) {
@@ -118,7 +118,7 @@ class SerializationTest extends AbstractTestCase
 
         @$reflection->date = '1990-01-17 10:28:07';
         @$reflection->timezone_type = 3;
-        @$reflection->timezone = 'US/Pacific';
+        @$reflection->timezone = 'America/Los_Angeles';
 
         $date = unserialize(serialize($reflection));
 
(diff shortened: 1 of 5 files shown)


kylekatarnls:
IIRC, those tests aim to check a variety of timezones including deprecated ones. It also sounds like we don't have an updated timezone DB on Windows on GitHub Actions.

Maybe we can instead have a check of the timezone version of check which timezones are available and use a set of TZ based on what the OS supports.

May I ask what kind of annoyance is it? I.E. What is the need to run those unit tests, are you developing extensions for Carbon or something?

rgson:
> IIRC, those tests aim to check a variety of timezones including deprecated ones.

I see what you mean. The changes to the `testDateSerializationReflectionCompatibility` tests can be reverted. Only one instance of US/Pacific causes a test failure for me: the one testing `__wakeup`.

All mentions of Europe/Kiev and Asia/Calcutta seem (to me) to be unrelated to the tests' purposes.


> Maybe we can instead have a check of the timezone version of check which timezones are available and use a set of TZ based on what the OS supports.

In cases where the exact time zone doesn't matter, you could also consider using another time zone whose name has not changed. Possibly even one that practically means the same offset.


> May I ask what kind of annoyance is it? I.E. What is the need to run those unit tests, are you developing extensions for Carbon or something?

I maintain a package of this library for Debian (and Ubuntu, etc.). The tests run as part of that build process and Debian's routine QA processes.

The new test failures were noticed in Debian unstable [1], where legacy time zones are no longer installed by default.


In the end, my problem with this was resolved by simply installing the missing time zones.

This PR was more of a long-term solution as PHP's documentation also advises against using the legacy time zones. But considering the failures in GitHub Actions, it may be too early for you to conveniently switch then.


[1] https://bugs.debian.org/1052778

kylekatarnls:
We'll make those tests pass whatever the running system support old, new or both timezone names via #2886

PR #2873, opened 2023-10-24
new function minuteOfDay() added

bashbeeo:
Added a new function called minuteOfDay. This function represents how many minute passed for the day for a given Carbon instance.

For eg:

00:00:00 => 0th minute of day
00:04:00 => 4th minute of day
00:12:00 => 720th minute of day
23:59:00 => 1439th minute of day.

This function act as both getter and setter.

Getter
=====
Also works as property of the object Eg: 
```php
$date = Carbon::now();
// Both are equivalent
$date->minuteOfDay
$date->minuteOfDay()
```

Setter 
=====
```php
$date = Carbon::now(); // Say: 2023-10-23 14:23:00
$date->minuteOfDay(0) // $date becomes 2023-10-23 00:00:00
```

### DIFF
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -543,6 +543,7 @@ trait Date
     use Timestamp;
     use Units;
     use Week;
+    use Minute;
 
     /**
      * Names of days of the week.
@@ -1069,6 +1070,9 @@ trait Date
             case $name === 'locale':
                 return $this->getTranslatorLocale();
 
+            case $name === 'minuteOfDay':
+                return $this->minuteOfDay();
+
             default:
                 $macro = $this->getLocalMacro('get'.ucfirst($name));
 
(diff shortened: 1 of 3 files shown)


kylekatarnls:
Hello, thanks for the proposition, this is well written and tested 👌 Before I go for some deeper review, may I ask what is the use-case?

Mainly what makes the combo minute in day particular. I'm thinking that since we have `dayOfYear`, `dayOfWeek`, we can basically have any unit of any unit handled in a more generic way (`secondOfHour`, `hourOfWeek`).

I'm trying to evaluate if this is a one-of-a-kind need or we could end with a wider need which would make it better not to have a different implementation for each unit pair.

kylekatarnls:
This will be implemented in a generic way by #2885 via `<unit>Of<Unit>` and `<units>In<Unit>` dynamic getters/setters properties and methods.

PR #2883, opened 2023-11-17
Fix withTestNow

### LINKED ISSUE #2882: Shouldn't `Carbon::withTestNow()` restore original `$testNow` after invoking `$callback()`?
Hello,

```php
Carbon::setTestNow('2023-01-01 00:00:00');
Carbon::withTestNow('2023-01-02 00:00:00', fn () => doSomething());
echo Carbon::now()->format('Y-m-d H:i:s');
```

Carbon version:  2.71.0

PHP version: 8.2

I expected to get:

```
2023-01-01 00:00:00
```

But I actually get:

```
2023-11-17 18:09:00 (Real Time)
```

Thanks!

fuwasegu:
fix #2882

### DIFF
--- a/src/Carbon/Traits/Test.php
+++ b/src/Carbon/Traits/Test.php
@@ -128,12 +128,14 @@ trait Test
      */
     public static function withTestNow($testNow, $callback)
     {
+        $original = static::now();
+
         static::setTestNow($testNow);
 
         try {
             $result = $callback();
         } finally {
-            static::setTestNow();
+            static::setTestNow($original);
         }
 
         return $result;


src/Carbon/Traits/Test.php
```diff
@@ -128,12 +128,14 @@ public static function setTestNowAndTimezone($testNow = null, $tz = null)
      */
     public static function withTestNow($testNow, $callback)
     {
+        $original = static::now();
```
kylekatarnls, line 131: For the record, using `static::now()` here will return n instance even if it was `null` before. So if the time was running (no mock) before, then after `withTestNow()` you won't go back to "real time" running, you will go back to frozen time of the exact moment before executing `withTestNow()`.

Here instead, to get the behavior you want, the right method to call would be `static::getTestNow()`.

kylekatarnls:
We won't implement this for backward compatibility reason.

Also for advanced time-mocking tooling we have [carbonite](https://github.com/kylekatarnls/carbonite)

It has a `Carbonite::do()` which has the behavior you expect.

fuwasegu:
i see!
that's current.
i'm sorry but i seem i jumped to conclusions

PR #2888, opened 2023-11-22
Allow Symfony 7

evertharmeling:
Tried to make the library compatible with Symfony 7

- Followed the same setup using the `Lazy` implementation already used for the `Translator` class, to handle the incompatible return types and type hinted arguments.
- Not sure how to cope with the change of `Translator::setLocale` returning `void` (throwing `\Symfony\Component\Translation\Exception\InvalidArgumentException`) instead of `bool`. As I have the feeling it's just a BC break? It's getting more and more quirky to keep supporting older versions (of Symfony). Symfony `5.4` is the lowest LTS version atm, see https://symfony.com/releases
  - For now added skipping tests when the return type is `void`

Thoughts?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "psr/clock": "^1.0",
         "symfony/polyfill-mbstring": "^1.0",
         "symfony/polyfill-php80": "^1.16",
-        "symfony/translation": "^3.4 || ^4.0 || ^5.0 || ^6.0"
+        "symfony/translation": "^3.4 || ^4.0 || ^5.0 || ^6.0 || ^7.0"
     },
     "require-dev": {
         "doctrine/dbal": "^2.0 || ^3.1.4 || ^4.0",
--- a/lazy/Carbon/TranslatorImmutableStrongType.php
+++ b/lazy/Carbon/TranslatorImmutableStrongType.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * This file is part of the Carbon package.
+ *
+ * (c) Brian Nesbitt <brian@nesbot.com>
+ *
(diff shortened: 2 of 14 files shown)


kylekatarnls:
Symfony 7 will be supported on Carbon 3, you can test it with
```
composer require nesbot/carbon:^3.0@dev
```

evertharmeling:
Ah missed that, cool will do!

VincentLanglet:
> Symfony 7 will be supported on Carbon 3, you can test it with
> 
> ```
> composer require nesbot/carbon:^3.0@dev
> ```

Hi @kylekatarnls, can't it be supported on Carbon 2 without BC break ?
This would allow to make a release now with SF 7 support to allow other open source library working on SF 7 from now (also without having to bump a new major dependencies).

Or at least could a stable release of the v3 could be done ?
This will allow other libraries to not rely on a `@dev` version.

Thanks a lot

kylekatarnls:
Hello, version 2 is there since 2018, PHP 7.1, we're now at 2.71 minor version, and it's already a huge gap to fill to be compatible from Symfony 3.4 to 6, it's time for it to end.

Next step will be to provide adjustments in Laravel for them to be able to upgrade smoothly, then 3.0 will get a stable release.

kylekatarnls:
v3 compatibility with Laravel is now being checked: laravel/framework#49764

If you are on Laravel or have a composer constraint in your project needed carbon v2, you can still try the v3 using:

```
composer require "nesbot/carbon:3.x-dev as 2.99"
```

Do not hesitate to give it a try, your help in testing this new version will be gratefully welcome. The more it can be tested, the faster it will reach the stable release.

Issue #2268, opened 2021-01-29
ceilMonth on 29 January ceiling to March


abou-uthmaan:
Hello,

I encountered an issue with the following code:
```php
echo Carbon::parse('2021-01-29')->ceilMonth();
```

Carbon version: **PUT HERE YOUR CARBON VERSION (exact version, not the range)**
2.44

PHP version: **PUT HERE YOUR PHP VERSION**
7.3.13

I expected to get:

```
2020-02-01
```

But I actually get:

```
2020-03-01
```

https://try-carbon.herokuapp.com/ not working
The dates 29 til 31 of januari only not working.

Thanks!

kylekatarnls:
`ceilMonth()` basically rely on the following steps:

```php
echo (new DateTime('2021-01-31'))
    ->modify('next month')
    ->modify('first day of this month')
    ->format('Y-m-d');
```

And as you can see, this produces `2020-03-01` also using only pure PHP. Not sure yet how to deal with it as ceil/round/floor are meant to be the same for all and with an optional `$precision` parameter. It would be annoying to write a specific implementation for month with specifically `$precision = 0`

kylekatarnls:
Work-around added for 2.45.0

willrowe:
> ceilMonth() basically relies on the following steps:

> ```php
> echo (new DateTime('2021-01-31'))
>    ->modify('next month')
>    ->modify('first day of this month')
>    ->format('Y-m-d');
> ```

Would this work more reliably?
```php
echo (new DateTime('2021-01-29'))
    ->modify('last day of')
    ->modify('next day')
    ->format('Y-m-d');
```

kylekatarnls:
This was just an illustration. `ceilXyz($n)` is actually a magic method. There is nothing specific to "month". It can take any unit from millisecond to millenium, any of the floor/round/ceil method. And it has a precision parameter, so the same process applies to: rounding 10 minutes, flooring 0.5 day, ceiling 1 month. Obviously we won't customize all possible combinations.

Still this is fixed on `dev-master` version, to be released during the month.

PR #2890, opened 2023-11-27
fix(Rounding): apply rounding changes to result before applying unit

Sjoertjuh:
I was having a problem that looked a bit like #2268.

When rounding down to quarters, the month overflows when a date is after a possible date of the rounded month. This PR causes the changes to be applied before the main unit is changed, so the day is set before the month.

You can see an example in the tests. When rounding down months by a precision of 3, the 31st of may now becomes the 1st of april, instead of the 1st of may.

### DIFF
--- a/src/Carbon/Traits/Rounding.php
+++ b/src/Carbon/Traits/Rounding.php
@@ -130,11 +130,13 @@ trait Rounding
         $normalizedValue = floor($function(($value - $minimum) / $precision) * $precision + $minimum);
 
         /** @var CarbonInterface $result */
-        $result = $this->$normalizedUnit($normalizedValue);
+        $result = $this;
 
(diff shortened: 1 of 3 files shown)


kylekatarnls:
I will run unit tests but I suspect this will break other cases.

Sjoertjuh:
I was thinking the same, but everything still works as expected. Looking at the code coverage report, the re-run in case of overflow part isn't even needed anymore when doing this, so I'll create a commit to remove that part.

If you have any idea on how to better test this let me know.

kylekatarnls:
🤔 This should be fine, the code was changed to fix #2268 and a unit test ensure this fix didn't regress.

kylekatarnls (approved):
Made few more tests locally and checked 3.x branch. All good.

PR #2892, opened 2023-12-01
Compatibility with symfony/translation v7

gjuric:


kylekatarnls:
This is not backward compatible, to make this nreaking change, ne need to bump major version.

Symfony 7 will be supported on Carbon 3, you can test it with:
```
composer require nesbot/carbon:^3.0@dev
```

gsjurebus:
is there an estimate for the v3 release yet? best

nclavaud:
@kylekatarnls Please tell me if I am wrong, but it seems like there are two things in this pull request:
1. allow for symfony/translation v7
2. changing `Carbon\AbstractTranslator` and `Carbon\TranslatorImmutable`

I think the later (2) is a breaking change, but upgrading Symfony 7.0 should do no harm.

Reading the changelog from symfony/translation 7.0, I can't find anything impacting this project:
https://github.com/symfony/translation/blob/7.0/CHANGELOG.md#70

image

or did I miss something?

kylekatarnls:
> is there an estimate for the v3 release yet? best

I preliminary reached out Laravel team, because getting Laravel upgrading to Carbon v3 is essential in order not to have a lot of people staying on v2 which would double the maintenance burden for a long time. I'm confident this month we'll have an alpha release and open merge-request in Laravel to suggest the upgrade.

@nclavaud We cannot support Symfony 7 AND the current range: `^3.4 || ^4.0 || ^5.0 || ^6.0` it's conflicting. Supporting everything from 3.4 to 6 has already a high complexity cost. We reached acceptable maintenance burden limit, support for v7 will then come with dropping old versions. Dropping old versions will also restrict the PHP level supported and so doing it in a minor version, let's say 2.73, some users would be stuck on 2.72, some will continue on the stream ^2.73, and at the same time ^3 will start. This has also a high maintenance cost on the long term = slowing down evolution on the higher stream.

PR #2907, opened 2023-12-20
Cap to carbon-doctrine-types v2 by default

### LINKED ISSUE #2905: Unable to install DBAL on a fresh Laravel application due to type package
Hello,

Tim here from the Laravel core team 👋

```sh
composer create-project laravel/laravel dbal-test
cd dbal-test
composer require doctrine/dbal
```

Carbon version: 2.72.1

PHP version: 8.2.11

I expected to get:

```
A successful installation of DBAL
```

But I actually get:

```
Your requirements could not be resolved to an installable set of packages.
```

We are currently trying to work out how to resolve this for our users. I've opened a PR on Laravel to conflict with the `^3.0` version of types package (https://github.com/laravel/framework/pull/49438), however this causes existing application to no longer be able to update their version of the framework without running special composer commands.

Is there any way we can remove this package from the `"require"` and possibly add it to `"suggest"` or document that it should be installed if you are using Doctrine?

Thanks!

kylekatarnls:
We are investigating why `composer` is not automatically resolving this when using `composer require doctrine/dbal` on a fresh Laravel project.

Anyone needing carbon-doctrine-types v3 (for dbal v4) can use:
```
composer require "carbonphp/carbon-doctrine-types:3.1.0 as 2.99.99"
```
for the moment.

Fix #2905

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     "require": {
         "php": "^7.1.8 || ^8.0",
         "ext-json": "*",
-        "carbonphp/carbon-doctrine-types": "*",
+        "carbonphp/carbon-doctrine-types": "^1.0 || ^2.0",
         "psr/clock": "^1.0",
         "symfony/polyfill-mbstring": "^1.0",
         "symfony/polyfill-php80": "^1.16",


kylekatarnls:
Alternative fix:

Fixed by https://github.com/laravel/framework/pull/49438

PR #2941, opened 2024-02-09
Fix Persian translation for before and after

azim-kordpour:
Hi,
Persian is my mother tongue and these little changes make a meaningful translation in which developers use methods like diffForhumans. 
An example in English would be like this:
2 years back
2 years forward

Although they might be right, they are not common and meaningful.


kylekatarnls (approved):
Thank you. 🙏

mokhosh:
@azim-kordpour you seem to be misunderstood about the semantics of what you have changed. These are relative to another time, not "now", so the previous translation was correct, and you have made it incorrect.

mokhosh:
@kylekatarnls please revert this.

kylekatarnls:
Hello @mokhosh and @azim-kordpour, just to double check the semantic can you please translate parts in brackets here:

- `from_now` I have an appointment [in 2 hours] (from now)
- `ago` I had an appointment [2 hours ago] (in the past)
- `after` I have a first appointment at 12 and another one [2 hours after] (relative to the first one, can be in the past or future)
- `before` I have a first appointment at 12 and another one [2 hours before] (relative to the first one, can be in the past or future)

mokhosh:
- `from_now` would be دیگر in pure Farsi, and بعد if you don't mind a little Arabic in your Farsi.
- `ago` would be پیش in pure Farsi, and قبل if you don't mind a little Arabic.
- `after` depend on whether you mention the time that you're comparing it to or not. If you don't, it's just بعد. If you do, you need to add از and it becomes بعد از. So, [2 hours after] would be بعد, but [2 hours after the first appointment] would be بعد از.
- `before` has the same dynamics as `after`, but it's ‍قبل and قبل از.

So, it depends whether you use this Carbon feature, to mention both times that you're comparing, or you just mention one.

And this would be a breaking change for those who rely on the previous semantics.
For example if they have something like this but in Farsi:
```blade
This article has been edited {{ $updated_at->diffForHumans($created_at) }} it was created.
```

azim-kordpour:
@mokhosh You are right, it would be a breaking change in some cases.
Thank you for the explanation.

@kylekatarnls these commits should be reverted

PR #2955, opened 2024-02-29
add the isLeapDay method

kevinb1989:
I attempt to call this method today and it's not there. It would be nice to have.

### DIFF
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -2701,6 +2701,19 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
     public function isLeapYear(): bool;
 
     /**
+     * Determines if the instance is the 29 February of a leap year.
+     *
+     * @example
+     * ```
+     * Carbon::parse('2024-02-29')->isLeapDay(); // true
+     * Carbon::parse('2020-02-29')->isLeapDay(); // true
+     * Carbon::parse('2017-02-29')->isLeapDay(); // false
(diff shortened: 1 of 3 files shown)


kylekatarnls:
It could be just:
```php
if ($date->format('m-d') === '02-29') {
  echo 'This is a leap day';
}
```

Which makes me believe it does need a specific method.

kevinb1989:
@kylekatarnls I have refactored accordingly.

kylekatarnls:
But actually even `$this->isLeapYear()` is superfluous, you cannot have `->format('m-d') === '02-29'` in any other year than a leap year.

So it's already straightforward and needs only 1 short condition to check if current day is a leap day. "leap day" is a 1:1 synonym of "February 29", and I don't think we actually need an alias. 

My other concern on that is that the need in application code to actually handle February 29th in a particular way is first an edge-case, for most of the applications, it's just a day like any other. Second, the fact that something would behave any differently on February just because it's a leap day is a sign of a sub-optimal design, and thus an `if ($date->isLeapDay())` would be a code smell.

kylekatarnls:
Closing this as not a generic-enough need, anyone with particular behavior set on February 29th just because it's a leap day is encouraged to first challenge the rule and need for this condition, and double-check that it's not a hack that can be solved in a better way at design-level.

If the condition is still judged legit in the context application, a macro can be used to get `$date->isLeapDay()` (checking Carbon instance) and `Carbon::isLeapDay()` (checking Today) methods available using:
```php
Carbon::macro('isLeapDay', static function () {
  return self::this()->format('m-d') === '02-29';
});
```

PR #2958, opened 2024-03-11
Add diffCalendarMonths to difference, date trait

JackedUpProgrammer:
**The diffCalendarMonths() method is used when checking the difference in calendar months for two financial dates.**

```
$fromDate = Carbon::parse("2023-12-25");
$toDate = Carbon::parse("2024-02-10");

// this will use days to check for months difference
dd($toDate->diffInMonths($fromDate)); // = 1 month

// unlike the above method, we will use calendar months difference (from Dec until Jan and from Jan until Feb) 
// this method will be used more in the financial fields
dd(diffCalendarMonths($fromDate, $toDate)); // = 2 months
```

### DIFF
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -2904,7 +2904,7 @@ trait Date
         return $this->get($name);
     }
 
-    private function diffCalenderMonths(mixed $fromDate, mixed $toDate){
+    private function diffCalendarMonths(mixed $fromDate, mixed $toDate){
      
         if (empty($fromDate) || empty($toDate)){
             return null;
--- a/src/Carbon/Traits/Date.php
+++ b/src/Carbon/Traits/Date.php
@@ -2903,4 +2903,13 @@ trait Date
 
         return $this->get($name);
     }
+
+    private function diffCalenderMonths(mixed $fromDate, mixed $toDate){
+     
+        if (empty($fromDate) || empty($toDate)){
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Hello, 👋

I would personally prefer to explicitly write:
```php
$fromDate = Carbon::parse("2023-12-25");
$toDate = Carbon::parse("2024-02-10");

dd($toDate->startOfMonth()->diffInMonths($fromDate->startOfMonth()));
```

Rather than having a `diffCalendarMonths` method which in my opinion has a name a bit ambiguous about what it does.

Also the need to get this value is rather specific to few business cases only.

Thus I would not see a high value in adding such method in Carbon itself.

On another note, we don't merge merge-requests unless they provide public API and include unit tests for it.

As is, private methods are not providing any feature to the library.

Last the Carbon convention is that such diff methods are using callable on Carbon objects and use this object as reference point, for instance you can create your method as a macro:

```php
Carbon::macro('diffInCalendarMonths', static function ($other) {
    $date = self::this();
    return $date->startOfMonth()->diffInMonths($date->resolveCarbon($other)->startOfMonth()));
});
```

And then you can call it in your code with:
```php
Carbon::parse("2023-12-25")->diffInCalendarMonths(Carbon::parse("2024-02-10"));
```
Or event:
```php
Carbon::parse("2023-12-25")->diffInCalendarMonths("2024-02-10");
```

PR #2969, opened 2024-03-15
[PHP 8.4] Fixes for implicit nullability deprecation

Ayesh:
Fixes all issues that emit deprecation notices on PHP 8.4 for implicit nullable parameter type declarations.

See:
 - [RFC](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)
 - [PHP 8.4: Implicitly nullable parameter declarations deprecated](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)


kylekatarnls (approved):
Thanks :pray:

PR #2999, opened 2024-04-15
Fix: Allow carbon instance to be rounded by a CarbonInterval, which is not in default language

kohlerdominik:
Hello there!

## Bug description
In [Carbon\Traits\IntervalRounding::45](https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Traits/IntervalRounding.php#L45), we convert a CarbonInterval to a string, regardless of the language in which it was given. This fails, if:

- the user passes a CarbonInterval in a specific locale
- the user passes an Object, that is converted to a CarbonInterval, while the CarbonLocale is set to a language different than english.

## Suggested fix
As we are parsing english words, we should ensure in the library, that the passed `unit` is english, regardless of what the user does.

Therefore i suggest always to create a new CarbonInterval, set it's locale to default and parse the output of this object.

### DIFF
--- a/src/Carbon/Traits/IntervalRounding.php
+++ b/src/Carbon/Traits/IntervalRounding.php
@@ -42,7 +42,7 @@ trait IntervalRounding
         $unit = 'second';
 
         if ($precision instanceof DateInterval) {
-            $precision = (string) CarbonInterval::instance($precision, [], true);
+            $precision = (string) CarbonInterval::instance($precision)->locale('');
         }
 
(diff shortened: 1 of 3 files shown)


kohlerdominik:
Fixed and forcepushed non-camelCase variables.

I'm insecure what to do about the coverage change in [CarbonInterval](https://app.codecov.io/gh/briannesbitt/Carbon/pull/2999/indirect-changes). I added a new test for it in [7c52966](https://github.com/briannesbitt/Carbon/pull/2999/commits/7c529664f7bf2df860fc6c3258ada1e596e2f706).

kylekatarnls (approved):

PR #3022, opened 2024-05-24
Remove overridden EXCLUDE_START_DATE constant

iluuu1994:
This constant is has the same value as the one of the base class. Overriding this constant breaks in PHP 8.4 due to the addition of the constant type.

See https://github.com/php/php-src/pull/12361 for the discussion. We might end up reverting this, but it still doesn't seem like this constant is needed.

I wasn't sure which branch to target. Please let me know which one is correct and I can rebase.


kylekatarnls:
It's a relic from v2 where `CarbonPeriod` was not extending `DatePeriod`. It should have been removed at 3.0.0, my mistake.

iluuu1994:
@kylekatarnls Great, thank you for the quick merge! Will this land in the 3.x branch then?

PR #3030, opened 2024-05-31
Fix New Zealand daylight saving time format to pass hasFormat

Luoti:
I noticed that `hasFormat` does not pass if we are using for example [New Zealand's daylight saving time](https://en.wikipedia.org/wiki/Time_in_New_Zealand) that has time zone of +13:00.

I made this PR against 2.0 branch. I'm not sure if you accepts PR's anymore against it. I can create rebased PR to 3.0 branch as well if desired.


kylekatarnls (approved):

kylekatarnls:
Thank you 🙏

PR #3031, opened 2024-05-31
Fix New Zealand daylight saving time format to pass hasFormat v3

Luoti:
This is the V3 version of https://github.com/briannesbitt/Carbon/pull/3030


kylekatarnls:
Thank you 🙏

PR #3071, opened 2024-10-03
Optimize `setUnitNoOverflow()`

takaram:
This PR reduces `avoidMutation()` calls.


kylekatarnls:
Sounds good 👍 I'm doing few more tests to check we don't have any mutation side effect for neither `Carbon` nor `CarbonImmutable`

PR #3140, opened 2025-01-31
feat: add phpstan @param-closure-this tags to macro methods

calebdw:
Hello!

This adds the `@param-closure-this` to the macro methods so that PHPStan properly understands the type of `$this` in the closures.

Should fix this: https://github.com/larastan/larastan/discussions/2189

Thanks!

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -31,14 +31,6 @@ parameters:
         - '#^Property Carbon\\Carbon::\$timezone \(Carbon\\CarbonTimeZone\) does not accept string\.$#'
         - '#^Method class@anonymous/tests/Carbon/TestingAidsTest\.php:\d+::modify\(\) should return class@anonymous/tests/Carbon/TestingAidsTest\.php:\d+ but returns \(?DateTimeImmutable#'
         -
-            message: '#^Undefined variable: \$this$#'
-            paths:
-                - src/Carbon/Traits/Mixin.php
-        -
-            message: '#^Variable \$this in isset\(\) is never defined\.$#'
-            paths:
-                - src/Carbon/Traits/Mixin.php
-        -
             message: '#^Call to an undefined method Carbon\\Carbon::[a-zA-Z]+Of[a-zA-Z]+\(\)\.$#'
             paths:
                 - tests/Carbon/SettersTest.php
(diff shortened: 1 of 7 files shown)


szepeviktor:
📜 https://phpstan.org/writing-php-code/phpdocs-basics#callables

phpstan.neon
```diff
@@ -30,14 +30,6 @@ parameters:
...
         - '#^Property Carbon\\Carbon::\$timezone \(Carbon\\CarbonTimeZone\) does not accept string\.$#'
         - '#^Method class@anonymous/tests/Carbon/TestingAidsTest\.php:\d+::modify\(\) should return class@anonymous/tests/Carbon/TestingAidsTest\.php:\d+ but returns \(?DateTimeImmutable#'
-        -
-            message: '#^Undefined variable: \$this$#'
-            paths:
-                - src/Carbon/Traits/Mixin.php
-        -
-            message: '#^Variable \$this in isset\(\) is never defined\.$#'
-            paths:
-                - src/Carbon/Traits/Mixin.php
```
szepeviktor, line 40: This is always good.

kylekatarnls: Depends, if it disapear by hacking $this into ${'this'} then I would say no, because it's just fooling the tool and adding complexity/execution time while not actually changing what Psalm is complaining about.

kylekatarnls (changes requested):
Thanks for this, I think it's very good to add  this annotation. You'll find a question to check in the code diff.

src/Carbon/Traits/Mixin.php
```diff
@@ -101,8 +101,8 @@ private static function loadMixinTrait(string $trait): void
             $closureBase = Closure::fromCallable([$context, $name]);
 
             static::macro($name, function (...$parameters) use ($closureBase, $className, $baseClass) {
-                $downContext = isset($this) ? ($this) : new $baseClass();
-                $context = isset($this) ? $this->cast($className) : new $className();
+                $downContext = isset(${'this'}) ? $this : new $baseClass();
+                $context = isset(${'this'}) ? $this->cast($className) : new $className();
```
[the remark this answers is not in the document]
calebdw: That's fair enough, I'll adjust. Because the `$this` variable might not be defined the type technically needs to be `static|null` but it seems like PHPStan doesn't understand that:

https://github.com/phpstan/phpstan/issues/12511

kylekatarnls:
Fixed via #3174 (which I preferred because it's not changing the actual source code)

calebdw:
@kylekatarnls, #3174 is an incomplete fix---it only added the `@param-closure-this` to two locations whereas it's needed in 5 locations

PR #3145, opened 2025-02-10
After updating Laravel in the Filament project, an error occurs when uploading an image.

iamafroz:
## Issue Description
- After updating Laravel version 9 to 11 in the Filament project, an error occurs when uploading an image. The error message is this: `Carbon\Carbon::rawAddUnit(): Argument #3 ($value) must be of type int|float, string given, called in /Users/afroz/Desktop/Projects/official-website-backend/vendor/nesbot/carbon/src/Carbon/Traits/Units.php on line 356.`
- i Fixed that error
- screenshot
image

### DIFF
--- a/src/Carbon/Traits/Units.php
+++ b/src/Carbon/Traits/Units.php
@@ -353,7 +353,7 @@ trait Units
         $previousException = null;
 
         try {
-            $date = self::rawAddUnit($date, $unit, $value);
+            $date = self::rawAddUnit($date, $unit, intval($value));
 
             if (isset($timeString)) {
                 $date = $date?->setTimeFromTimeString($timeString);


kylekatarnls (changes requested):
Hello,

A likely better fix would be to fix the type of the value that is actually passed (by the project, Filament or Laravel framework or a third-party they use).

For instance, if there is a `$date->addDays("6")` to replace it with `$date->addDays(6)` (or `$date->addDays((int) $days)` if it's a variable).

The reason not to do `(int)` or `intval()` is to avoid silent bugs. When you do `$date->addDays(6.99)` there is a big chances `6` (i.e. `intval(6.99)` is not what you want. When you do `$date->addDays("foobar")` there is a high chance you have an issue to fix in your project (wrong variable passed, or missing validation), in such as you wouldn't be fine with `$date->addDays(0)` (what would happen with `intval()`) you actually want to throw something so you can be aware of the problem and fix it.

-------

If the above does not apply then we need to have a test that shows a method call that Carbon should legitimately support and throw this error.

PR #3174, opened 2025-04-05
Fix Incorrect Type Inference when Declaring Macros

### LINKED ISSUE #3173: Incorrect Type Inference when Declaring Macros
Hey - hope you are well!

When declaring a macro `$this` (in the context of the callable) is inferred as being the class in which the macro is being created, rather than the class on which the macro is being created.

[A minimal reproduction](https://phpstan.org/r/7b52fac9-4df8-4de7-9412-4542ad44807b)

liamduckett:
Hey, resolves #3173 using [param-closure-this](https://phpstan.org/writing-php-code/phpdocs-basics#callables).

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -35,6 +35,10 @@ parameters:
             paths:
                 - src/Carbon/Traits/Mixin.php
         -
+            message: '#^Variable \$this in isset\(\) always exists and is not nullable\.$#'
+            paths:
+                - src/Carbon/Traits/Mixin.php
+        -
             message: '#^Variable \$this in isset\(\) is never defined\.$#'
             paths:
                 - src/Carbon/Traits/Mixin.php
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -31,15 +31,7 @@ parameters:
         - '#^Property Carbon\\Carbon::\$timezone \(Carbon\\CarbonTimeZone\) does not accept string\.$#'
         - '#^Method class@anonymous/tests/Carbon/TestingAidsTest\.php:\d+::modify\(\) should return class@anonymous/tests/Carbon/TestingAidsTest\.php:\d+ but returns \(?DateTimeImmutable#'
         -
-            message: '#^Undefined variable: \$this$#'
-            paths:
-                - src/Carbon/Traits/Mixin.php
-        -
             message: '#^Variable \$this in isset\(\) always exists and is not nullable\.$#'
-            paths:
-                - src/Carbon/Traits/Mixin.php
-        -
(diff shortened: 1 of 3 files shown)


kylekatarnls (approved):

liamduckett:
I had to alter the `ignoreErrors` entries for the Mixin trait, to accomodate this change. It doesn't get PHPStan issues to 0 - but does get them back to where they were before this PR.

I needed to swap:

```
'#^Undefined variable: \$this$#'
'#^Variable \$this in isset\(\) is never defined\.$#'
```

for

```
'#^Variable \$this in isset\(\) always exists and is not nullable\.$#'
```

This is because this `(Macro|CarbonInterval)::macro` method now believes `$this` is `static`. In reality this isn't actually true: given the rather unsual context in which this sits, `$this` could actually be `static` or `undefined` (for reference, it previously would have been though of as just `undefined`). I still believe this is the best path forward (despite having to ignore an error), as the situation in which `$this` is `undefined` is not exposed to userland code - and this change does resolve my original issue.

I think to resolve this properly (as in, without needing to ignore an error) - the logic that handles "whether we're in a static context" (see below) would need to rejigged. I think this would be out of the scope of this PR?

```php
private static function loadMixinTrait(string $trait): void
{
    $context = eval(self::getAnonymousClassCodeForTrait($trait));
    $className = \get_class($context);
    $baseClass = static::class;

    foreach (self::getMixableMethods($context) as $name) {
        $closureBase = Closure::fromCallable([$context, $name]);

        static::macro($name, function (...$parameters) use ($closureBase, $className, $baseClass) {
            // This is the "whether we're in a static context" I'm referring to
            $downContext = isset($this) ? ($this) : new $baseClass();
            $context = isset($this) ? $this->cast($className) : new $className();
    // ...
```

I only added `@param-closure-this` to the methods that were required to satisfy PHPStan. In reality I think this should also be added to:
1. `src/Carbon/Factory::macro`
2. `src/Carbon/CarbonInterface::macro`
3. `src/Carbon/CarbonPeriod::macro`

However I've left them out to keep the PR as slim as possible (feels slight pointless after writing this wall of text 😅) - happy to be directed otherwise.

Thanks!

kylekatarnls:
Only the error ignore is to be removed from the phpstan config, the rest is fine.

liamduckett:
I think the Laravel test failures are due to [this](https://github.com/laravel/framework/pull/55297).

liamduckett:
I've resolved that PHPStan issue. It occurs (for me) on `master` (with none of my changes applied) - which is why I didn't do anything about it initially.

 I can't immediately see how it could be related to my change?

kylekatarnls (approved):

PR #3179, opened 2025-04-19
feat: Add tryParse method

DarronEngelbrecht:
This PR introduces a tryParse method.

Previously...
```php
try { 
    $date = Carbon::parse('invalid-input');
} catch (InvalidFormatException) { 
    $date = null;
}
```

Now...
```php
$date = Carbon::tryParse('invalid-input');
```

### DIFF
--- a/src/Carbon/CarbonInterface.php
+++ b/src/Carbon/CarbonInterface.php
@@ -3607,6 +3607,15 @@ interface CarbonInterface extends DateTimeInterface, JsonSerializable
     public static function parse(DateTimeInterface|WeekDay|Month|string|int|float|null $time, DateTimeZone|string|int|null $timezone = null): static;
 
     /**
+     * Try create a carbon instance from a string.
+     *
+     * This wraps the regular parse method in a try catch statement and returns null if an exception is caught.
+     *
+     * @return static|null
+     */
+    public static function tryParse(DateTimeInterface|WeekDay|Month|string|int|float|null $time, DateTimeZone|string|int|null $timezone = null): static|null;
+
+    /**
      * Create a carbon instance from a localized string (in French, Japanese, Arabic, etc.).
      *
      * @param string                       $time     date/time string in the given language (may also contain English).
--- a/src/Carbon/Traits/Creator.php
+++ b/src/Carbon/Traits/Creator.php
@@ -222,6 +222,26 @@ trait Creator
     }
 
     /**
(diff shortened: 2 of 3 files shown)


kylekatarnls:
Hello, this is not a path I'm confident to take. Throwing exception is a safer way to work with invalid values. And if we do that for `parse` we should do it for dozens of methods.

On user-side you can create a generic helper for that:

```php
/**
 * @template T
 * @param callable():T $callback
 * @return T|null
 */
function tryOrNull(callable $callback): mixed
{
  try { 
    return $callback();
  } catch (Throwable) { 
    return null;
  }
}

$date = tryOrNull(fn () => Carbon::tryParse('invalid-input'));
```

It's more explicit and you would be able to re-use it on anything, not just Carbon:
```php
$date = tryOrNull($someObject->someMethod(...));
```

kylekatarnls:
You can also have a whitelist of exception classes to be more precise:

```php
/**
 * @template T
 * @param callable():T $callback
 * @return T|null
 */
function tryOrNull(callable $callback, array $mutedErrors = []): mixed
{
  try { 
    return $callback();
  } catch (Throwable $e) {
    if ($mutedErrors === []) {
      return null;
    }

    foreach ($mutedErrors as $class) {
      if (is_a($e, $class, true)) {
        return null;
      }
    }

    throw $e;
  }
}
```

DarronEngelbrecht:
@kylekatarnls - Thanks for the input. I guess this can be implemented in user-land using Carbon Macros to create a `tryParse` method as well. I understand the desire to show up exceptions and to implement this on other parse methods to keep some standardization but will leave this open for further input for now. 

Time permitting I could try to implement a proposal for fluent syntax on other methods if there is some demand for it? Something like the below? 

````php
Carbon::try()->parse()
Carbon::try()->createFromFormat()
etc. 
```

kylekatarnls:
I'm closing for now.

If anyone is looking for a convenient but clean way to work such exceptions, I recommend the [texthtml/maybe](https://github.com/texthtml/maybe) package:

```php
<?php

use function TH\Maybe\Option\tryOf;

$value = tryOf(fn () => Carbon::tryParse($dateString));
// or tryOf(fn () => Carbon::tryParse($dateString), exceptionClass: InvalidFormatException::class);
// to specify the exception

$date = $value->unwrapOr(null); // return Carbon instance of null if $dateString is invalid
```

PR #3190, opened 2025-05-06
[2.x] Fix float and real month diff inaccuracy

### LINKED ISSUE #2604: diffInMonths inconsistent, adding a day makes the difference bigger instead of smaller
Hello,

```php
$start = new Carbon("2022-11-11 22:29:50.000000");
$end = $start->copy()->addMonths(3);
$now = $start->copy()->addDay();

echo "diffInMonths\n";
echo $start->diffInMonths($end) . "\n";
echo $now->diffInMonths($end) . "\n";

echo "\nfloatDiffInMonths\n";
echo $start->floatDiffInMonths($end) . "\n";
echo $now->floatDiffInMonths($end) . "\n";

echo "\nfloatDiffInRealMonths\n";
echo $start->floatDiffInRealMonths($end) . "\n";
echo $now->floatDiffInRealMonths($end) . "\n";
```

Carbon version: 2.x/3.x

PHP version: 7.4.28 / Whatever https://try-carbon.herokuapp.com/ uses

I expected to get:

```
diffInMonths
3
2

floatDiffInMonths
3
2.9<something>

floatDiffInRealMonths
3
2.9<something>
```

But I actually get:

```
diffInMonths
3
2

floatDiffInMonths
3
3.0055439548131

floatDiffInRealMonths
3
3.0055439548131
```

Thanks!

I have seen and read #2264 

The "counting every day by fraction" way from https://github.com/briannesbitt/Carbon/issues/2264#issuecomment-774691425 results in:
```
>>> CarbonHelper::monthDifferenceDayByDay($start, $end)
=> 3.0595238095238
>>> CarbonHelper::monthDifferenceDayByDay($now, $end)
=> 3.0261904761905
```
Which is better - at least adding a day makes the difference smaller now, but I'd still expect 3 and < 3 as the results

Kofful:
Fixes #2604 for 2.x.

Backported from 3.x (Origin: 37b3bd5f341da63f93aa0abd3ad6a95c14ab9041)

### DIFF
--- a/src/Carbon/Traits/Difference.php
+++ b/src/Carbon/Traits/Difference.php
@@ -573,14 +573,13 @@ trait Difference
             return $sign * $monthsDiff;
         }
 
-        /** @var Carbon|CarbonImmutable $startOfMonthAfterFloorEnd */
-        $startOfMonthAfterFloorEnd = $floorEnd->avoidMutation()->addMonth()->startOfMonth();
+        /** @var Carbon|CarbonImmutable $ceilEnd */
+        $ceilEnd = $start->avoidMutation()->addMonths($monthsDiff + 1);
 
-        if ($startOfMonthAfterFloorEnd > $end) {
-            return $sign * ($monthsDiff + $floorEnd->floatDiffInDays($end) / $floorEnd->daysInMonth);
-        }
+        $daysToFloor = $floorEnd->diffInDays($end);
+        $daysToCeil = $end->diffInDays($ceilEnd);
 
-        return $sign * ($monthsDiff + $floorEnd->floatDiffInDays($startOfMonthAfterFloorEnd) / $floorEnd->daysInMonth + $startOfMonthAfterFloorEnd->floatDiffInDays($end) / $end->daysInMonth);
+        return $sign * ($monthsDiff + $daysToFloor / ($daysToCeil + $daysToFloor));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kylekatarnls:
Hello, the version number get bumped to next major version mainly to be able to introduce breaking changes in the behavior of diff methods.

There are still many people on 2.x and many might rely on the current logic.

Basically if we would allow breaking changes and follow the approach of back-porting stuff from 3.x we could just merge everything, it would break badly for many existing applications, and there would be no point in following semantic versioning.

Instead of modifying 2.x (which by now should only receive security patches), I would encourage whoever stuck in such version to chase for the blockers (could it be some incompatible interface in the app, or some third-party requiring specifically 2.x), then to upgrade/replace those blockers, submit PRs on their repositories for them to support Carbon v3, and then you could upgrade to the latest version and benefit from the active support on this version.

Kofful:
Thanks @kylekatarnls for clarification.

The thing that made me want to backport this change was that I've encountered a bug with similar behavior as in #2604.
The following code:
```php
use Carbon\Carbon;

$start           = new Carbon('2025-02-18');
$end             = new Carbon('2026-02-17');
$floatMonthsDiff = $start->floatDiffInMonths($end);
$monthsDiff      = $start->diffInMonths($end);

echo "$floatMonthsDiff float months\n";
echo "$monthsDiff months\n";

```
Results in:
```bash
12.023041474654 float months
11 months
```
So for me this difference was critical enough to create a PR here. Since it's a legacy project with Laravel 7 in the dependencies, I couldn't just bump `nesbot/carbon` to v3 and I though it may be useful for such projects as Laravel requires Carbon v3 [only after](https://packagist.org/packages/laravel/framework#v11.0.0) Laravel 11.

For me the workaround with just calculating months diff by ourselves was easier than upgrading through 4 versions of Laravel.

Anyway, happy to close the PR if it's not going to be merged into v2. And thank you for the fast response.

PR #3213, opened 2025-07-16
Change ?self return type to ?static in Creator Trait

Disservin:
All of the returned instances are created of static::, so I think it only makes sense to document this as `?static` and not `?self`. Also matches the manual docstring.


kylekatarnls:
Thanks, it's a relic of PHP 7 compatibility

PR #3218, opened 2025-07-30
Fix 0 second diff edge case in translations

Joeb454:
From what I can see, most of these translate as something like ‘a few moments ago’. I think handling the edge case of a 0-second diff this way is ok, but I’m open to suggestions on this.


kylekatarnls:
> most of these translate as something like ‘a few moments ago’

This initially come from moment.js from where some translations were imported.

We still have a mix of language with units and some with this form. Ideally user should opt-in this behavior with the option `aUnit` ```php
echo Carbon::now()->diffForHumans(['aUnit' => true]);
```
And by default we should keep the numbers.

So I updated those file to make them support this option.

PR #3257, opened 2025-11-07
[Translation-bn] Align Bangla locale files with English reference strings

imhayatunnabi:
## Description

- filled in the remaining relative time units (millisecond/microsecond, short codes, and ordinal helper) in `src/Carbon/Lang/bn.php` so the Bangla base locale matches the English reference
- localized the recently added relative-day phrases (`diff_before_yesterday`, `diff_after_tomorrow`) to ensure all helpers render in Bangla
- verified that the derived locales (`bn_IN`, `bn_BD`) inherit the new keys and no longer miss entries compared to `en.php`

### DIFF
--- a/src/Carbon/Lang/bn.php
+++ b/src/Carbon/Lang/bn.php
@@ -115,7 +115,7 @@ return [
 
         // Special cases for teens (11-19) always use তম
         if ($lastTwoDigits >= 11 && $lastTwoDigits <= 19) {
-            return $bengaliNumber . 'তম';
+            return $bengaliNumber.'তম';
         }
 
         // For numbers 1-10, use specific rules
@@ -127,21 +127,21 @@ return [
                 case 8:
                 case 9:
                 case 10:
-                    return $bengaliNumber . 'ম';
+                    return $bengaliNumber.'ম';
                 case 2:
                 case 3:
-                    return $bengaliNumber . 'য়';
+                    return $bengaliNumber.'য়';
                 case 4:
-                    return $bengaliNumber . 'র্থ';
+                    return $bengaliNumber.'র্থ';
                 case 6:
-                    return $bengaliNumber . 'ষ্ঠ';
+                    return $bengaliNumber.'ষ্ঠ';
                 default:
-                    return $bengaliNumber . 'তম';
+                    return $bengaliNumber.'তম';
             }
         }
 
         // For numbers > 20, all use তম
-        return $bengaliNumber . 'তম';
+        return $bengaliNumber.'তম';
     },
     'list' => [', ', ' এবং '],
     'first_day_of_week' => 0,


src/Carbon/Lang/bn.php
```diff
@@ -90,6 +99,9 @@
     'weekdays' => ['রবিবার', 'সোমবার', 'মঙ্গলবার', 'বুধবার', 'বৃহস্পতিবার', 'শুক্রবার', 'শনিবার'],
     'weekdays_short' => ['রবি', 'সোম', 'মঙ্গল', 'বুধ', 'বৃহস্পতি', 'শুক্র', 'শনি'],
     'weekdays_min' => ['রবি', 'সোম', 'মঙ্গ', 'বুধ', 'বৃহঃ', 'শুক্র', 'শনি'],
+    'ordinal' => static function ($number) {
+        return $number;
+    },
```
kylekatarnls: Hello @imhayatunnabi we don't need this closure at all because it's the default behavior already to `return $number;`

However you previously added a version with `return $number.'তম';` which sounded correct.

Are you sure [it should be reverted](https://github.com/briannesbitt/Carbon/pull/3257/commits/066d51152b3986ac4dfc6d1ec5825b6e008eb4fb)?

If you reverted this because of the failed test, please note that tests for translation are not carved in the stone, if the translation is indeed better with `তম` then we should update the tests to reflect it.

When `December 1st` in English, `st` here is the ordinal. And so here is the ordinal function for it:
```php
'ordinal' => static function ($number) {
    $lastDigit = $number % 10;

    return $number.(
        ((int) ($number % 100 / 10) === 1) ? 'th' : (
            ($lastDigit === 1) ? 'st' : (
                ($lastDigit === 2) ? 'nd' : (
                    ($lastDigit === 3) ? 'rd' : 'th'
                )
            )
        )
    );
},
```

In some languages, not all the numbers can have ordinal and it may depends on the unit, for instance in French:
```php
'ordinal' => static function ($number, $period) {
    return match ($period) {
        // In French, only the first has to be ordinal, other number remains cardinal
        // @link https://fr.wikihow.com/%C3%A9crire-la-date-en-fran%C3%A7ais
        'D' => $number.($number === 1 ? 'er' : ''),
        default => $number.($number === 1 ? 'er' : 'e'),
        'w', 'W' => $number.($number === 1 ? 're' : 'e'),
    };
},
```

So basically the question is: how do you translate `December 1st` and `December 24th` in Bangla?

kylekatarnls:
Perfect then we should do that (and `$bengaliNumber` should be in sync with `alt_numbers`, with a variable at the beginning of the file for instance) and update the test to reflect it.

You can do the source change and I'll generate updates for the test file, then I would be glad if you could double-check the changes of the test files after that.

imhayatunnabi:
@kylekatarnls now it is full proof bengali. Thank you for your cooperation and review check. 
Thank you @kylekatarnls .

PR #3258, opened 2025-11-09
Fix DST (Daylight Saving Time) Transition Issues

imhayatunnabi:
**DST (Daylight Saving Time) Transition Issues**
- `create()` Method Silently Adjusts Invalid DST Times
- `createSafe()` Has DST Detection Bug
- Timezone Conversion During DST Transitions


imhayatunnabi:
Tests are running but coverage collection might have an issue. The helper class is new and may need to be included in coverage configuration. This is common for new files and will resolve once the PR is merged

Issue #3259, opened 2025-11-09
Infinite Loop Risk in CarbonPeriod with Empty Intervals


imhayatunnabi:
**Describe the bug**
I was building a calendar feature where users could view events in a date range. Everything worked fine until one day the server started crashing. After hours of debugging, I found the problem:

**To Reproduce**

```php
use Carbon\Carbon;
use Carbon\CarbonPeriod;
use Carbon\CarbonInterval;

// Create period with empty interval (can happen through reflection, 
// edge cases, or if validation is bypassed)
$period = CarbonPeriod::create(
    Carbon::parse('2024-01-01'),
    CarbonInterval::days(1), // Start with valid interval
    Carbon::parse('2024-01-31')
);

// Bypass validation and set empty interval (edge case scenario)
$reflection = new ReflectionClass($period);
$property = $reflection->getProperty('dateInterval');
$property->setAccessible(true);
$property->setValue($period, CarbonInterval::days(0)); // Empty interval!

// This loop runs FOREVER
foreach ($period as $date) {
    echo $date->format('Y-m-d') . "\n";
    // Loop never ends, server crashes
}
```

The loop kept running forever because adding 0 days means the date never changes, so it never reaches the end date. My server ran out of memory and crashed.

**Expected behavior**
When iterating a `CarbonPeriod` with an empty interval, it should:
1. Either throw an exception immediately
2. Or yield only the start date once and then stop iteration
3. Or detect the empty interval during iteration and stop gracefully

**Actual behavior**
The iteration continues infinitely:
- The date never changes (adding 0 days/hours/etc. doesn't advance)
- The loop never reaches the end condition
- Memory consumption grows until the server crashes

**Versions**
 - PHP: 8.2.29
 - Carbon: 3.x

PR #3260, opened 2025-11-10
Infinite Loop in CarbonPeriod with Empty Intervals

imhayatunnabi:
Github Issue: https://github.com/briannesbitt/Carbon/issues/3259
- Added checks in getIterator() and incrementCurrentDateUntilValid() to handle empty intervals without step functions, ensuring iteration stops immediately.
- Introduced EmptyIntervalTest to validate behavior of CarbonPeriod with empty intervals, confirming that only the start date is yielded and iteration halts as expected.

### DIFF
--- a/tests/CarbonPeriod/EmptyIntervalTest.php
+++ b/tests/CarbonPeriod/EmptyIntervalTest.php
@@ -137,7 +137,7 @@ class EmptyIntervalTest extends AbstractTestCase
         foreach ($emptyIntervals as $emptyInterval) {
             try {
                 $period->setDateInterval($emptyInterval);
-                $this->fail('Expected InvalidIntervalException for empty interval: ' . $emptyInterval->spec());
+                $this->fail('Expected InvalidIntervalException for empty interval: '.$emptyInterval->spec());
             } catch (InvalidIntervalException $e) {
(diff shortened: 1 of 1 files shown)


imhayatunnabi:
I am failure in codecov/patch.

src/Carbon/CarbonPeriod.php
```diff
@@ -327,6 +327,15 @@ class CarbonPeriod extends DatePeriodBase implements Countable, JsonSerializable
 
     public function getIterator(): Generator
     {
+        // Safety check: if interval is empty AND has no step function, yield only start date and stop
+        // Note: Empty intervals with step functions (callables) are valid and should continue
+        if ($this->dateInterval->isEmpty() && $this->dateInterval->getStep() === null) {
+            $this->rewind();
+            yield 0 => $this->current();
```
kylekatarnls: If we can demonstrate that an infinite loop is indeed possible (i.e. that it's possible to iterate more than `END_MAX_ATTEMPTS` times), then I don't think silently returning just 1 date is a correct way to handle this. It's putting the dirt under the carpet. I think it's way safer to throw an exception in such case so the user can discover that the interval is wrong and act upon it (not create such period then decide on their application what it should do instead).

PR #3298, opened 2026-02-08
Fix negative durations lost when converting to spec

### LINKED ISSUE #3256: Negative durations are lost when converting to spec
**Describe the bug**
When using CarbonInterval with negative intervals, the negative is lost when calling `spec()` method.

**To Reproduce**
```php
echo CarbonInterval::weeks(-2)->spec()
```

**Expected behavior**
Should output `-P2W` or `-P14D`

**Actual behavior**
Outputs `P14D`

**Versions**
 - PHP: 8.4.13
 - Carbon: 3.10.3

**Additional context**
Just wanna say thanks for this amazing library!

QDenka:
## Summary

Fixes #3256

When creating intervals with negative values (e.g. `CarbonInterval::weeks(-2)`), the `spec()` method was returning `P14D` instead of `-P14D` because `getDateIntervalSpec()` used `abs()` on all values without preserving the sign.

## Changes

- **`getDateIntervalSpec()`**: Added detection for intervals where all individual values are negative, prepending `-` to the ISO 8601 duration spec string
- The `invert` flag is intentionally **not** handled here, as it's the standard PHP mechanism and consumers (e.g. `format()`) typically handle it separately

## Examples

| Before | After |
|--------|-------|
| `CarbonInterval::weeks(-2)->spec()` → `P14D` | `CarbonInterval::weeks(-2)->spec()` → `-P14D` |
| `CarbonInterval::days(-5)->spec()` → `P5D` | `CarbonInterval::days(-5)->spec()` → `-P5D` |
| `CarbonInterval::hours(-3)->spec()` → `PT3H` | `CarbonInterval::hours(-3)->spec()` → `-PT3H` |

## Tests

Added 5 new test cases in `tests/CarbonInterval/SpecTest.php`:
- Negative weeks, days, hours
- Negative mixed date and time interval
- Verification that inverted intervals (`invert=1`) don't include sign in spec (handled separately by consumers)

### DIFF
--- a/src/Carbon/CarbonInterval.php
+++ b/src/Carbon/CarbonInterval.php
@@ -2503,7 +2503,27 @@ class CarbonInterval extends DateInterval implements CarbonConverterInterface, U
             static::PERIOD_SECONDS => $seconds,
         ]);
 
-        $specString = static::PERIOD_PREFIX;
+        // Determine if the interval is negative based on individual negative values
+        // Note: $interval->invert is not checked here as it is the standard PHP mechanism
+        // and consumers (e.g. format()) typically handle it separately.
+        $negative = false;
+        $values = [$interval->y, $interval->m, $interval->d, $interval->h, $interval->i, $interval->s];
+        $hasNegative = false;
+        $hasPositive = false;
+
+        foreach ($values as $value) {
+            if ($value < 0) {
+                $hasNegative = true;
+            } elseif ($value > 0) {
+                $hasPositive = true;
+            }
+        }
+
+        if ($hasNegative && !$hasPositive) {
+            $negative = true;
+        }
+
+        $specString = ($negative ? '-' : '').static::PERIOD_PREFIX;
 
         foreach ($date as $key => $value) {
             $specString .= $value.$key;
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Hi 👋 

Thank you for your contribution.

I checked [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601#Durations), and could not find any mention of "negative" in the  duration section.

So, I don't think we should change the default behavior of the `spec()` method.

I also think that handling negative units, but not handling  `->invert` can lead to even more surprise than the current behavior (not handling any negative value).

We also have to think about how to handle mixed intervals (1 day - 5 minutes).

It should probably be another method to keep `spec()` strictly ISO-8601 compliant.

I will consider this for the version 3.13.

kylekatarnls:
Will be fixed via https://github.com/CarbonPHP/carbon/pull/111 in version 3.12.0:

`CarbonInterval::weeks(-2)->spec(withNegatives: true)` => `P-14D`

`CarbonInterval::weeks(2)->invert()->spec(withNegatives: true)` => `-P14D`

`CarbonInterval::years(-2)->months(5)->days(-8)->spec(withNegatives: true)`
=> `P-2Y5M-8D`

`CarbonInterval::years(-2)->months(5)->days(-8)->invert()->spec(withNegatives: true)`
=> `-P-2Y5M-8D`

Without `withNegatives: true` all values are absolutes (no minus sign anywhere) so the behavior is backward-compatible.

PR #3333, opened 2026-05-24
#3307 Fix ->end property not updating after setEndDate() on PHP 8.2+

Abdu-Ahmed:
Fixes #3307

## Problem
On PHP 8.2+, `DatePeriod`'s native properties (`$start`, `$end`, etc.) are managed internally by the engine and cannot be written directly after construction. Calling `setEndDate()` correctly updated Carbon's internal `$endDate` property but left the native `$end` property stale:

```php
$period = CarbonPeriod::between('2024-01-01', '2024-12-31');
$period->setEndDate('2025-12-31');
echo $period->end;          // returned 2024-12-31 which is wrong, (stale)
echo $period->getEndDate(); // returned 2025-12-31 which is correct
```

## Fix
created a new method, `syncNativePeriod()` which re-runs `parent::__construct()` with the current state after `setStartDate()` or `setEndDate()` is called, forcing the native `DatePeriod` properties to sync. the method is a no-op on PHP < 8.2 where the properties are writable directly.

## Tests
Added `testEndPropertySyncsAfterSetEndDate()` in `GettersTest.php`.
All 288 existing `CarbonPeriod` tests pass (1302 assertions).

### DIFF
--- a/src/Carbon/CarbonPeriod.php
+++ b/src/Carbon/CarbonPeriod.php
@@ -1313,6 +1313,7 @@ class CarbonPeriod extends DatePeriodBase implements Countable, JsonSerializable
             $self = $self->toggleOptions(static::EXCLUDE_START_DATE, !$inclusive);
         }
 
+        $self->syncNativePeriod();
         return $self;
     }
 
@@ -1332,11 +1333,15 @@ class CarbonPeriod extends DatePeriodBase implements Countable, JsonSerializable
             throw new InvalidPeriodDateException('Invalid end date.');
         }
 
+        $self = $this->copyIfImmutable();
+
         if (!$date) {
-            return $this->removeFilter(static::END_DATE_FILTER);
+            $self = $self->removeFilter(static::END_DATE_FILTER);
+            $self->syncNativePeriod();
+
+            return $self;
         }
 
-        $self = $this->copyIfImmutable();
         $self->endDate = $date;
 
         if ($inclusive !== null) {
@@ -1344,10 +1349,12 @@ class CarbonPeriod extends DatePeriodBase implements Countable, JsonSerializable
         }
 
         if (!$self->hasFilter(static::END_DATE_FILTER)) {
-            return $self->addFilter(static::END_DATE_FILTER);
+            $self = $self->addFilter(static::END_DATE_FILTER);
+        } else {
+            $self->handleChangedParameters();
         }
 
-        $self->handleChangedParameters();
+        $self->syncNativePeriod();
 
         return $self;
     }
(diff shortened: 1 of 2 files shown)


kylekatarnls:
Hello, at first sight, `testEndPropertySyncsAfterSetEndDate` is indeed a desired behavior, sadly it does not pass here in the CI: https://github.com/briannesbitt/Carbon/actions/runs/26362167994/job/77760262550?pr=3333

Abdu-Ahmed:
hey there!
thanks for the quick feedback! The test was not reassigning the result of  `setEndDate()`, which meant the immutable variant returned a new object  that was discarded. fixed by reassigning:

```php
$period = $period->setEndDate('2025-06-15');
```
All 288 tests now pass locally. finger crossed and pushing the fix now.

kylekatarnls (approved):

src/Carbon/CarbonPeriod.php
```diff
@@ -1313,6 +1313,7 @@ public function setStartDate(mixed $date, ?bool $inclusive = null): static
             $self = $self->toggleOptions(static::EXCLUDE_START_DATE, !$inclusive);
         }
 
+        $self->syncNativePeriod();
         return $self;
```
kylekatarnls, line 1318:
```suggestion

        return $self;
```

src/Carbon/CarbonPeriod.php
```diff
@@ -2471,6 +2478,29 @@ protected function handleChangedParameters(): void
...
+        // Reinitialize the parent DatePeriod to update $start, $end, etc.
+        // This mirrors the logic in __construct and initializeSerialization.
+        parent::__construct(
+            $this->startDate,
+            $interval,
+            $this->endDate ?? max(1, min(2147483639, $this->recurrences ?? 1)),
+            $this->options ?? 0,
+        );
+    }
+
```
kylekatarnls, line 2503:
```suggestion
```

kylekatarnls (dismissed):
Small changes needed to comply with PHP-CS-Fixer

src/Carbon/CarbonPeriod.php
```diff
@@ -2471,6 +2478,29 @@ protected function handleChangedParameters(): void
         $this->validationResult = null;
     }
 
+    /**
+     * Synchronize the native DatePeriod properties with the current state.
+     */
+    protected function syncNativePeriod(): void
+    {
+        if (\PHP_VERSION_ID < 80200) {
+            return;
```
kylekatarnls: We don't run coverage on PHP 8.1, so it's fine to ignore the coverage here, so to keep the coverage indicator at 100%
```suggestion
            return; // @codeCoverageIgnore
```

PR #3337, opened 2026-06-11
Fix incorrect casing of `ucFirst()` in `phpdoc.php`

Soean:
Replace `ucFirst()` with its canonical lowercase form `ucfirst()`.

This is flagged as a case-sensitivity violation by the upcoming [PHP 8.6 RFC](https://wiki.php.net/rfc/case_sensitive_php), which will emit `E_DEPRECATED` for function references that don't match their declared casing. Fixing it now keeps Carbon ahead of the deprecation.

### DIFF
--- a/phpdoc.php
+++ b/phpdoc.php
@@ -243,7 +243,7 @@ foreach ($tags as $tag) {
                     $autoDocLines[] = [
                         '@method',
                         'bool',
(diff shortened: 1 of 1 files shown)


kylekatarnls (approved):

kylekatarnls:
Thanks 🙏

PR #3357, opened 2026-09-02
Add e, p, X and x to the translatedFormat() replacement map

dylanpulver:
`getFormatsToIsoReplacements()` is missing `e`, `p`, `X` and `x`, so `translatedFormat()` emits the literal character instead of the value. On released 3.13.2, `translatedFormat('e')` returns `e` where `format('e')` returns `America/New_York`; same shape for `p`, `X` and `x`.

They are added as `true`, the same treatment the neighbouring `T`/`O`/`P`/`Z` tokens already get (`'['.rawFormat($char).']'`), since none of them contains anything translatable. #2840 added `T` this way.

Measured in the `en` locale over 8 timezones x 63 dates x all 41 documented `date()` characters = 20,664 cells, compared against `DateTimeImmutable::format()`:

| | mismatching cells |
|---|---|
| master | 2016 (the 4 characters x 504 date/zone pairs) |
| this branch | 0 |

Worth stating because it looks right and is not: mapping them onto existing isoFormat tokens instead (`'e' => 'zz'`, `'p' => 'Z'`, `'X'`/`'x' => 'YYYY'`) still leaves 567 cells wrong — isoFormat `Z` yields `+00:00` where `date('p')` yields `Z` in UTC, and `YYYY` drops the sign that `date('X')` requires.

The new test compares every documented `date()` character against `DateTimeImmutable::format()` rather than hard-coded strings, so a future omission from the map fails too. It is version-safe on PHP 8.1, where `X`/`x` are not yet format characters and both sides pass the letter through unchanged.

`vendor/bin/phpunit`: master 6207 tests / 194111 assertions, this branch 6209 / 194199, no new failures. `TestingAidsTest::testSetTestNow` fails intermittently on both arms — it asserts two consecutive `Carbon::now()` calls differ — and passes in isolation; unrelated to this change. `php-cs-fixer --dry-run` and `phpstan` report an identical set of pre-existing findings on master and on this branch.

Disclosure: prepared with AI assistance (Claude Opus 5, `claude-opus-5`). I have reviewed the diff and verified every measurement above myself.


kylekatarnls (approved):

kylekatarnls:
Thanks.

PR #3359, opened 2026-09-05
fix: numeric round()/floor()/ceil() precision >= 60 seconds does not carry into minute/hour/day

dualfroz:
## Problem (self-found)

`Carbon::round()`, `floor()`, and `ceil()` accept a numeric `$precision` that is
documented and tested (see `tests/Carbon/RoundTest.php::testRoundWithDefaultUnit`)
as "a number of seconds". For precisions below 60 this works correctly (e.g.
`round(5)` rounds to the nearest 5 seconds). But for a numeric precision of
`>= 60` (e.g. `round(3600)`, meant to mean "nearest hour"), the result silently
stays confined to the `second` field and never carries into `minute`/`hour`/
`day`, unlike the equivalent string form `round('hour')`, which is correct.

Reproduced on HEAD (`83a67d4`, unpatched):

```
$dt = Carbon::parse('2026-01-01 14:21:37');

$dt->copy()->round(3600)   -> 2026-01-01 14:21:00   (WRONG: only seconds zeroed)
$dt->copy()->round('hour') -> 2026-01-01 14:00:00   (correct)

$dt2 = Carbon::parse('2026-01-01 14:30:00');
$dt2->copy()->ceil(3600)   -> 2026-01-01 14:30:00   (WRONG: no carry at all)
$dt2->copy()->ceil('hour') -> 2026-01-01 15:00:00   (correct)

$dt3 = Carbon::parse('2026-01-01 14:59:59');
$dt3->copy()->floor(3600)   -> 2026-01-01 14:59:00  (WRONG: only seconds zeroed)
$dt3->copy()->floor('hour') -> 2026-01-01 14:00:00  (correct)
```

The same defect also affects a numeric-precision string that resolves to the
`second` unit, e.g. `round('3600 seconds')` (also wrong on HEAD, also fixed by
this change), because it goes through the exact same code path once parsed.

## Root cause

`src/Carbon/Traits/IntervalRounding.php::roundWith()` hardcodes `$unit =
'second'` for any raw numeric `$precision` and forwards the raw value
unchanged to `roundUnit('second', $precision, $function)`.

Inside `src/Carbon/Traits/Rounding.php::roundUnit()` (originally around line
130, `$normalizedValue = floor($function(($value - $minimum) / $precision) *
$precision + $minimum);`), the value being rounded is `$this->second` (range
`0-59`) plus a negligible microsecond fraction — never the minute/hour
contribution. With `$precision = 3600`, `(value - 0) / 3600` is always `< 1`
for any legal second value, so the rounded result is always `0`, and only the
`second` (and finer, `microsecond`) fields ever get touched. Coarser fields
(minute, hour, day, ...) are only ever adjusted by `roundUnit()` through the
*setter* of the *matched* unit overflowing into the next coarser field (e.g.
`->hour(24)` naturally carries into the next day) — but since the matched
unit here is always `second`, and `second`'s own natural value never reaches
anywhere near `3600`, that carry mechanism never triggers.

This was confirmed by comparing with `roundHour()`, `roundMinute()`,
`roundDay()` (all correct today) and with `round(CarbonInterval::hours(1))` /
`round(new DateInterval('PT1H'))` (also correct today, because the
`DateInterval` branch of `roundWith()` converts to a human string like
`"1 hour"` via `forHumans()`, which resolves to `$unit = 'hour'`, not
`'second'`).

## Fix

`src/Carbon/Traits/Rounding.php`, inside `roundUnit()`, right after the
existing `week` -> `day` re-targeting block (`$precision *=
static::DAYS_PER_WEEK`) and the `$precision *= $factor;` line: when the
resolved unit is `second` and the precision is a plain `int`/`float` (i.e.
not a fractional-second precision that was already correct, and not a
`DateInterval`), re-target the unit to `minute`/`hour`/`day` once the
precision reaches or exceeds that unit's own span in seconds, rescaling the
precision accordingly:

- `precision >= 86400` (`HOURS_PER_DAY * MINUTES_PER_HOUR *
  SECONDS_PER_MINUTE`) -> unit `day`, `precision /= 86400`
- `precision >= 3600` -> unit `hour`, `precision /= 3600`
- `precision >= 60` -> unit `minute`, `precision /= 60`
- otherwise unchanged (`second`, same as before)

This mirrors the pre-existing `week` -> `day` pattern already in the method
and reuses the exact same, already-correct per-unit rounding algorithm that
`round('hour')`, `roundHour()`, etc. use — so all existing carry logic (DST,
month-length, negative years, fractional sub-unit precision) is inherited
rather than reimplemented. Sub-minute numeric precisions (`< 60`) are
completely untouched by the new branch and keep their exact previous
behavior.

Only `src/Carbon/Traits/Rounding.php` (source) and
`tests/Carbon/RoundTest.php` (test) were changed, per this fix's scope.

## Consistency vs. the string-unit path

There is no ambiguity for the primary cases (exact multiples of
60/3600/86400): the numeric result now matches the
string-unit result exactly (verified below, `eq()` assertions). For a numeric
precision that is not an exact multiple (e.g. `round(90)`), the fix uses the
same fractional-precision mechanism already used by `roundMinute(1.5)` etc.
(pre-existing, unrelated to this bug) — this was not required by the bug
report and is not asserted in the new test, but does not regress or throw
where it previously silently (and incorrectly) succeeded.

## Full suite (with the fix, real output)

All three pre-existing issues are unrelated to this change and were
reproduced identically with the fix reverted:
- The two `LocalizationTest` errors are environmental (the `fr_FR.UTF-8`
  locale is not installed in this sandbox); they errored the same way before
  this fix was applied.
- `TestingAidsTest::testSetTestNow` passes cleanly (27/27) when run in
  isolation (`vendor/bin/phpunit tests/Carbon/TestingAidsTest.php`), with or
  without this fix; it only fails as part of the full-suite run, indicating a
  pre-existing test-order/state-leakage flake unrelated to rounding.

Also ran, scoped to the changed files only (both clean):


kylekatarnls:
Hello, some may have started to rely on this behavior (can use `->round($n)` where `$n` can be any number but they rely on Carbon to cap it at `60`), so I have to think how acceptable is the BC here.

Also, if we change the behavior we should opt for something consistent, not just a fix for the `second` unit.

For instance, following the logic you propose, then `Carbon::parse('2026-01-01 14:21:03')->roundMinutes(24 * 60)` should return `2026-01-02 00:00:00`, but currently it returns `2026-01-01 14:00:00` and the result remains the same on your branch.

So while I can see why you consider the current result incorrect and where the fix can make sense, I will still hold on this change for now as users are not blocked and the alternatives available (`->roundHour()` and `->roundDay()`) are actually more explicit (so IMO better) than `->round(3600)` or `->round(86400)`

BTW, as a last note, I generally recommend to avoid using "24 hours" as a substitute of "1 day", if you mean to talk about the duration of a day, then prefer the explicit "1 day" because not all days last 24 hours because of daylight saving time, for example 2026-11-01 will last 25 hours, so the expected result for a 24-hours rounding on that day is ambiguous.

```php
$a = Carbon::parse('2026-11-01 00:00 America/Chicago');
$b = Carbon::parse('2026-11-02 00:00 America/Chicago');
$a->diffInHours($b); // 25
$a->diffInDays($b); // 1
```

dualfroz:
fair, extended it to every unit, so `roundMinutes(24 * 60)` gives `2026-01-02 00:00:00` now. can put it behind an opt-in if you prefer

kylekatarnls (approved):

